PR #14, opened 2020-08-25
Hugo website first draft

amueller:
this is a very basic version of the hugo website, no sphinx yet.


website/build_site.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+cp -r ../documentation content/
+jupyter nbconvert ../source/Mlos.Notebooks/*.ipynb --to markdown --output-dir content/notebooks
+
+cp ../*.md content/
+cp ../LICENSE content/
+cp ../README.md content/_index.md
+mv content/documentation/README.md content/documentation/_index.md
+sed -i 's/.md/\//g' *.md
```
amueller, line 10: not the most specific regex in the world, but it works? hm should the ``.`` be escaped? I'm no good with sed

README.md
```diff
@@ -66,18 +66,54 @@ To achieve this MLOS provides:
...
-  1. [documentation/01-Prerequisites.md](./documentation/01-Prerequisites.md)
-  2. [documentation/02-Build.md](./documentation/02-Build.md)
+  1. [Prerequisites](./documentation/01-Prerequisites.md)
+  2. [Build](./documentation/02-Build.md)
 
 ## Examples
 
-Code and documentation for examples of using MLOS to optimize a system are described in [documentation/03-ExampleUsage.md](./documentation/04-ExampleUsage.md) and in the [source/Examples](./source/Examples/) tree.
+Code and documentation for examples of using MLOS to optimize a system are described in the [Notebooks](./notebooks/) section. Additional code is in the  [source/Examples](https://github.com/microsoft/MLOS/tree/main/source/Examples) source directory.
+You can find the source of the notebooks [on github as well](https://github.com/microsoft/MLOS/tree/main/source/Mlos.Notebooks).
```
bpkroth, line 116: Do relative links not work here?

One issue with this is if they're browsing on one commit of the tree, then it could drop them to a different one.

README.md
```diff
@@ -66,18 +66,54 @@ To achieve this MLOS provides:
...
+
+```
+$ conda create -n mlos_environment
+$ conda activate mlos_environment
+```
+
+You can download the mlos code using git:
+
+```
+$ git clone git@github.com:microsoft/MLOS.git
```
bpkroth, line 83: Should we include an https:// example as well?

website/build_site.sh
```diff
@@ -0,0 +1,16 @@
...
+cp ../*.md content/
+cp ../LICENSE content/
+cp ../README.md content/_index.md
+mv content/documentation/README.md content/documentation/_index.md
+sed -i 's/\.md/\//g' content/*.md
+sed -i 's/\.md/\//g' content/documentation/*.md
+
+
+if [ ! -d "themes/book" ]; then
+    git clone --depth 1 https://github.com/alex-shpak/hugo-book.git themes/book/
```
bpkroth, line 15: Thanks for this.

Should we include a tag that we want it checked out at so that it doesn't magically change style on us at some point?

website/build_site.sh
```diff
@@ -0,0 +1,16 @@
...
+
+cp -r ../documentation content/
+jupyter nbconvert ../source/Mlos.Notebooks/*.ipynb --to markdown --output-dir content/notebooks
+
+cp ../*.md content/
+cp ../LICENSE content/
+cp ../README.md content/_index.md
+mv content/documentation/README.md content/documentation/_index.md
+sed -i 's/\.md/\//g' content/*.md
+sed -i 's/\.md/\//g' content/documentation/*.md
```
bpkroth, line 11: Thanks for this as well.

Maybe also include a comment regarding why we're doing the `sed` (e.g. to preserve relative links for browsing in github as well).

README.md
```diff
@@ -87,9 +123,9 @@ Code and documentation for examples of using MLOS to optimize a system are descr
 
 ## Contributing
 
-We welcome contributions!  Please see [CONTRIBUTING.md](./CONTRIBUTING.md) and [CODE_OF_CONDUCT.md](./CODE_OF_CONDUCT.md) for details.
+We welcome contributions!  Please see [Contributing](./CONTRIBUTING.md) and [Code of Conduct](./CODE_OF_CONDUCT.md) for details.
```
bpkroth, line 126: Much nicer :)

README.md
```diff
@@ -66,18 +66,54 @@ To achieve this MLOS provides:
...
+You can also install the package directly without checking out the code:
+
+```
+$ pip install "git+https://github.com/microsoft/MLOS.git#egg=mlos&subdirectory=source/Mlos.Python"
+```
+
+However, this does not include the examples.
+
+After this installation, you can run any of the Python-only example notebooks.
+A good place to start is the [introduction to Bayesian Optimization](#).
```
bpkroth, line 102: This seems incomplete.

"A good place to start the [I{captial I?}ntroduction to Bayesian Optimization](# a link should go here?) notebook (missing word?)."

README.md
```diff
@@ -66,18 +66,54 @@ To achieve this MLOS provides:
...
+
+```
+$ cd MLOS/source/Mlos.Python
+$ pip install -e .
+```
+
+You can also install the package directly without checking out the code:
+
+```
+$ pip install "git+https://github.com/microsoft/MLOS.git#egg=mlos&subdirectory=source/Mlos.Python"
```
bpkroth, line 96: Nice.

Probably a discussion for later if we should publish that code as it's own package to make this even easier eventually.

README.md
amueller: relative from where to where? this is rendered as a html on ``microsoft.github.io/mlos`` so I'm not sure where you'd want to link from there.

README.md
amueller: Adding "notebook" might be more explicit but saying "A good place to start is the Introduction to Bayesian Optimization Notebook" is a complete sentence, right? Is it not clear?

README.md
amueller: Should we do only one or the other?

bpkroth: I think it depends on whether we want to encourage commits or not, right?

If it's only for consumption, then `https` is more cross platform and readily available without additional setup cost.

Should we just assume that if you're going to submit PRs, then you probably know how to fill in the details of either using a different clone URL or else updating the `git remote` afterwards?

README.md
bpkroth: This is the top level `README.md`, so if I for some reason end up at https://github.com/amueller/MLOS/tree/hugo_website, then if I click the "Prerequisites" link, I will end up on your version of that page still.  On the other hand the "source/Examples" link will now send me back to the original repo with a different revision.

website/build_site.sh
bpkroth: Looks good.  Thanks!

README.md
amueller: Yes. I'm just asking where else it should send you for the rendered page, because there is no relative paths to the source, right?

I can also remove this but then there is no link to github from the rendered pages.

README.md
amueller: yeah probably, so let's do https.

README.md
bpkroth: My comment was more about the fact that the link looked incomplete.

The rest of the wording tweaks you could take or leave, they were just minor nits.

README.md
bpkroth: For the rendered page I think we can send them to the `main` branch version as you outlined here - that makes perfect sense to me since they're already navigating from an external source.

It was moreso can we do that (e.g. with more `sed` magic?) but still keep the relative links inside the git markdown?

README.md
amueller: ah ok. Yes, that required the PR to be merged that you just merged :)

README.md
amueller: I don't see how we would do that. Maybe if the link ends with a ``/`` it should go to github? But that seems super brittle and we probably also need to start using a markdown parser if we want to do these kinds of rewrites.

README.md
amueller: done :)

README.md
bpkroth: I think I see your point now.  A hackish way I could see to do it would be to append an anchor to the link and then regexp replace those strings specially.

For instance: 

```
[source/Examples/](./source/Examples/#mlos-github-tree-browse)
```

```sh
sed -r -e 's|\(\.(/[^#]+)#mlos-github-tree-browse\)|(https://github.com/microsoft/MLOS/tree/main/\1)/'
```

However, your point about it being brittle is well taken and this is perhaps not worth it, so I'm fine if you just want to leave these as is.

bpkroth:
We should make a separate issue for tracking regenerating this and deploying it via CI.

bpkroth:
> We should make a separate issue for tracking regenerating this and deploying it via CI.

Also probably documenting how to.

README.md
amueller: the anchor is a neat idea, I was wondering how we could attach information to the link. But I think it's fine to leave as-is for now.

PR #15, opened 2020-08-26
Sphinx docs

amueller:
this adds stuff to generate sphinx docs for the python module. not perfect yet but better than nothing.


amueller:
unfortunately the exclude pattern in apidoc doesn't really work, so we probably have to clean up the files manually to remove the tests.

PR #28, opened 2020-09-02
Porting changes from ADO to Github

bpkroth:


bpkroth:
Needs to be rebase merged when ready.

amueller:
looks good. Some of the makefiles and .mk files look generated right? I assume you checked in only what needed to be check in, since I haven't reviewed the whole build process.

bpkroth:
> 
> 
> looks good. Some of the makefiles and .mk files look generated right? I assume you checked in only what needed to be check in, since I haven't reviewed the whole build process.

I took another look, but didn't see what you were referring to.  I think they're all my own work 😬  Do you have an example you're concerned about?  Else, we can fix it up after this.

Issue #31, opened 2020-09-02
Fixup python dependency setup instructions for consistency


bpkroth:
I think during #12 (https://github.com/microsoft/MLOS/pull/12/files#diff-758669e740e7424c31dfb6cf2edbd749) @amueller and I discussed making the instructions here https://github.com/microsoft/MLOS/blob/main/documentation/01-Prerequisites.md#install-python-dependencies consistent with the instructions at https://github.com/microsoft/MLOS#python-only-installation

PR #41, opened 2020-09-02
Reorganize doc flow by OS rather than by Task

curino:
Attempt to reorder docs by OS instead of by step. Maybe good maybe not.


documentation/01-Prerequisites.md
```diff
@@ -6,76 +6,46 @@ These are one-time setup instructions that should be executed prior to following
...
-
-> Where `UbuntuVersion` can also be set to another supported version of Ubuntu.
-
-> Tip: you can also pass `--build-arg=http_proxy=http:/some-proxy-caching-host:3128` to help direct `apt` and `pip` to fetch the necessary packages via local caches.
-
-See [02-Build.md](./02-Build.md#docker) for instructions on how to run this image.
 
 ### Linux build tools
 
+#### Linux Build Manually
```
bpkroth, line 48: This section isn't about building on Linux manually.
How about "Manual Linux Build Environment Setup" or something less verbose like "Manual Setup"

documentation/01-Prerequisites.md
```diff
@@ -111,6 +81,113 @@ sudo apt-get install exuberant-ctags
 
 > When available `make ctags` can be invoked to help generate a `tags` database at the root of the source tree to allow easier code navigation in editors that support it.
 
+#### Docker
```
bpkroth, line 84: Maybe too terse in this case?
"Create a docker image for building"

bpkroth, line 84: Just above here, there is a path error:
`./script/...` should be `./scripts/...`

Do you want to fix that while you're in here or shall I make a separate PR?

documentation/01-Prerequisites.md
```diff
@@ -6,76 +6,46 @@ These are one-time setup instructions that should be executed prior to following
...
-    - [Linux](#linux-python-install)
-    - [Windows](#windows-python-install)
-  - [Install Docker](#install-docker)
-    - [Linux Docker Install](#linux-docker-install)
+      - [Windows Build Manually](#win-build-manually)
+    - [Windows Pythoon Install](#windows-python-install)
     - [Windows Docker Install](#windows-docker-install)
+      
 
 > Note: Most Windows shell commands here expect `powershell` (or [`pwsh`](https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell-core-on-windows)).
```
bpkroth, line 31: We should probably move that comment down by the Windows section instead.

documentation/01-Prerequisites.md
```diff
@@ -369,3 +405,13 @@ As mentioned above, Docker on Windows first requires a Linux VM.
...
 
       From within this shell environment, `docker` cli commands should be proxied through to your `docker-dev-vm` prepared by `docker-machine`.
+
+
+
+--------------
+
+
+
+## Install build tools
```
bpkroth, line 415: Dangling section

documentation/01-Prerequisites.md
```diff
@@ -6,76 +6,46 @@ These are one-time setup instructions that should be executed prior to following
...
       - [Using Chocolatey](#using-chocolatey)
-      - [Manually](#manually)
-  - [Install Python Dependencies](#install-python-dependencies)
-    - [Linux](#linux-python-install)
-    - [Windows](#windows-python-install)
-  - [Install Docker](#install-docker)
-    - [Linux Docker Install](#linux-docker-install)
+      - [Windows Build Manually](#win-build-manually)
+    - [Windows Pythoon Install](#windows-python-install)
     - [Windows Docker Install](#windows-docker-install)
```
bpkroth, line 27: Should see about having vscode automatically update this for us.  Might need an extension.  Perhaps one of these?
davidanson.vscode-markdownlint
shd101wyy.markdown-preview-enhanced

documentation/01-Prerequisites.md
```diff
@@ -111,6 +81,113 @@ sudo apt-get install exuberant-ctags
...
+> Where `UbuntuVersion` can also be set to another supported version of Ubuntu.
+
+> Tip: you can also pass `--build-arg=http_proxy=http:/some-proxy-caching-host:3128` to help direct `apt` and `pip` to fetch the necessary packages via local caches.
+
+See [02-Build.md](./02-Build.md#docker) for instructions on how to run this image.
+
+
+### Clone the repository
+
+Cross platform
```
bpkroth, line 104: This comment can probably be dropped.

documentation/01-Prerequisites.md
```diff
@@ -111,6 +81,113 @@ sudo apt-get install exuberant-ctags
...
+    sudo apt-get -y install python3-pip python3.7-dev \
+        build-essential libfreetype-dev unixodbc-dev
+    ```
+
+    ```sh
+    python3.7 -m pip install --upgrade pip
+    python3.7 -m pip install setuptools
+
+    python3.7 -m pip install -r source/Mlos.Python/requirements.txt
+    ```
```
bpkroth, line 138: Dunno if this should be addressed here or as a separate PR but #31, #45, and #43 all mention additional improvements in Python dependency setup.

documentation/01-Prerequisites.md
```diff
@@ -6,76 +6,46 @@ These are one-time setup instructions that should be executed prior to following
 
 - [Prerequisites for building and using MLOS](#prerequisites-for-building-and-using-mlos)
   - [Contents](#contents)
-  - [Requirements](#requirements)
-  - [Clone the repository](#clone-the-repository)
-  - [Install build tools](#install-build-tools)
-    - [Docker](#docker)
+  
+  - [Linux](#linux)
```
bpkroth, line 10: Should we add an OS X section as well, maybe just for Python and Docker?

documentation/01-Prerequisites.md
bpkroth: Nope it was joffreykern.markdown-toc or yzhang.markdown-all-in-one

documentation/01-Prerequisites.md
bpkroth: Discussed with @amueller earlier.  I think we're going to leave it here for now and fix it up afterwards.

For now I left a section for Conda install instructions as well.

bpkroth:
@curino I finished another round of reorging this.  Can you talk a look please?

Will try and take a look at the Build instructions tomorrow as well.

sergiy-k:
This is much easier to follow than before. Thank you!

sergiy-k (approved):

documentation/01-Prerequisites.md
sergiy-k: If we decide to add description for OS X then may be we should split the guide into separate files (one file for each OS) to keep each guide compact and focused? What do you think?

bpkroth: Probably makes sense.  There's some redundancy then but so it goes.
One complication we're still working out is how best to keep relative links in the markdown files working so that browsing on github is natural (e.g. stays within tree, renders, etc.) but the github.io pages always take you back to master.
I think I have a hack to handle it.  Will work on it tomorrow.

curino: I was about to start the OS X section, but shy away as I suspected I was not going to be able to finish it. Agreed on splitting if we add more than 2 OS

PR #47, opened 2020-09-03
Add Environment yml, improve conda docs

amueller:


source/Mlos.Python/setup.py
```diff
@@ -9,14 +9,14 @@
     version="0.0.1",
     packages=find_packages(),
     install_requires=[
-        'bayesian-optimization==1.0.1',
-        'scikit-learn==0.22.1',
-        'scipy==1.3.2',
-        'numpy==1.18.1',
-        'pandas==1.0.3',
-        'pylint==2.3.1',
-        'pyodbc==4.0.30',
```
amueller, line 18: I don't like this but I can't get pip to pick up on the conda version of pyodbc. conda shows 4.0.30, pip shows 4.0.0-unsupported.

README.md
```diff
@@ -70,33 +70,39 @@ To achieve this MLOS provides:
...
 Some of the examples require only the installation of the mlos Python library. These examples do not use the shared memory infrastructure.
-To use this simplified installation, it's recommended to use the Anaconda python distribution and create a new conda environment:
+
+
+First, download the mlos code using git:
 
 ```
-$ conda create -n mlos_environment
-$ conda activate mlos_environment
+$ git clone git@github.com:microsoft/MLOS.git
```
bpkroth, line 78: I think we discussed this earlier, but I don't know if we came to a conclusion.

I'm thinking we should standardize on https git remote urls for now.

README.md
```diff
@@ -70,33 +70,39 @@ To achieve this MLOS provides:
 ## Python only installation
 
 Some of the examples require only the installation of the mlos Python library. These examples do not use the shared memory infrastructure.
-To use this simplified installation, it's recommended to use the Anaconda python distribution and create a new conda environment:
+
+
+First, download the mlos code using git:
```
bpkroth, line 75: backticks? if you're trying to reference the python package, else capitalization to be consistent with the project name

README.md
```diff
@@ -70,33 +70,39 @@ To achieve this MLOS provides:
...
-$ git clone git@github.com:microsoft/MLOS.git
+$ conda env create -f MLOS/source/Mlos.Notebooks/environment.yml
+```
+The environment will be called ``mlos_python_environment`` and you can activate it as follows:
+```
+$ conda activate mlos_python_environment
 ```
 
-To install the Python library, change to the Python directory and install with ``pip``:
+Use ``pip`` to install the Python library:
```
bpkroth, line 93: Does ADO respect double backtick?

source/Mlos.Python/setup.py
bpkroth: But they're exact version matches in the environment.yml?  Shouldn't they be consistent?

README.md
bpkroth: Sorry, that didn't highlight the way I was expecting.  The comment was in reference the the bare lowercase "mlos" in that sentence.

source/Mlos.Python/setup.py
amueller: When they match, it doesn't work, is what I'm saying.

README.md
amueller: that's ... in another PR? I'll fix it here.

README.md
amueller: fixed.

README.md
amueller: did I do the thing that you wanted? (you can also suggest changes here ;)

source/Mlos.Python/setup.py
bpkroth: Weird, but okay.

README.md
```diff
@@ -70,33 +70,39 @@ To achieve this MLOS provides:
...
+Use `pip` to install the Python library:
 
 ```
-$ cd MLOS/source/Mlos.Python
-$ pip install -e .
+$ pip install MLOS/source/Mlos.Python/
 ```
 
-You can also install the package directly without checking out the code:
+Alternatively you can also install the package directly without checking out the code:
```
bpkroth, line 99: complete nit:
```suggestion
Alternatively, you can also install the package directly without checking out the code:
```

bpkroth (approved):

source/Mlos.Python/setup.py
amueller: yes, if you conda install 4.0.30, conda list shows 4.0.30, but pip shows 4.0.0-unsupported which is not >=4.0 and not even >3. So I didn't really know what else to do. Going to pip for pyodbc doesn't seem to be an option because many wheels seem to be missing.

Issue #40, opened 2020-09-02
Add timeout to CI checks


bpkroth:
A python unit test failed in Github Action pipeline, but it took a long time:
https://github.com/microsoft/MLOS/pull/29/checks?check_run_id=1061557146#step:9:41362

We should see about controlling that with some timeouts.

bpkroth:
As mentioned by @edcthayer this could also just be due to that particular run selecting a particularly complex polynomial that took longer to execute.

We could fix that individual check to restrict the polynomial complexity, but we should probably also set explicit limits on CI checks if we can.

PR #51, opened 2020-09-03
Fixup pipeline timeouts

bpkroth:
Add explicit CI pipeline timeouts and reduce some function complexity for a test to address some flaky CI failures #40 

Also adds explicit calling of python3.7 to address environments that have both 3.7 and 3.8 with python3 linking to 3.8

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
   windows-msbuild-debug:
     name: Run debug msbuild on Windows
     runs-on: windows-2019
-    timeout-minutes: 10
+    timeout-minutes: 12
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
     # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
@@ -70,7 +70,7 @@ jobs:
   python-checks:
     name: Run Python checks
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 10
+    timeout-minutes: 12
     strategy:
       matrix:
         os: [windows-2019, ubuntu-20.04]
@@ -101,9 +101,9 @@ jobs:
       run: scripts/run-python-checks.sh
     - name: Run Python unit tests (Windows)
       if: startsWith(matrix.os, 'windows')
-      timeout-minutes: 5
+      timeout-minutes: 8
       run: scripts\run-python-tests.cmd
     - name: Run Python unit tests (Linux)
       if: startsWith(matrix.os, 'ubuntu')
-      timeout-minutes: 5
+      timeout-minutes: 8
       run: scripts/run-python-tests.sh
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,6 +17,7 @@ jobs:
   windows-msbuild-debug:
     name: Run debug msbuild on Windows
     runs-on: windows-2019
+    timeout-minutes: 10
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
     # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
@@ -31,6 +32,7 @@ jobs:
   windows-msbuild-release:
     name: Run release msbuild on Windows
     runs-on: windows-2019
+    timeout-minutes: 10
     steps:
     - uses: actions/checkout@v2
     - name: Run release msbuild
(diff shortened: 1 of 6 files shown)


scripts/run-python-tests.sh
```diff
@@ -10,6 +10,9 @@ set -eu
 scriptdir=$(readlink -f "$(dirname "$0")")
 cd "$scriptdir/.."
 
+. "$scriptdir/util.sh"
+pythoncmd=`getPythonCmd`
+
 # Linux filesystems are case-sensitive, so we need to tell the python unit test
 # scanner to look for upper-case files as well:
-python3 -m unittest discover --verbose --locals --failfast -s source/Mlos.Python -p "[Tt]est*.py"
+$pythoncmd -m unittest discover --verbose --locals --failfast -s source/Mlos.Python -p "[Tt]est*.py"
```
bpkroth, line 18: I can split this change out to a separate PR if you want.

bpkroth: Basically, in one container I tested with 3.8 and 3.7 were both available, but python3 called 3.8 which broke things.

scripts/util.sh
```diff
@@ -3,6 +3,11 @@
 #
 # A set of helper functions to use in other scripts.
 
+if [ -z "$BASH" ]; then
+    echo "ERROR: This script currently only works using a bash shell." >&2
+    return 1
+fi
```
bpkroth, line 9: Checks for just bash for now since I used a `local ` var below which `sh` doesn't support.

source/Mlos.Python/mlos/SynthethicFunctions/PolynomialObjective.py
```diff
@@ -21,8 +21,8 @@ class PolynomialObjective:
         name="polynomial_objective_config",
         dimensions=[
             ContinuousDimension(name='seed', min=1, max=2 ** 32),
-            DiscreteDimension(name='input_domain_dimension', min=1, max=16),
-            DiscreteDimension(name='max_degree', min=1, max=16),
+            DiscreteDimension(name='input_domain_dimension', min=1, max=5),
+            DiscreteDimension(name='max_degree', min=1, max=5),
```
bpkroth, line 25: Checked with @edcthayer on reducing this.

.github/workflows/main.yml
```diff
@@ -17,6 +17,7 @@ jobs:
   windows-msbuild-debug:
     name: Run debug msbuild on Windows
     runs-on: windows-2019
+    timeout-minutes: 12
```
sergiy-k, line 20: Do you how long it normally takes to execute all these jobs? How much buffer do you have here?

bpkroth: It's been 9-10 the last few I've looked at.  We can increase it if we want.

scripts/run-python-tests.sh
amueller: I feel like conda might be the better end-game but we don't have to do that now.

.github/workflows/main.yml
bpkroth: Sorry, that comment (8-10) was just for the msbuild-debug task.  The others are similarly given a slight buffer based on their current run lengths.
My feeling was to try and keep them somewhat tight at first and open them up later if we need to.

scripts/run-python-tests.sh
bpkroth: Might be true.
My intention was to use the Docker build images to run all of the Linux based checks.  The reduces the need to refetch a lot of the pip/apt requirements everytime.

.github/workflows/main.yml
sergiy-k: Please make sure that the buffer is large enough so we avoid frequent false positives.

sergiy-k (approved):

.github/workflows/main.yml
bpkroth: Sure.  Would you prefer we bump them up further now, or fix them as we get more experience?

sergiy-k: I'm fine either way as long as CI is stable. :) If you think that these are the right number, then go for it!

bpkroth: I'm not a fan of flaky CI either, but I don't have enough data points to be certain on that yet I think.
I'm fine to be flexible about it and fix it up later if we need to.  If nothing else having tighter timeouts should force us to evaluate when something got introduced that is too long vs. something being flaky.

PR #56, opened 2020-09-04
Redrawn MLOS Architecture

curino:
Adding some details on User Input and Code Gen


bpkroth:
First, this looks really nice!
Series of nits to follow in separate comments (for easier tracking)

bpkroth:
Lower dark box: missing closing paren ) around `(Channel` for `Feedback/Actions`

bpkroth:
Middle layer's `Config Shared Memory C#` region points to the shared memory region, but should really point into the `Mlos.Agent (C#)` box.  I think pointing into the `SettingsRegistryManager` is fine.

Then from there a dashed line (no arrow heads?) from right to left connecting to the `ConfigSharedMemory` blue box at the top of the dark shared memory box makes sense to me.  For parity the an arrow from the Internal Components on the left, could also go to the `ConfigSharedMemory` region.

bpkroth:
Is the background for this explicitly white now?  I had an issue with an earlier version reading the dark elements on a dark background when I think the image background was transparent.

bpkroth:
Is the visio version in this generated from the graffle one?  Which is authoritative?
Should we remove the old ones?

bpkroth:
Update the `MlosArchitecture.md` to refer to the `.svg`

amueller:
Nice!
There's a typo in "Config Shared Memory (C+)", missing a "+".

curino:
> Is the visio version in this generated from the graffle one? Which is authoritative?
> Should we remove the old ones?

I think OmniGraffle should be the authorative one, the other is for compatibility and ease of "show me what you need" in Visio. It should export ok back and forth, but not sure.

bpkroth:
> 
> 
> > Is the visio version in this generated from the graffle one? Which is authoritative?
> > Should we remove the old ones?
> 
> I think OmniGraffle should be the authorative one, the other is for compatibility and ease of "show me what you need" in Visio. It should export ok back and forth, but not sure.

The only concern I have with that is that it puts a requirement on having a Mac (you?) to edit it, but I definitely like the output image.

curino:
> Given that the blue arrows are RPC, do we need the GRPC logo? I would try to minimize the noise.
> 
> Also, it's not entirely obvious what the blue vs black text means. It's user system vs MLOS? I'm not sure if I understand the box on the bottom left. So that would be like SQLServer or SQLite, right? I guess it's a bit strange to me that it's on the level that's labeled "MLOS Platform".

I left the GRPC for now (I kinda like it, but happy to remove if other agree is redundant)

and I agree on the complaint on Platform, but I am struggling to visually separate without blowing up the archticture size (with one more layer).  shall we just say Systems/MLOS ?

amueller:
Yes, I think relabeling it to something like "System/MLOS" would make sense.

bpkroth:
> 
> 
> Yes, I think relabeling it to something like "System/MLOS" would make sense.

Maybe some dashed vertical lines to help visualize the separation?

I'd previously suggested that for the local vs. remote parts, but given we discussed eventually making that pluggable, I think it's fine to skip them for that.

bpkroth (approved):
This looks really nice.  Thanks!

PR #58, opened 2020-09-08
make the logging be less verbose in python smart cache notebook

amueller:


source/Mlos.Notebooks/SmartCacheOptimization.ipynb
```diff
@@ -69,7 +70,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "logger = create_logger('Optimizing Smart Cache')\n",
+    "logger = create_logger('Optimizing Smart Cache', logging_level=logging.WARN)\n",
```
bpkroth, line 73: Just to be clear, this is basically the substantive change, correct?

Everything else is just updated execution output?

bpkroth (approved):
LGTM

source/Mlos.Notebooks/SmartCacheOptimization.ipynb
amueller: yes indeed.

PR #59, opened 2020-09-08
More Python docstrings and minor usability improvements for Jupyter

amueller:
Changing `__str__` to `__repr__` and some other slight polish

### DIFF
--- a/source/Mlos.Python/mlos/Grpc/BayesianOptimizerFactory.py
+++ b/source/Mlos.Python/mlos/Grpc/BayesianOptimizerFactory.py
@@ -42,7 +42,7 @@ class BayesianOptimizerFactory:
 
 
         Returns
-        =======
+        -------
         BayesianOptimizerProxy
 
         """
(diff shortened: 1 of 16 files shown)


amueller:
hm not sure about the change in the dimension repr, should check with Adam.

bpkroth:
> 
> 
> hm not sure about the change in the dimension repr, should check with Adam.

@byte-sculptor

source/Mlos.Python/mlos/Grpc/BayesianOptimizerFactory.py
```diff
@@ -42,7 +42,7 @@ def create_remote_optimizer(self, optimization_problem: OptimizationProblem, opt
 
 
         Returns
-        =======
+        -------
```
bpkroth, line 45: Are the fixups resulting from the WARNING output from the sphinx script?

amueller: No, they are are results of me having had more coffee. We should probably fix the warnings. There are probably >100, we can error once we fixed them.

bpkroth (approved):
LGTM

byte-sculptor:
Looks good to me!

PR #60, opened 2020-09-08
Using Context features

amueller:
todo:
* [x] fix computation of optimum
* [x] add tests
* [x] use base class
* [x] properly define product of context and parameters (for some value of properly)


source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/NumericOptimizers/RandomSearchOptimizer.py
```diff
@@ -64,10 +64,19 @@ def suggest(self, context_values_dataframe=None):  # pylint: disable=unused-argu
         :return:
         """
 
-        feature_values_dataframe = self.optimization_problem.parameter_space.random_dataframe(num_samples=self.config.num_samples_per_iteration)
```
byte-sculptor, line 67: We'll need to repeat this logic for glowworm and soon for random near incumbent. I wonder if we could keep it in the UtilityFunctionOptimizer base class to avoid duplication.

source/Mlos.Python/mlos/Optimizers/BayesianOptimizer.py
```diff
@@ -116,8 +118,12 @@ def get_experiment_data(self):
     @trace()
     def suggest(self, random=False, context=None):
```
byte-sculptor, line 121: While we are here, could we add a type annotation to context: Point=None?

source/Mlos.Python/mlos/Optimizers/BayesianOptimizer.py
```diff
@@ -74,12 +74,14 @@ def __init__(
         assert optimizer_config in BayesianOptimizerConfig.CONFIG_SPACE, "Invalid config."
         self.optimizer_config = optimizer_config
 
+        joint_feature_space = SimpleHypergrid('joint_feature_space', [*self.optimization_problem.parameter_space.dimensions, *self.optimization_problem.context_space.dimensions])
```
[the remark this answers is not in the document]
amueller: the ``feature_space`` contains two boolean columns, one of which is always ``True``. Is there a way to drop them when passing them to the random forest surrogate model?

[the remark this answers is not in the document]
byte-sculptor: Several:
1. Least effort - do nothing. The random forest should figure out that these dimensions cannot be used to reduce entropy. But of course it's wasteful too.
2. Write a simple HypergridAdapter that drops either columns by name, or an adapter that drops columns with a constant value. This has the advantage of being reusable.
3. Drop them manually. It has the benefit of being simple, but not reusable.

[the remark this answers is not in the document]
amueller: Doing nothing also means creating additional dataframe columns.
For the other two options, would you do that when creating ``Optimizer.feature_space``? Or just for the forest?

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/NumericOptimizers/RandomSearchOptimizer.py
amueller: this is not on github yet, but will do as soon as it is.

source/Mlos.Python/mlos/Optimizers/BayesianOptimizer.py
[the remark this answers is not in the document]
amueller: did the simple flat thing which is the easiest and also backwards-compatible

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctionOptimizers/GlowWormSwarmOptimizer.py
```diff
@@ -77,7 +77,7 @@ def __init__(
         self.dimension_names = [dimension.name for dimension in self.parameter_adapter.dimensions]
 
     @trace()
-    def suggest(self, context_values_dataframe=None):  # pylint: disable=unused-argument
+    def maximize(self, target_function, context_values_dataframe=None):  # pylint: disable=unused-argument
```
byte-sculptor, line 80: What's the benefit of doing this?

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctionOptimizers/UtilityFunctionOptimizer.py
```diff
@@ -28,6 +28,13 @@ def __init__(
...
-    @abstractmethod
     def suggest(self, context_values_dataframe: pd.DataFrame) -> Point:
-        raise NotImplementedError
+        config_to_suggest = self.maximize(lambda features: self.utility_function(features).utility,
+                                          context_values_dataframe=context_values_dataframe)
+        self.logger.debug(f"Suggesting: {str(config_to_suggest)}")
+        return config_to_suggest
+
+    @abstractmethod
+    def maximize(self, target_function, context_values_dataframe=None):
```
byte-sculptor, line 38: why?

source/Mlos.Python/mlos/Optimizers/OptimizationProblem.py
```diff
@@ -99,19 +104,12 @@ def __init__(
...
-        ).join(
-            subgrid=self.context_space,
-            on_external_dimension=CategoricalDimension(name="contains_context", values=[True])
-        )
+        parameter_names = {parameter.name for parameter in parameter_space.root_dimensions}
+        context_names = [context.name for context in context_space.root_dimensions]
+        if parameter_names.intersection(context_names):
+            raise ValueError("Parameter space and context space can't share dimensions.")
+        self.feature_space = SimpleHypergrid(name='features',
+                                             dimensions=[*parameter_space.dimensions, *context_space.dimensions])
```
byte-sculptor, line 112: I think it would be better to perform a Hypergrid.join() here so that each feature is prefixed with either `parameter.` or `context.` It will prevent name collisions, as well as ease debugging and data analysis since these prefixes will be propagated all the way to the csv files and to JMP for @edcthayer to see. So he won't need to wonder which columns are parameters and which are context.

Also, it's plausible that context features will contain benchmark parameters, and these could easily collide with target parameters.

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
```diff
@@ -40,13 +40,15 @@ def get_optimizer_convergence_state(self):
     def get_surrogate_model_fit_state(self):
         return self.get_optimizer_convergence_state().surrogate_model_fit_state
 
-    @abstractmethod
-    def suggest(self, random=False, context=None) -> Point:
-        """Suggest the next set of parameters to try.
+    @trace()
+    def suggest(self, context_values_dataframe=None):
+        """ Returns the next best configuration to try.
```
byte-sculptor, line 45: What happened to the exception and the `@abstractmethod` decorator?

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
```diff
@@ -40,13 +40,15 @@ def get_optimizer_convergence_state(self):
...
-        :return:
+        It does so by generating num_samples_per_iteration random configurations,
+        passing them through the utility function and selecting the configuration with
+        the highest utility value.
         """
-        raise NotImplementedError("All subclasses must implement this method.")
+
 
     @abstractmethod
     def register(self, feature_values_pandas_frame, target_values_pandas_frame) -> None:
```
byte-sculptor, line 54: Would it make sense to split up the feature_values into: parameters_df and context_df? This way the user does not to do the prefixing...

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
```diff
@@ -91,10 +94,22 @@ def optimum(self, optimum_definition: OptimumDefinition = OptimumDefinition.BEST
...
             return self._best_observation_optimum(features_df=features_df, objectives_df=objectives_df)
+        elif optimum_definition == OptimumDefinition.BEST_PREDICTED_WITHIN_CONTEXT:
+            if context is None:
+                raise ValueError(f"{optimum_definition} requires context to be not None.")
+            return self._optimum_within_context(context=context)
         return self._prediction_based_optimum(features_df=features_df, optimum_definition=optimum_definition, alpha=alpha)
 
+    @trace()
+    def _optimum_within_context(self, context: pd.DataFrame):
+        return self.experiment_designer.numeric_optimizer.maximize(
```
byte-sculptor, line 110: Is this returning speculative configurations? In other words, can it return a config we never tried?

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
```diff
@@ -91,10 +94,22 @@ def optimum(self, optimum_definition: OptimumDefinition = OptimumDefinition.BEST
         if not len(features_df.index):
             raise ValueError("Can't compute optimum before registering any observations.")
 
+        if context is not None and optimum_definition != OptimumDefinition.BEST_PREDICTED_WITHIN_CONTEXT:
+                raise ValueError(f"{optimum_definition} not supported if context is provided.")
+
         if optimum_definition == OptimumDefinition.BEST_OBSERVATION:
             return self._best_observation_optimum(features_df=features_df, objectives_df=objectives_df)
+        elif optimum_definition == OptimumDefinition.BEST_PREDICTED_WITHIN_CONTEXT:
```
byte-sculptor, line 102: Can we make context orthogonal here? If the user supplies context, they get an optimum for that context, otherwise they get an optimum across all contexts?

And for the second case, should we return a config that's robust across contexts or one that was the best ever? I guess here you chose a speculative config for a given context. Should we make it explicit in the name?

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctionOptimizers/RandomSearchOptimizer.py
```diff
@@ -43,21 +43,32 @@ def __init__(
...
-        :return:
+        Returns
+        -------
+        position_of_optimum : Point
         """
+        config_values_dataframe = self.optimization_problem.parameter_space.random_dataframe(num_samples=self.optimizer_config.num_samples_per_iteration)
+        if context_values_dataframe is not None:
+            assert len(context_values_dataframe) == 1
+            config_values_dataframe['_join_key'] = 0
+            copied_context = context_values_dataframe.copy()
```
byte-sculptor, line 64: This exact same logic is needed for GlowWormSwarmOptimizer and every future utility function optimizer we write, no?

source/Mlos.Python/mlos/Optimizers/BayesianOptimizer.py
```diff
@@ -102,10 +102,14 @@ def get_all_observations(self):
         return self._feature_values_df.copy(), self._target_values_df.copy()
 
     @trace()
-    def suggest(self, random=False, context=None):
+    def suggest(self, random=False, context: Point = None):
         # TODO: pass context to the suggest method
+        if self.optimization_problem.context_space is not None:
+            if context is None:
```
bpkroth, line 108: This makes context a required parameter whenever the context space is setup.  Is that desired/necessary?
Static context may not change in between suggestions and seems unnecessary/expensive to provide each time.
Dynamic context may also be expensive to collect each time.
Maybe the answer in both cases is just a cache?

source/Mlos.Python/mlos/Optimizers/OptimizationProblem.py
```diff
@@ -99,19 +104,12 @@ def __init__(
...
-            ]
-        ).join(
-            subgrid=self.parameter_space,
-            on_external_dimension=CategoricalDimension(name="contains_parameters", values=[True])
-        ).join(
-            subgrid=self.context_space,
-            on_external_dimension=CategoricalDimension(name="contains_context", values=[True])
-        )
+        parameter_names = {parameter.name for parameter in parameter_space.root_dimensions}
+        context_names = [context.name for context in context_space.root_dimensions]
```
bpkroth, line 108: @amueller the error you're running into comes from the fact that currently the c# side doesn't provide any context space in the OptimizationProblem setup during the CreateOptimizer request.

I think some error handling here to default to an empty context space (the current method?) should solve that for now.

There's discussion from @byte-sculptor and myself on whether to require context or not already.

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctionOptimizers/RandomSearchOptimizer.py
amueller: oh sorry, I thought It had put it in the parent class, will fix.

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctionOptimizers/UtilityFunctionOptimizer.py
amueller: Sorry, what is the question? Why is this abstract?

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
amueller: Yes. I'm not entirely sure what else we should do. If the context is continuous, then we likely will not have any configurations for that context, right?
or do you mean configurations independent of context, like configurations we tried with other context? That might be another plausible method. It would still be speculative in the sense that the combination of configuration and context would be new.

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
amueller: in what setting would a user want an optimum across contexts? And you're saying you'd take the maximum over contexts? Why the max and not the minimax (i.e. the one that's the best in the worst context) or the mean?

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctionOptimizers/GlowWormSwarmOptimizer.py
amueller: we can optimize any function, not just the acquisition function. I'm now a bit less certain that we really need that. I thought directly maximizing the surrogate instead of the acquisition function would be more useful for the optimum.

source/Mlos.Python/mlos/Optimizers/BayesianOptimizer.py
amueller: I hadn't really thought about the grpc case. When using the Python interface this doesn't make a difference, right?
If we have remote agents, then there could be multiple agents with different contexts that each don't change, but we would need to keep track of which agent corresponds to which context. Which I guess is the same as a cache?

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
bpkroth: I think there are production scenarios where one may prefer to only use points that have actually been sampled.  Basically, it should be a configurable constraint on what to return.

Definitely like your other comment about returning speculative configs for new contexts though (e.g. a machine SKU you haven't tested yet, but is similar to an existing one).

source/Mlos.Python/mlos/Optimizers/BayesianOptimizer.py
bpkroth: You bring up a good point about having to map "saved context" to the appropriate "session" (agent instance).
Would we need that in a local python case as well?  (e.g. multi-process tests)

source/Mlos.Python/mlos/Optimizers/OptimizationProblem.py
bpkroth: I think for now we should have it emit a warning and continue with empty context.
Eventually, our goal is I think to automatically capture some context features (e.g. OS versions/HW specs).
We can add that in a future PR or I can try and help submit a rudimentary patch for the C# side of things to this PR if you want (just for proof of concept purposes).

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
bpkroth: One example I could think of is to use the medium/mean best across all workload types in order to encode a sensible "robust" default when the instance optimized is not known/available/working well.

source/Mlos.Python/mlos/Optimizers/BayesianOptimizer.py
amueller: Well all the information should be encoded in the context, right? And what's not distinguished by context should be shared if you connect to the same optimizer.

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
amueller: yes indeed, that might make sense. The maximum seemed just an odd choice. Doing mean/median it might make sense to use just historical data.
If we do predictions, would you want to average over only the context we've seen? So we run an optimization for each context and select the one with best mean performance?
You might also want to integrate over the whole space but then the distribution over contexts is a bit tricky, so using the empirical distribution there might make sense.

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
amueller: Looks like some merge issues maybe, let me double check.

source/Mlos.Python/mlos/Optimizers/OptimizationProblem.py
amueller: I vaguely remember that it was a bit awkward to do that but let me try again. Not sure any more what issue I ran into.

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
amueller: So we are thinking of the context as discrete? Then I might reconsider some of the assumptions.
If we look at featured derived from a workload as the context, it's unlikely we'll observe all of them, right?

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctionOptimizers/UtilityFunctionOptimizer.py
byte-sculptor: Sorry, I didn't understand why we needed the `maximize` function in the first place. After our call, I now understand.

For the benfit of other reviewers:

It's to use the UtilityFunctionOptimizer to optimize the predicted value as well as the utility function (for example to compute a speculative optimum).

Also, per our conversation, I think one way to achieve that goal is to instantiate a separate UtilityFunctionOptimizer for that purpose. This would allow us to keep the UtilityFunctionOptimizers stateful. 

Stateful optimizers can remember "good" points from previous invocations and use them to jump-start the next invocation (under the assumption that the surrogate model changes only a little between invocations).

source/Mlos.Python/mlos/Optimizers/OptimizationProblem.py
```diff
@@ -99,20 +105,47 @@ def __init__(
 
         # Fit functions / surrogate models will be fed features consisting of both context and parameters.
         # Thus, the feature space is comprised of both context and parameters.
+        has_context = self.context_space is not None
         self.feature_space = SimpleHypergrid(
             name="features",
             dimensions=[
-                CategoricalDimension(name="contains_parameters", values=[True]),
-                CategoricalDimension(name="contains_context", values=[self.context_space is not None])
+                CategoricalDimension(name="contains_context", values=[has_context])
```
amueller, line 112: I changed it to having only one join dimension, assuming we always have parameters.

source/Mlos.Python/mlos/Optimizers/OptimizationProblem.py
```diff
@@ -99,20 +105,47 @@ def __init__(
...
-            on_external_dimension=CategoricalDimension(name="contains_context", values=[True])
+            on_external_dimension=CategoricalDimension(name="contains_context", values=[has_context])
+        )
+        if has_context:
+            self.feature_space = self.feature_space.join(
+                subgrid=self.context_space,
+                on_external_dimension=CategoricalDimension(name="contains_context", values=[True])
         )
 
+    def construct_feature_dataframe(self, config_values, context_values=None, product=False):
```
amueller, line 124: Putting this into the Optimization problem seems the right place. wdyt?

source/Mlos.Python/mlos/Optimizers/OptimumDefinition.py
```diff
@@ -43,6 +43,8 @@ class OptimumDefinition(Enum):
     #
     LOWER_CONFIDENCE_BOUND_FOR_OBSERVED_CONFIG = 'prediction_lower_confidence_bound_for_observed_config'
 
+    BEST_PREDICTED_WITHIN_CONTEXT = 'best_predicted_within_context'
```
amueller, line 46: For now I left it as-is to keep the PR small-ish. We can add a different way to specify the different optima later maybe?

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
```diff
@@ -91,10 +91,23 @@ def optimum(self, optimum_definition: OptimumDefinition = OptimumDefinition.BEST
...
             return self._best_observation_optimum(features_df=features_df, objectives_df=objectives_df)
+        elif optimum_definition == OptimumDefinition.BEST_PREDICTED_WITHIN_CONTEXT:
+            if context is None:
+                raise ValueError(f"{optimum_definition} requires context to be not None.")
+            return self._optimum_within_context(context=context)
         return self._prediction_based_optimum(features_df=features_df, optimum_definition=optimum_definition, alpha=alpha)
 
+    @trace()
+    def _optimum_within_context(self, context: pd.DataFrame):
+        greedy_utility = GreedyUtilityFunction(
```
amueller, line 107: This seems a bit strange since now the ``OptimizerBase`` needs to know about ``self.surrogate_model``, but I'm not sure how else to do it (without moving the optimum dispatching into ``BayesianOptimizer``). @byte-sculptor opinions?

amueller (commented):
Thanks for all the feedback so far! Seems cleaner after discussion with @byte-sculptor

source/Mlos.Python/mlos/Optimizers/BayesianOptimizer.py
```diff
@@ -102,19 +102,27 @@ def get_all_observations(self):
...
+            assert context in self.optimization_problem.context_space
         random = random or self.num_observed_samples < self.optimizer_config.min_samples_required_for_guided_design_of_experiments
-        suggested_config = self.experiment_designer.suggest(random=random)
+        suggested_config = self.experiment_designer.suggest(random=random, context_values_dataframe=context.to_dataframe())
         assert suggested_config in self.optimization_problem.parameter_space
         return suggested_config
 
     @trace()
-    def register(self, feature_values_pandas_frame, target_values_pandas_frame):
+    def register(self, config_values_pandas_frame, target_values_pandas_frame, context_values_pandas_frame=None):
```
byte-sculptor, line 117: Could we name it parameter_values_df or something similar? I mainly mean the rename from config to parameters to keep it in line with the name of the parameter space. The same would apply to the variable names below.

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/UtilityFunction.py
```diff
@@ -13,3 +19,24 @@ class UtilityFunction(ABC):
     @abstractmethod
     def __call__(self, feature_values_pandas_frame):
         raise NotImplementedError("All subclasses must implement this.")
+
+
+class GreedyUtilityFunction(UtilityFunction):
```
byte-sculptor, line 24: There are many ways to be greedy. Could we name it PredictedValueUtilityFunction?

byte-sculptor: Also, could we have it in its own file please?

source/Mlos.Python/mlos/Optimizers/OptimizationProblem.py
```diff
@@ -99,20 +105,47 @@ def __init__(
...
+        """Construct feature value dataframe from config value and context value dataframes.
+
+        If product is True, creates a cartesian product, otherwise appends columns.
+
+        """
+        if self.context_space is not None and context_values is None:
```
byte-sculptor, line 130: Could you add parentheses to disambiguate precedence?

source/Mlos.Python/mlos/Optimizers/OptimizationProblem.py
```diff
@@ -99,20 +105,47 @@ def __init__(
...
+            raise ValueError("Context space required by optimization problem but not provided.")
+        if product is False and len(config_values) != len(context_values):
```
byte-sculptor, line 132:
```suggestion
        if (product is False) and (len(config_values.index) != len(context_values.index)):

```

source/Mlos.Python/mlos/Optimizers/OptimizationProblem.py
byte-sculptor:
```suggestion
        if (self.context_space is not None) and (context_values is None):

```

source/Mlos.Python/mlos/Optimizers/OptimizationProblem.py
byte-sculptor: It violates the Law of Demeter a little bit, but we can deal with it in some subsequent PR.

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
byte-sculptor: So this is still a speculative context. In other words, the optimizer might suggest a config that had never been tried. Can we make it explicit in the name or by supplying another parameter? 

I think it might be a useful parameter for all other definitions of optimum...

source/Mlos.Python/mlos/Optimizers/OptimumDefinition.py
byte-sculptor: Sure, but please rename it to BEST_SPECULATIVE_WITHIN_CONTEXT or something like that. This is potentially very risky and the user should be made aware.

source/Mlos.Notebooks/BayesianOptimization-Context.ipynb
```diff
@@ -0,0 +1,464 @@
...
+   "outputs": [],
+   "source": [
+    "from mlos.Optimizers.BayesianOptimizer import BayesianOptimizer, bayesian_optimizer_config_store\n",
+    "from mlos.Optimizers.RegressionModels.HomogeneousRandomForestConfigStore import homogeneous_random_forest_config_store\n",
+    "\n",
+    "from mlos.Spaces import Point\n",
+    "\n",
+    "optimizer_config = bayesian_optimizer_config_store.default\n",
+    "optimizer_config.experiment_designer_config_fraction_random_suggestions = .1\n",
+    "random_forest_config = homogeneous_random_forest_config_store.default\n",
```
byte-sculptor, line 222: A bunch of these are the default already.

source/Mlos.Notebooks/BayesianOptimization-Context.ipynb
```diff
@@ -0,0 +1,464 @@
...
+    "    input_values_df = suggested_value.to_dataframe()\n",
+    "    # we add the context to the input_values_df for the optimizer\n",
+    "    input_values_df['y'] = y.y\n",
+    "    # suggested value are dictionary-like, keys are input space parameter names\n",
+    "    # evaluate target function\n",
+    "    target_value = f(suggested_value['x'], input_values_df['y'][0])\n",
+    "    # print(suggested_value, target_value)\n",
+    "    \n",
+    "    target_values_df = pd.DataFrame({'function_value': [target_value]})    \n",
+    "    optimizer.register(input_values_df, target_values_df)\n",
```
byte-sculptor, line 287: I think this will need an update

source/Mlos.Python/mlos/Optimizers/OptimumDefinition.py
amueller: done :)

source/Mlos.Python/mlos/Optimizers/OptimizationProblem.py
amueller: where would you put it?

source/Mlos.Python/mlos/Optimizers/BayesianOptimizer.py
amueller: should I get rid of the ``pandas_frame`` everywhere if I'm doing the rename?

source/Mlos.Notebooks/BayesianOptimization-Context.ipynb
amueller: which ones? ``random_forest_config.decision_tree_regression_model_config.n_new_samples_before_refit = 1``? any other ones?

source/Mlos.Notebooks/BayesianOptimization-Context.ipynb
amueller: Sorry, apparently didn't check this in.

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/UtilityFunction.py
amueller: there's also many ways to use the predicted value ;) But sure, I'll rename / move.

amueller:
I think I addressed most of the comments, will add some tests.

amueller:
Should the interface of ``predict`` mirror that of ``register`` and have ``parameter_values`` and ``context_values``?

amueller:
Changing the interface of ``optimizer.predict`` to take ``parameter_values`` makes it inconsistent with the surrogate model interface which takes ``feature_values`` but that seems to be the best option, as I don't think we want the surrogate models to be aware of the distinction of context and parameter values.

amueller:
I'm not entirely sure what ``get_all_observations`` should return now, though. A tuple of parameters, targets and context?

amueller:
I'm not sure I understand how to fix the remaining issues. @bpkroth mentioned something here https://github.com/microsoft/MLOS/pull/60#discussion_r522381071

but to me it looks like BayesianOptimizerProxy erroring on most method calls. The methods have changed signatures but I'm not sure what that requires in terms of rewriting the grpc part. @byte-sculptor any suggestions?

amueller:
I have not changed the ``RegressionModel.predict`` interface for now. I think the current state is pretty coherent and the change is already quite big. let me know what you think :)

bpkroth:
A high level comment about the notebook, there are a few long lines that the browser wraps (at least for me).  Might be nice to wrap them ourselves so they render better.
https://github.com/microsoft/MLOS/blob/8fe84f38f061545fcef863ecc6f459e755bd69df/source/Mlos.Notebooks/BayesianOptimization-Context.ipynb

A few examples, though probably not all:
- In [8]: `# right now ...` comment
- In [12] and [14]: `surrogate_predictions = ...`

source/Mlos.Python/mlos/Grpc/BayesianOptimizerProxy.py
```diff
@@ -80,16 +80,22 @@ def compute_surrogate_model_goodness_of_fit(self):
 
     @trace()
     def suggest(self, random=False, context=None):  # pylint: disable=unused-argument
+        if context is not None:
+            raise NotImplementedError("Context not supported on remote optimizers")
```
bpkroth, line 84:
```suggestion
            raise NotImplementedError("Context not currently supported on remote optimizers")

```

source/Mlos.Python/mlos/Grpc/BayesianOptimizerProxy.py
```diff
@@ -80,16 +80,22 @@ def compute_surrogate_model_goodness_of_fit(self):
...
         )
         suggestion_response = self._optimizer_stub.Suggest(suggestion_request)
         suggested_params_dict = json.loads(suggestion_response.ParametersJsonString)
         return Point(**suggested_params_dict)
 
     @trace()
-    def register(self, feature_values_pandas_frame, target_values_pandas_frame):
+    def register(self, parameter_values_pandas_frame, target_values_pandas_frame, context_values_pandas_frame=None):
+        if context_values_pandas_frame is not None:
+            raise NotImplementedError("Context not supported on remote optimizers")
```
bpkroth, line 97:
```suggestion
            raise NotImplementedError("Context not currently supported on remote optimizers")

```

source/Mlos.Python/mlos/Grpc/BayesianOptimizerProxy.py
```diff
@@ -102,17 +108,19 @@ def register(self, feature_values_pandas_frame, target_values_pandas_frame):
         self._optimizer_stub.RegisterObservations(register_request)
 
     @trace()
-    def get_all_observations(self) -> Tuple[pd.DataFrame, pd.DataFrame]:
+    def get_all_observations(self) -> Tuple[pd.DataFrame, pd.DataFrame, pd.DataFrame]:
```
bpkroth, line 111: This is probably not the PR to do this, but a thought for later perhaps: would this be a good place for named tuples?
After some point returning a list of data potentially becomes a readability issue.

source/Mlos.Python/mlos/Optimizers/BayesianOptimizer.py
bpkroth: From other discussions, I think the intention for this PR is to only handle context for the local case, and address the remote case later on, so I think this conversation can be closed for now.

source/Mlos.Python/mlos/Optimizers/BayesianOptimizer.py
```diff
@@ -117,54 +124,82 @@ def get_optimizer_convergence_state(self):
         return self._optimizer_convergence_state
 
     def get_all_observations(self):
-        return self._feature_values_df.copy(), self._target_values_df.copy()
+        return self._parameter_values_df.copy(), self._target_values_df.copy(), self._context_values_df.copy()
 
     @trace()
-    def suggest(self, random=False, context=None):
+    def suggest(self, random=False, context: Point = None):
         # TODO: pass context to the suggest method
```
bpkroth, line 131:
```suggestion

```

This comment can be removed now?

source/Mlos.Python/mlos/Optimizers/OptimizationProblem.py
bpkroth: Makes sense.  `assert` that to help the user?

bpkroth: Also do you need to remove `contains_parameters` from the `META_DIMENSIONS_NAMES` above?

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
```diff
@@ -87,39 +90,53 @@ def optimum(self, optimum_definition: OptimumDefinition = OptimumDefinition.BEST
...
 
-        features_df, objectives_df = self.get_all_observations()
-        if not len(features_df.index):
+        parameters_df, objectives_df, _ = self.get_all_observations()
+
+        if not len(parameters_df.index):
             raise ValueError("Can't compute optimum before registering any observations.")
+        have_context = (context is not None) or (self.optimization_problem.context_space is not None)
+        if have_context and optimum_definition != OptimumDefinition.BEST_SPECULATIVE_WITHIN_CONTEXT:
+            raise ValueError(f"{optimum_definition} not supported if context is provided.")
```
bpkroth, line 99: I think this was probably already discussed so sorry if I missed/forgot it.
This will throw an error when the optimization problem includes a context space as well, not just if the `optimum()` caller passes context, right?
Is the idea that we aren't going to do anything about the question of "optimum() over all contexts" for now (or ever)?

source/Mlos.Python/mlos/Grpc/BayesianOptimizerProxy.py
byte-sculptor: Agreed.

We spoke about this with Andreas a couple of days ago. I think a full class would be best here. For once, named tuples are immutable and they can only be composed of immutable objects, so they cannot hold dataframes. Additionally, we might want to carry some metadata along too (i.e. the OptimizationProblem with the schema definitions for parameters, context, and objectives).

But like you said, it's a bigger change and we agreed to punt it to some later PR.

source/Mlos.Python/mlos/Grpc/BayesianOptimizerProxy.py
```diff
@@ -102,17 +108,19 @@ def register(self, feature_values_pandas_frame, target_values_pandas_frame):
         self._optimizer_stub.RegisterObservations(register_request)
 
     @trace()
-    def get_all_observations(self) -> Tuple[pd.DataFrame, pd.DataFrame]:
+    def get_all_observations(self) -> Tuple[pd.DataFrame, pd.DataFrame, pd.DataFrame]:
         response = self._optimizer_stub.GetAllObservations(self.optimizer_handle)
         features_df = pd.read_json(response.Features.FeaturesJsonString, orient='index')
         objectives_df = pd.read_json(response.ObjectiveValues.ObjectiveValuesJsonString, orient='index')
-        return features_df, objectives_df
+        return features_df, objectives_df, None
```
byte-sculptor, line 115:
```suggestion
        context_df = None
        return features_df, objectives_df, context_df 
```
Also, calling it features_df instead of parameters_df doesn't make sense anymore, but I guess that's too big of a change...

source/Mlos.Python/mlos/Grpc/BayesianOptimizerProxy.py
byte-sculptor: Also, if we keep all three dataframes in an object, we can frequently assert that they have the same length, which should help protect data integrity.

source/Mlos.Python/mlos/Grpc/BayesianOptimizerProxy.py
bpkroth: Is it?  Would be good to be consistent if we can.

source/Mlos.Python/mlos/Optimizers/unit_tests/TestBayesianOptimizer.py
```diff
@@ -488,15 +489,131 @@ def test_registering_multiple_objectives(self):
...
+
+        with pytest.raises(ValueError, match="Context required"):
+            local_optimizer.suggest()
+
+        with pytest.raises(ValueError, match="Context required"):
+            local_optimizer.predict(parameter_values_pandas_frame=parameter_df)
+
+        suggestion = local_optimizer.suggest(context=context_space.random())
+        assert isinstance(suggestion, Point)
+        assert input_space.contains_point(suggestion)
```
byte-sculptor, line 547:
```suggestion
        assert suggestion in input_space
```

source/Mlos.Python/mlos/Optimizers/unit_tests/TestOptimizationProblem.py
```diff
@@ -0,0 +1,81 @@
...
+
+        output_space = SimpleHypergrid(
+            name="output",
+            dimensions=[
+                ContinuousDimension(name='y', min=-math.inf, max=math.inf)
+            ]
+        )
+        optimization_problem = OptimizationProblem(
+            parameter_space=objective_function.parameter_space,
+            objective_space=output_space,
```
byte-sculptor, line 71:
```suggestion
            objective_space=objective_function.output_space,
```

source/Mlos.Python/mlos/Optimizers/unit_tests/TestOptimizationProblem.py
```diff
@@ -0,0 +1,81 @@
...
+            objectives=[Objective(name='y', minimize=True)]
+        )
+        n_samples = 100
+        parameter_df = optimization_problem.parameter_space.random_dataframe(n_samples)
+        feature_df = optimization_problem.construct_feature_dataframe(parameter_values=parameter_df)
+        assert feature_df.shape == (n_samples, len(optimization_problem.parameter_space.dimension_names) + 1)
+        expected_columns = sorted([f"three_level_quadratic_config.{n}" for n in optimization_problem.parameter_space.dimension_names])
+        assert (feature_df.columns[:-1].sort_values() == expected_columns).all()
+        assert feature_df.columns[-1] == "contains_context"
+        assert not feature_df.contains_context.any()
```
byte-sculptor, line 81:
```suggestion
        assert not feature_df.contains_context.any()
        
```

source/Mlos.Python/mlos/Grpc/BayesianOptimizerProxy.py
amueller: Hm you can have named tuples of mutable data I think, but I agree a full object might be nicer.

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
amueller: yes exactly. Sorry I discussed that in the 1:1 with Adam. We concluded that doing max makes little sense (i.e. it would always find the configuration that works best on the easiest context, like the biggest machine or smallest workload).
And adding some more robust thing is left for a future PR.

source/Mlos.Python/mlos/Grpc/BayesianOptimizerProxy.py
amueller: Basically I didn't want to change the messages and I left the names here to be consistent with the messages.
Changing the messages is a bigger change.

amueller:
@bpkroth so re wrap-around, I generally run everything in jupyter lab and encourage everybody to do so ;)
@byte-sculptor usually includes some CSS to change the rendering. But I can add some page-brakes for the github rendering.

source/Mlos.Python/mlos/Grpc/OptimizerMicroservice.py
```diff
@@ -133,8 +133,10 @@ def Suggest(self, request, context): # pylint: disable=unused-argument
 
         # TODO: return an error if optimizer not found
         #
+        if request.Context.ContextJsonString != "":
+            raise NotImplementedError("context not supported in remote optimizers")
         with self.exclusive_optimizer(optimizer_id=request.OptimizerHandle.Id) as optimizer:
-            suggested_params = optimizer.suggest(random=request.Random, context=request.Context)
```
bpkroth, line 137: Should we add a `# TODO` comment on adding support for handling `Context` in the future.

byte-sculptor (approved):

source/Mlos.Python/mlos/Grpc/OptimizerMicroservice.py
amueller: Done. I think my thought was that the errors are implicit TODOs, we clearly want to support that soon-ish.

bpkroth (approved):

bpkroth:
> @bpkroth so re wrap-around, I generally run everything in jupyter lab and encourage everybody to do so ;)

Sure.  I was just thinking some amount of these will probably be skimmed on github.com or github.io.  Would be nice if they rendered well there too, but not a hard requirement.

PR #66, opened 2020-09-10
Increase python unit test CI timeout

bpkroth:
I'm still seeing a few timeout failures on the unit tests.

This is a quick change to simply increase the timeout for now.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -108,10 +108,12 @@ jobs:
         BRANCH: gh-pages
         FOLDER: website/public
 
+  # FIXME: These timeouts are egregiously long right now.
+  # See Also: https://github.com/microsoft/MLOS/pull/66#issuecomment-690760863
   python-checks:
     name: Run Python checks
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 15
+    timeout-minutes: 90
     strategy:
       matrix:
         os: [windows-2019, ubuntu-20.04]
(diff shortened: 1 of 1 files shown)


bpkroth:
Hmm, this still failed.  Something else might be going on than the Python unit tests just randomly taking a long time due to high-degree polynomials being chosen (which I think we already reduced).

amueller (approved):

amueller (changes requested):
still timing out :-/

bpkroth:
We've done some digging on this, and it appears that when executed on Github, several of these models simply take much longer.

In particular:
- test_random_search_optimizer (2x local)
- test_random_decision_tree_models (100x local)
- test_lasso_feature_discovery (29x local)
- test_polynomial_gradient (12x local)

Not clear why that is yet.

bpkroth:
> 
> 
> We've done some digging on this, and it appears that when executed on Github, several of these models simply take much longer.
> 
> In particular:
> 
>     * test_random_search_optimizer (2x local)
> 
>     * test_random_decision_tree_models (100x local)
> 
>     * test_lasso_feature_discovery (29x local)
> 
>     * test_polynomial_gradient (12x local)
> 
> 
> Not clear why that is yet.

We may simply be hitting up against a resource limit.
https://docs.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources

For now, I'm going to increase the limit substantially to unblock us and we can revisit this later.  Possibly we need to point them at some internal resources with more powerful runners.

amueller (approved):

PR #69, opened 2020-09-10
Attempt of a logo

curino:
Just a quick attempt using (properly licensed icons) of a logo. (we haven't discuss if we even need one, so ok to pass on this or the idea of a logo at all).


bpkroth:
Love the idea!  Thanks for getting started on this.

bpkroth:
One thought: more colors

PR #70, opened 2020-09-10
Fixup documentation build scripts

bpkroth:
Some minor tweaks following #61

### DIFF
--- a/website/Makefile
+++ b/website/Makefile
@@ -40,6 +40,7 @@ public/index.html:
 clean-sphinx-site:
 	rm -rf sphinx/_build
 	rm -rf sphinx/api
+	git restore sphinx/api/mlos.rst
 	rm -rf public/python_api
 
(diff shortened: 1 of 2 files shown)


website/Makefile
```diff
@@ -40,6 +40,7 @@ public/index.html:
 clean-sphinx-site:
 	rm -rf sphinx/_build
 	rm -rf sphinx/api
+	git restore sphinx/api/mlos.rst
```
bpkroth, line 43: Kind of a hack, but it works ...

amueller: I mean the whole sphinx business is a bit hacky right now and I take all the blame :-/

amueller (approved):

website/Makefile
bpkroth: I added to it, so I think we can share that blame :)

Issue #72, opened 2020-09-10
Move to pytest


amueller:
we should use pytest as test runner

bpkroth:
For the record, one of the goals here is to be able to get some better stats on how long tests take.
See Also #67

byte-sculptor:
I'm still educating myself about the relative merits of these two frameworks. What features of pytest particularly take your fancy? I know you mentioned parameterizable tests in the past...

amueller:
parametrizable tests, better error messages, plugins for coverage and some other plugins.

pytest is a bit magic, it parses the AST to figure out a way to give you a good error, so if you write
```python
b = 5
assert b == 6
```
It tells you "assert failed 5 != 6", while unittest or just plain python would say "AssertionError(False)".

I don't recall seeing a package not using pytest recently. There used to be nosetests, but that is no longer maintained and most packages moved from nosetests to pytest.
Examples of packages that use pytest are numpy, pandas, sklearn, requests, hummingbird.

bpkroth:
The other reason we were discussing this was in the course of troubleshooting #67  and trying to figure out which tests were taking a long time.

I had written this nasty oneliner to try and identify the seconds elapsed that each test took:
```
last_ts=''; ./scripts/run-python-tests.sh | logger -s -t python-test 2>&1 | grep 'python-test: test_' | while read line; do echo "$line"; ts=$(echo "$line" | egrep -o '^<13>Sep 10 [0-9:]+ ' | sed 's/^<13>//' | xargs -I{} date -d"{}" +%s); if [ -n "$last_ts" ]; then let time_took=ts-last_ts; echo $time_took; fi; last_ts=$ts; done | egrep -B1 '^[1-9][0-9]+'
<13>Sep 10 21:01:39 python-test: test_random_search_optimizer (mlos.Optimizers.ExperimentDesigner.NumericOptimizers.unit_tests.TestRandomSearchOptimizer.TestRandomSearchOptimizer) ... ok
85
--
<13>Sep 10 21:01:50 python-test: test_random_function_configs (mlos.Optimizers.ExperimentDesigner.UtilityFunctions.unit_tests.TestConfidenceBoundUtilityFunction.TestConfidenceBoundUtilityFunction) ... ok
11
--
[...]

PR #73, opened 2020-09-10
various python related tweaks

bpkroth:
A collection of minor python related tweaks while working on #66.

### DIFF
--- a/scripts/run-python-tests.sh
+++ b/scripts/run-python-tests.sh
@@ -15,4 +15,4 @@ pythoncmd=`getPythonCmd`
 
 # Linux filesystems are case-sensitive, so we need to tell the python unit test
 # scanner to look for upper-case files as well:
-$pythoncmd -m unittest discover --verbose --locals --failfast -s source/Mlos.Python -p "[Tt]est*.py"
+$pythoncmd -m unittest discover --verbose --locals --failfast -s source/Mlos.Python -p "[Tt]est*.py" $*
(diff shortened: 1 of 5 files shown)


scripts/run-python-tests.sh
```diff
@@ -15,4 +15,4 @@ pythoncmd=`getPythonCmd`
 
 # Linux filesystems are case-sensitive, so we need to tell the python unit test
 # scanner to look for upper-case files as well:
-$pythoncmd -m unittest discover --verbose --locals --failfast -s source/Mlos.Python -p "[Tt]est*.py"
+$pythoncmd -m unittest discover --verbose --locals --failfast -s source/Mlos.Python -p "[Tt]est*.py" $*
```
amueller, line 18: I don't actually know what that does.

amueller (approved):
My bash scripts aren't entirely on your level but looks good :)

scripts/run-python-tests.sh
bpkroth: The explanation was in the commit message, but I suppose that isn't as helpful.
https://github.com/microsoft/MLOS/pull/73/commits/93eeb691b491aa33af9a7b65873f77f5a7ccb815

Basically, if you call:
```sh
./run-python-test.sh foo bar
```

Then `$*` evaluates as `foo bar`.

The main purpose was to allow more quickly rerunning just the failed unit test with `-k some_test_name` appended afterwards to more quickly iterate on fixes.

Should I document that, or do you want to leave it for a `pytest` conversion fixup (#72)?

amueller: ah, cool. I think this is fine for now and we can leave it for later.

Issue #71, opened 2020-09-10
Can't call optimum from BayesianOptimizerProxy


amueller:
currently the ``optimum`` implementation of BayesianOptimizerProxy is empty.

amueller:
fixed in #76.

PR #76, opened 2020-09-14
add base implementation of optimum in OptimizerInterface

amueller:
This adds an implementation of "optimum" to the BayesianOptimizerProxy based on the public ``get_all_observations``.

I'm not sure how @byte-sculptor feels about adding implementations to the interface class. This implementation could probably be shared by all derived classes so this felt like a natural place.

Another concern with the current trivial implementation is that right now it adds a copy of the data in ``BayesianOptimizer``. That could be easily addressed by adding a ``allow_view=False`` argument to ``get_all_observations`` to not make the copy unless it's required.

Obviously still missing tests


bpkroth:
> 
> 
> This adds an implementation of "optimum" to the BayesianOptimizerProxy based on the public `get_all_observations`.
> 
> I'm not sure how @byte-sculptor feels about adding implementations to the interface class. This implementation could probably be shared by all derived classes so this felt like a natural place.

Depends on whether the language will accept it, but I'm generally okay with that.
Other option is to define a base class that by convention everything inherits from to provide your default implementations, but that seems like unnecessary extra classes to me, so avoid it if you can?

> Another concern with the current trivial implementation is that right now it adds a copy of the data in `BayesianOptimizer`. That could be easily addressed by adding a `allow_view=False` argument to `get_all_observations` to not make the copy unless it's required.

:thumbsup:

Only question I have is what the default should be.  How often do you need a copy, and what are the potential damages by not always returning a copy by default?

> Obviously still missing tests

bpkroth:
For the record, I believe this is to address issue #71, which one of the students reported.

byte-sculptor:
Looks, good. I think this class is the best place for the implementation, though it should never have been named an OptimizerInterface. So if it's not too much work to rename the class to OptimzierBase, I think that would be a nice improvement.

amueller:
> Only question I have is what the default should be. How often do you need a copy, and what are the potential damages by not always returning a copy by default?

I think the potential danger is having the copy overhead. I assume that to be negligible, based on no data at all. If you call "optimum" in a tight loop then the copy might become a bottleneck, but I would call the premature optimization.

bpkroth:
> 
> 
> > Only question I have is what the default should be. How often do you need a copy, and what are the potential damages by not always returning a copy by default?
> 
> I think the potential danger is having the copy overhead. I assume that to be negligible, based on no data at all. If you call "optimum" in a tight loop then the copy might become a bottleneck, but I would call the premature optimization.

Oh, my fear was the other way.  By returning the reference, not a copy, then you might risk unintended modification to the internals.

I think the risk of it being slow would be somehow more obvious and benign (it just takes longer, not we've accidentally altered the internal data which is harder to track down).

amueller:
Oh yeah I totally agree, the overwriting is the much bigger risk. The current implementation basically requires ``get_all_observations`` to return a copy which should be super safe, though the responsibility of being safe is on ``get_all_observations`` (which should be reasonably easy to test).

source/Mlos.Python/mlos/Optimizers/SimpleBayesianOptimizer.py
```diff
@@ -300,18 +300,6 @@ def estimate_local_parameter_importance(self, params, t=None):
 
         return local_parameter_importance
 
-    def optimum(self, stay_focused=False):
```
amueller, line 303: I removed some functionality here but this class will be removed anyway from what I understand.

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
```diff
@@ -56,15 +56,29 @@ def predict(self, feature_values_pandas_frame, t=None) -> Prediction:
         """
         raise NotImplementedError("All subclasses must implement this method.")
 
-    @abstractmethod
-    def optimum(self, stay_focused=False) -> Dict: # TODO: make it return an object
```
bpkroth, line 60: Is the "TODO: make it return an object" comment still worth keeping?

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
```diff
@@ -56,15 +56,29 @@ def predict(self, feature_values_pandas_frame, t=None) -> Prediction:
...
 
         This could be either min or max, depending on the settings.
 
-        :return:
+        Returns
+        -------
+        best_config_point : Point
+            Configuration that corresponds to the optimum objective value.
+        best_objective : Point
+            Best objective value observed so far.
```
bpkroth, line 69:
```suggestion
        best_objective_point : Point
            Best objective value observed so far (corresponding to the best_config_point).

```

bpkroth, line 69:
```suggestion
        Best objective value observed so far (corresponding to the best_config_point).

```

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
```diff
@@ -56,15 +56,29 @@ def predict(self, feature_values_pandas_frame, t=None) -> Prediction:
...
         """ Return the optimal value found so far along with the related parameter values.
 
         This could be either min or max, depending on the settings.
 
-        :return:
+        Returns
+        -------
+        best_config_point : Point
+            Configuration that corresponds to the optimum objective value.
+        best_objective : Point
```
bpkroth, line 68:
```suggestion
        best_objective_point : Point

```

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
amueller: Hm probably not as it's now returning two points?

source/Mlos.Python/mlos/Optimizers/long_haul_tests/LongHaulTestBayesianOptimizer.py
```diff
@@ -167,9 +166,9 @@ def test_bayesian_optimizer_on_simple_2d_quadratic_function_cold_start(self):
 
             bayesian_optimizer.register(input_values_df, target_values_df)
             if i > 20 and i % 20 == 0:
-                self.logger.info(f"[{i}/{num_guided_samples}] Optimum: {bayesian_optimizer.optimum()}")
+                self.logger.info(f"[{i}/{num_guided_samples}] Optimum: {bayesian_optimizer.optimum()[1]}")
 
-        self.logger.info(f"Optimum: {bayesian_optimizer.optimum()}")
+        self.logger.info(f"Optimum: {bayesian_optimizer.optimum()[1]}")
```
bpkroth, line 171: `[1]` seems like the more common value to be asking about, maybe it should be returned first?
Also, named values/properties seems maybe better.

source/Mlos.Python/mlos/unit_tests/TestBayesianOptimizerGrpcClient.py
```diff
@@ -177,6 +179,8 @@ def optimize_quadratic(self, optimizer, num_iterations):
             prediction = optimizer.predict(features_df)
             prediction_df = prediction.get_dataframe()
 
+
+
             y = quadratic(**params_dict)
```
bpkroth, line 184: Nit: extra whitespace

.gitignore
```diff
@@ -428,3 +428,6 @@ website/themes
 website/sphinx/_build
 website/sphinx/api
 website/python_api
+
+# python code coverage
+coverage
```
bpkroth, line 433: does that need to go in the top level?

bpkroth: Actually, seems unrelated for this PR, no?

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
bpkroth: I noted down below that it may make more sense to return named params as well, but that can certainly get pushed off to a different PR.

.gitignore
amueller: depends? Not necessarily, that's the default if this is where we run the python tests from.

source/Mlos.Python/mlos/Optimizers/long_haul_tests/LongHaulTestBayesianOptimizer.py
amueller: having configuration, optimum makes more sense to me, and I think actually the configuration is the more interesting one, but it's not used in the tests right now because it might not have been available before?

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
```diff
@@ -11,7 +11,7 @@
 from mlos.Optimizers.RegressionModels.Prediction import Prediction
 from mlos.Spaces import Point
 
-class OptimizerInterface(ABC):
+class OptimizerBase(ABC):
     """ Defines the interface to all our optimizers.
```
byte-sculptor, line 15: Defines the base class for all our optimizers.

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
byte-sculptor: I think you are addressing the TODO with this PR so I think it's good that you changed it. I would also change the type hint of the return value to: ```-> Tuple[Point, Point]```

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
```diff
@@ -56,15 +56,31 @@ def predict(self, feature_values_pandas_frame, t=None) -> Prediction:
...
+        -------
+        best_config_point : Point
+            Configuration that corresponds to the optimum objective value.
+        best_objective_point : Point
+
+            Best objective value observed so far (corresponding to the best_config_point).
+
         """
-        raise NotImplementedError("All subclasses must implement this method.")
+        features_df, objectives_df = self.get_all_observations()
```
byte-sculptor, line 73: What if the user calls optimum before registering any observations?

source/Mlos.Python/mlos/Optimizers/SimpleBayesianOptimizer.py
byte-sculptor: Yeah, I wouldn't worry about it. The more we break it before deleting it, the less likely it is that someone will take a dependency on it :D

source/Mlos.Python/mlos/Optimizers/long_haul_tests/LongHaulTestBayesianOptimizer.py
```diff
@@ -115,8 +115,7 @@ def test_bayesian_optimizer_on_simple_2d_quadratic_function_pre_heated(self):
 
             # Register the observation with the optimizer
             bayesian_optimizer.register(input_values_df, target_values_df)
-
-        self.logger.info(f"Optimum: {bayesian_optimizer.optimum()}")
+        self.logger.info(f"Optimum: {bayesian_optimizer.optimum()[1]}")
```
byte-sculptor, line 118: Could we please change it to:
```
best_config_point, best_objective = bayesian_optimizer.optimum()
self.logger.info(f"Optimum: {best_objective}")
```
Unpacking the tuple this way makes it clear to the reader what's going on. It also creates a neat pattern for people to copy-paste in their own client code.

source/Mlos.Python/mlos/Optimizers/long_haul_tests/LongHaulTestBayesianOptimizer.py
```diff
@@ -167,9 +166,9 @@ def test_bayesian_optimizer_on_simple_2d_quadratic_function_cold_start(self):
 
             bayesian_optimizer.register(input_values_df, target_values_df)
             if i > 20 and i % 20 == 0:
-                self.logger.info(f"[{i}/{num_guided_samples}] Optimum: {bayesian_optimizer.optimum()}")
+                self.logger.info(f"[{i}/{num_guided_samples}] Optimum: {bayesian_optimizer.optimum()[1]}")
```
byte-sculptor, line 169: same as above

source/Mlos.Python/mlos/Optimizers/long_haul_tests/LongHaulTestBayesianOptimizer.py
byte-sculptor: We can always log both :)

source/Mlos.Python/mlos/Optimizers/long_haul_tests/LongHaulTestBayesianOptimizer.py
byte-sculptor: or even:

```
best_config_point, best_objective = bayesian_optimizer.optimum()
self.logger.info(f"Optimum config: {best_config_point}, optimum objevtive: {best_objective}")
```

amueller: the linter will complain about the unused variable, though, right?
I had it first as 
```python
_, best_objective = bayesian_optimizer.optimum()
```
but I'm not sure that's more useful?

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
amueller: It errors awkwardly right now. I can make it throw a nicer error?

source/Mlos.Python/mlos/Optimizers/long_haul_tests/LongHaulTestBayesianOptimizer.py
byte-sculptor: Also, are there any invariants we can assert about the optima in this test and in the tests below? Specifically:
1. That it is in fact the optimum
2. That the config belongs to the parameter space
3. That the optimum belongs to the objective space
4. Anything else you can think of

source/Mlos.Python/mlos/unit_tests/TestBayesianOptimizerGrpcClient.py
```diff
@@ -192,4 +194,10 @@ def optimize_quadratic(self, optimizer, num_iterations):
                 registered_objectives_df = objectives_df
             else:
                 registered_objectives_df = registered_objectives_df.append(objectives_df, ignore_index=True)
+
+            best_params, optimum = optimizer.optimum()
+            # ensure current optimum doesn't go up
+            assert optimum.y <= old_optimum
```
byte-sculptor, line 200: self.assertTrue(optimum.y <= old_optimum)

amueller: Is that different? If we move to pytest, the current version will give a nicer error afaik.

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
amueller: throwing a nicer error now.

source/Mlos.Python/mlos/Optimizers/long_haul_tests/LongHaulTestBayesianOptimizer.py
amueller: I'm happy to add some of these but I feel that's mostly unrelated to this PR, right? More testing is always good but I'd love to push this to the students.

source/Mlos.Python/mlos/unit_tests/TestBayesianOptimizerGrpcClient.py
amueller: The unittest way to get the same message would be ``self.assertLessEqual(optimum.y, old_optimum)``

.gitignore
bpkroth: k, cool

source/Mlos.Python/mlos/Optimizers/long_haul_tests/LongHaulTestBayesianOptimizer.py
```diff
@@ -322,5 +329,5 @@ def run_optimization(optimizer):
 
             for _ in range(40):
                 run_optimization(optimizer)
-            print(optimizer.optimum()['function_value'])
-            self.assertLessEqual(sign * optimizer.optimum()['function_value'], -5.5)
+            print(optimizer.optimum()[1]['function_value'])
```
bpkroth, line 332: Do you want to replace these too?

amueller:
had to resolve some merge conflicts so please check again.

source/Mlos.Python/mlos/Optimizers/unit_tests/TestBayesianOptimizer.py
```diff
@@ -167,7 +193,14 @@ def test_bayesian_optimizer_on_simple_2d_quadratic_function_cold_start(self):
 
             bayesian_optimizer.register(input_values_df, target_values_df)
             if i > optimizer_config.min_samples_required_for_guided_design_of_experiments and i % 10 == 1:
-                print(f"[{i}/{num_iterations}] Optimum: {bayesian_optimizer.optimum()}")
+                _, all_targets = bayesian_optimizer.get_all_observations()
+                best_config, optimum = bayesian_optimizer.optimum()
+                print(f"[{i}/{num_iterations}] Optimum: {optimum}")
+                assert optimum.y == all_targets.min()[0]
+                assert input_space.contains_point(best_config)
```
byte-sculptor, line 200: Is it just awkward to use self.assertTrue(), is it conflicting with the upcoming switch to pytest or is it just habit? Either way it will throw, but I'm curious if there are any reasons not to use the unittest's built-ins?

byte-sculptor: Just saw your comment below.

byte-sculptor, line 200: Just for the future, the __contains__ operator is overriden, so we can do:
```
best_config in input_space
```

amueller: Thanks, that's cool!
Also, for one-line fixes, github allows you to suggest so I can immediately commit your suggestion :)

Do you want me to change this here and/or happy to approve otherwise?

byte-sculptor (approved):

bpkroth (approved):
LGTM.  Thanks!

PR #79, opened 2020-09-16
reorganize and update conda/python install docs

amueller:
cc @bpkroth @markusweimer 

Not sure if I'm undoing some of the work @markusweimer has been doing here before.
I grouped the OS independent steps in the beginning.

I think the "manual python" version is not that useful because it's ubuntu only (not sure if the ppas work for debian?), and there's really no reason to compile your own version of pyodbc, so I removed it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,58 +66,10 @@ To achieve this MLOS provides:
 
     Once hooks are created in the target system, iteration on the external agent can be more rapidly developed and deployed.
 
-## Python only installation
+## Python quickstart
+The easiest way to get started with MLOS is to just just the Python package.
+You can find installation instructions in the [Prerequisites: Python Quickstart](./documentation/01-Prerequisites.md#python-quickstart).
 
-Some of the examples require only the installation of the `mlos` Python library.
-These are simplified examples for getting started and do not use the shared memory infrastructure for cross language/process interactions.
-
-First, download the MLOS code using git:
-
-```shell
-git clone https://github.com/microsoft/MLOS.git
-cd MLOS
-```
-
-For this simplified installation, it's recommended to use the [Anaconda python distribution](https://www.anaconda.com/products/individual).
-For a slimmer installation experience, you can also use the [miniconda installer](https://docs.conda.io/en/latest/miniconda.html).
-After installing either anaconda or miniconda, you can create a new environment with all requirements for the examples using
-
-> Note currently MLOS needs Python 3.7, but by default conda packages Python 3.8.
-> For the moment, you may need to search the [installer archives](https://repo.anaconda.com/miniconda/) for a python 3.7 version.
-> e.g. [Miniconda3-py37_4.8.3-Windows-x86_64.exe](https://repo.anaconda.com/miniconda/Miniconda3-py37_4.8.3-Windows-x86_64.exe)
-
-```shell
-conda env create -f MLOS/source/Mlos.Notebooks/environment.yml
-```
-
-The environment will be called `mlos_python_environment` and you can activate it as follows:
-
-```shell
-conda activate mlos_python_environment
-```
-
-Use `pip` to install the Python library:
-
-```shell
-pip install MLOS/source/Mlos.Python/
-```
-
-Alternatively, you can also install the package directly without checking out the code:
-
-```shell
-pip install "git+https://github.com/microsoft/MLOS.git#egg=mlos&subdirectory=source/Mlos.Python"
-```
-
-However, this does not include the examples and requires you to set up your environment manually.
-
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -66,58 +66,10 @@ To achieve this MLOS provides:
 
     Once hooks are created in the target system, iteration on the external agent can be more rapidly developed and deployed.
 
-## Python only installation
+## Python quickstart
+The easiest way to get started with MLOS is to just just the Python package.
```
bpkroth, line 70:
```suggestion
The easiest way to get started with MLOS is to just the Python package.
```

documentation/01-Prerequisites.md
```diff
@@ -6,26 +6,67 @@ These are one-time setup instructions that should be executed prior to following
...
+pip install MLOS/source/Mlos.Python/
+```
+
+After this installation, you can run any of the Python-only example notebooks. To do so you can:
+
+```shell
+jupyter-notebook --notebook-dir=MLOS/source/Mlos.Notebooks
+```
+
+Jupyter will list a few notebooks. A good place to start is the *BayesianOptimization.ipynb*, which provides an [Introduction to Bayesian Optimization](./source/Mlos.Notebooks/BayesianOptimization.ipynb#mlos-github-tree-view).
```
bpkroth, line 68:
```suggestion
Jupyter will list a few notebooks. A good place to start is the *BayesianOptimization.ipynb*, which provides an [Introduction to Bayesian Optimization](../source/Mlos.Notebooks/BayesianOptimization.ipynb#mlos-github-tree-view).
```

bpkroth: I think this should fix the docs build error.

bpkroth (approved):
Looks reasonable to me.  Just two minor issues to fix that I saw.

Are we abandoning any support for straight pip requirements support, or just relegating it to the advanced users that want to peer under the hood (e.g. inside the Dockerfile or some such).
Should we make an issue to track fixing up our build pipelines to do the same?

bpkroth:
> 
> 
> cc @bpkroth @markusweimer
> 
> Not sure if I'm undoing some of the work @markusweimer has been doing here before.
> I grouped the OS independent steps in the beginning.
> 
> I think the "manual python" version is not that useful because it's ubuntu only (not sure if the ppas work for debian?), and there's really no reason to compile your own version of pyodbc, so I removed it.

re: Debian, no (and it pains me to say so), I don't think they work on pure Debian, but that has more to do with our current need to pin to certain older versions than the pip stuff.

amueller:
I think delegating the pip installation to more experienced users is fine. Most python folks will use conda. And if they want to use pip, the requirements for installation are actually all in setup.py, and giving full instructions on how to build the requirements is quite hard (CI just failed the build because it couldn't compile parts of matplotlib, which requires a *lot* of stuff to work properly).

The way it's set up, someone that knows a bit about how python packaging works shouldn't really need any instructions as the standard "pip install the folder with the setup.py" just does the trick if your environment is "reasonable".

amueller:
ohh it's failing because you're parsing the environment.yml and installing it with pip which obviously fails for conda packages...

amueller:
closes #31.

amueller:
Do we have an issue to move the CI to conda?

scripts/parse-pip-requirements-from-environment-yaml.py
```diff
@@ -29,7 +29,9 @@
             pattern=r'([^=])=([^=])',
             repl='\\1==\\2',
             string=item)
-        new_dependencies.append(new_item)
+        if "conda" not in new_item and not new_item.startswith("python"):
```
bpkroth, line 32: Was that not handled with the `excluded_dependencies` below?  Oh, I see, it's now no longer an exact match.  Maybe move it down there just to consolidate?

Honestly this whole ~script~ hack can probably just go away if we decide that we're consolidating on some form of conda, but let's leave that for a future PR.

bpkroth:
> 
> 
> Do we have an issue to move the CI to conda?

We do now :) #83

scripts/parse-pip-requirements-from-environment-yaml.py
amueller: do you want me to consolidate or leave it for moving to conda?

bpkroth: My suggestion for now is to just switch to something like this below for now:

```
-    if item in excluded_dependencies:
+    dep_name = re.split('[><=]+', item)[0]
+    if dep_name in excluded_dependencies:
```

(tested locally, seems to work as expected)

Issue #54, opened 2020-09-04
Improve mlos library to support Python 3.8


bpkroth:
Several of the modern tool installers default to Python 3.8.  We should see what can be done to make the mlos python library work for that as well rather than pinning on 3.7 which complicates the install/setup process.

PR #81, opened 2020-09-16
stalled: enable python ci on 3.8

amueller:
try to run tests


bpkroth:
My recollection is that this fails the pylint checks first.

Also: related #54

amueller:
Looks like installing the pip dependencies fails with a compilation error. Might be able to solve that with apt-getting some more libs, or we can move to conda.

amueller:
This requires some changes from #79 so lets wait for that.

Issue #80, opened 2020-09-16
Internal references broken on hugo website


amueller:
Clicking a link in the Contents section of Prerequisites sends you to a different page:
https://microsoft.github.io/MLOS/documentation/01-Prerequisites/

PR #82, opened 2020-09-16
Fix and check for bad anchor links in docs as well

bpkroth:
Addresses #80

### DIFF
--- a/website/build_site.sh
+++ b/website/build_site.sh
@@ -88,7 +88,17 @@ for content_filepath in $(find content/ -type f -name '*.md'); do
     fi
 
     base_filepath=$(echo "$content_filepath" | sed 's|^content/||')
+
     parent_path=$(dirname "$base_filepath")
+    if [ "$parent_path" == '.' ]; then
+        parent_path=''
+    elif [ -n "$parent_path" ]; then
+        parent_path="${parent_path}/"
+    fi
+
+    echo "content_filepath: $content_filepath"
+    echo "base_filepath: $base_filepath"
+    echo "parent_path: $parent_path"
 
     # 1. replace a special fake anchor with a link to the main github repo site
     # (this allows browsing back to the main published source from the website)
@@ -100,11 +110,9 @@ for content_filepath in $(find content/ -type f -name '*.md'); do
     # Also keep a backup for comparison/debugging purposes.
 
     sed -i.bak -r \
-        -e "s|\]\(([./]*[^:#)]+)#mlos-github-tree-view\)|](https://github.com/microsoft/MLOS/tree/main/${parent_path}/\1)|g" \
-        -e "s|\]\(([./]*[^:#)]+)(#[^)]*)?\)|](/MLOS/${parent_path}/\1\2)|g" \
+        -e "s|\]\(([./]*[^:#)]+)#mlos-github-tree-view\)|](https://github.com/microsoft/MLOS/tree/main/${parent_path}\1)|g" \
+        -e "s|\]\(([./]*[^:#)]+)(#[^)]*)?\)|](/MLOS/${parent_path}\1\2)|g" \
         -e "s|\]\(([./]*[^:#)]+)\.md(#[^)]*)?\)|](\1/\2)|g" \
-        -e "s|\]\(([^)]*)/\./([^)]*)\)|](\1/\2)|g" \
-        -e "s|\]\(([^)]*)/\./([^)]*)\)|](\1/\2)|g" \
         -e "s|\]\(([^)]*)/\./([^)]*)\)|](\1/\2)|g" \
         "$content_filepath"
 done
--- a/website/build_site.sh
+++ b/website/build_site.sh
@@ -95,13 +95,17 @@ for content_filepath in $(find content/ -type f -name '*.md'); do
     # 2. convert relative paths to be relative to the website root instead
     # 3. strip the .md file ending to replace it with a trailing slash
     # (to be consistent with the hugo md -> html conversion)
+    # 4. Strip any unnecessary '/./' path components we introduce by doing that.
 
     # Also keep a backup for comparison/debugging purposes.
 
     sed -i.bak -r \
         -e "s|\]\(([./]*[^:#)]+)#mlos-github-tree-view\)|](https://github.com/microsoft/MLOS/tree/main/${parent_path}/\1)|g" \
-        -e "s|\]\(([./]*[^:)]+)\)|](/MLOS/${parent_path}/\1)|g" \
-        -e "s|\]\(([./]*[^:#)]+)\.md(#[^)]+)?\)|](\1/\2)|g" \
+        -e "s|\]\(([./]*[^:#)]+)(#[^)]*)?\)|](/MLOS/${parent_path}/\1\2)|g" \
+        -e "s|\]\(([./]*[^:#)]+)\.md(#[^)]*)?\)|](\1/\2)|g" \
+        -e "s|\]\(([^)]*)/\./([^)]*)\)|](\1/\2)|g" \
+        -e "s|\]\(([^)]*)/\./([^)]*)\)|](\1/\2)|g" \
+        -e "s|\]\(([^)]*)/\./([^)]*)\)|](\1/\2)|g" \
         "$content_filepath"
 done
 
--- a/website/sphinx/apidoc.sh
+++ b/website/sphinx/apidoc.sh
@@ -21,6 +21,9 @@ echo "Installing dependencies for generating Python API docs"
 $pythonCmd -m pip install -e $MLOS_ROOT/source/Mlos.Python/
 $pythonCmd -m pip install sphinx sphinx_rtd_theme numpydoc matplotlib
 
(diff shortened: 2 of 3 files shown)


website/build_site.sh
```diff
@@ -95,13 +95,17 @@ for content_filepath in $(find content/ -type f -name '*.md'); do
...
+    # 4. Strip any unnecessary '/./' path components we introduce by doing that.
 
     # Also keep a backup for comparison/debugging purposes.
 
     sed -i.bak -r \
         -e "s|\]\(([./]*[^:#)]+)#mlos-github-tree-view\)|](https://github.com/microsoft/MLOS/tree/main/${parent_path}/\1)|g" \
-        -e "s|\]\(([./]*[^:)]+)\)|](/MLOS/${parent_path}/\1)|g" \
-        -e "s|\]\(([./]*[^:#)]+)\.md(#[^)]+)?\)|](\1/\2)|g" \
+        -e "s|\]\(([./]*[^:#)]+)(#[^)]*)?\)|](/MLOS/${parent_path}/\1\2)|g" \
+        -e "s|\]\(([./]*[^:#)]+)\.md(#[^)]*)?\)|](\1/\2)|g" \
```
bpkroth, line 105: Let's trailing anchors be optional for that match.

website/build_site.sh
```diff
@@ -95,13 +95,17 @@ for content_filepath in $(find content/ -type f -name '*.md'); do
...
     # (to be consistent with the hugo md -> html conversion)
+    # 4. Strip any unnecessary '/./' path components we introduce by doing that.
 
     # Also keep a backup for comparison/debugging purposes.
 
     sed -i.bak -r \
         -e "s|\]\(([./]*[^:#)]+)#mlos-github-tree-view\)|](https://github.com/microsoft/MLOS/tree/main/${parent_path}/\1)|g" \
-        -e "s|\]\(([./]*[^:)]+)\)|](/MLOS/${parent_path}/\1)|g" \
-        -e "s|\]\(([./]*[^:#)]+)\.md(#[^)]+)?\)|](\1/\2)|g" \
+        -e "s|\]\(([./]*[^:#)]+)(#[^)]*)?\)|](/MLOS/${parent_path}/\1\2)|g" \
```
bpkroth, line 104: Makes sure not to swallow the anchor (if one exists) in the first capture so that we don't replace bare `#anchor-in-the-current-page` links with `/this/pages/parent/path/from/root/#anchor-in-the-current-page` (which no longer work).

website/build_site.sh
```diff
@@ -95,13 +95,17 @@ for content_filepath in $(find content/ -type f -name '*.md'); do
...
+        -e "s|\]\(([./]*[^:#)]+)\.md(#[^)]*)?\)|](\1/\2)|g" \
+        -e "s|\]\(([^)]*)/\./([^)]*)\)|](\1/\2)|g" \
+        -e "s|\]\(([^)]*)/\./([^)]*)\)|](\1/\2)|g" \
+        -e "s|\]\(([^)]*)/\./([^)]*)\)|](\1/\2)|g" \
```
bpkroth, line 108: replaces `/./` patterns introduced by the previous sed expressions with just `/`.  This isn't strictly necessary, but is nicer.

website/sphinx/apidoc.sh
```diff
@@ -21,6 +21,9 @@ echo "Installing dependencies for generating Python API docs"
 $pythonCmd -m pip install -e $MLOS_ROOT/source/Mlos.Python/
 $pythonCmd -m pip install sphinx sphinx_rtd_theme numpydoc matplotlib
 
+# Work around some issues with python not picking up the system installed package sometimes.
+$pythonCmd -m pip install 'kiwisolver>=1.0.2'
```
bpkroth, line 25: Addresses an issue I had while trying this in my WSL environment instead of docker where the `python-kiwisolver` package had been pulled in by `python-matplotlib` and the additional `python3.7` setup wasn't finding it.

website/test_site_links.sh
```diff
@@ -84,7 +84,39 @@ if [ -n "$linklint_output" ]; then
...
+sed -i.bak -r 's|<a href="#|<a href="/MLOS/#|g' public/index.html
+
+# Make / and /MLOS synonmous for the local checks.
+if [ -L public/MLOS ]; then
+    rm -f public/MLOS
+elif [ -e public/MLOS ]; then
+    echo "ERROR: public/MLOS already exists and si not a symlink." >&2
+    exit 1
+fi
+ln -s . public/MLOS
```
bpkroth, line 106: This adds an `MLOS` symlink to `.` to mimic `/MLOS/` URL paths working in the real webserver even when they're just being evaluated as local files.

website/test_site_links.sh
```diff
@@ -84,7 +84,39 @@ if [ -n "$linklint_output" ]; then
...
+
+linklint_output=$(linklint -quiet -silent -error -root public -local /@)
```
bpkroth, line 108: This only checks links locally, but does check anchors.

bpkroth: Checking anchors in the remote version threw lots of errors and slows things down.

amueller:
FYI not sure if you know about the magic issue mentions: https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword 
It seems "addresses" is not in the list.

website/sphinx/apidoc.sh
amueller: I'm not sure I understand that comment. It might be that you have an old/incomplete installation of matplotlib somewhere?

website/build_site.sh
amueller: I don't think I understand why this line is here three times.
Generally, as I mentioned in another place, I would prefer to give up on having the markdown links work and not try to maintain a patchwork of regexps to try to make the same file render with two different frameworks.

amueller (approved):
Seems ok though I trust you on the regexps. I think it'll be easiest to see if the check pass appropriately after merging.

amueller (commented):
changing to "comment" to indicate there's a question about the kiwi solver

website/sphinx/apidoc.sh
bpkroth: Very possible.  Without that, I was getting this error:

```
Running Sphinx v3.2.1

Extension error:
Could not import extension matplotlib.sphinxext.plot_directive (exception: No module named 'kiwisolver')
```

Poking around I see this:
```
$ python3.7 -m pip show matplotlib kiwisolver
Name: matplotlib
Version: 3.3.2
Summary: Python plotting package
Home-page: https://matplotlib.org
Author: John D. Hunter, Michael Droettboom
Author-email: matplotlib-users@python.org
License: PSF
Location: /home/bpkroth/.local/lib/python3.7/site-packages
Requires: kiwisolver, python-dateutil, pyparsing, pillow, numpy, certifi, cycler
Required-by: seaborn
---
Name: kiwisolver
Version: 1.0.1
Summary: A fast implementation of the Cassowary constraint solver
Home-page: https://github.com/nucleic/kiwi
Author: The Nucleic Development Team
Author-email: sccolbert@gmail.com
License: UNKNOWN
Location: /usr/lib/python3/dist-packages
Requires:
Required-by: matplotlib

$ dpkg -l | grep kiwisolver
ii  python3-kiwisolver                   1.0.1-3build1                       amd64        fast implementation of the Cassowary constraint solver - Python 3.X
```

Possibly the better thing to do is add it to the previous line as

```
$pythonCmd -m pip install --upgrade sphinx sphinx_rtd_theme numpydoc matplotlib kiwisolver
```

bpkroth: Of course, in the future we can switch to conda and consider this moot, but for now ...

amueller: ah so it's system python polluting the environment. I'm curious why installing matplotlib doesn't fix that, but if you use system python, all bets are off. Sure, feel free to keep it for now and we'll move to conda later.

website/build_site.sh
bpkroth: Late night laziness honestly.
Only two are necessary.
One comes from what's originally in the `.md` file links.
One comes from the `dirname` call which resolves to `.` for the top-level items, but just `path/to/subdir` instead of `./path/to/subdir` for others.
The latest changes address that.
Make sense?  Seems okay?

website/sphinx/apidoc.sh
bpkroth: Tried to consolidate to the latter form for now, and ended up with another dependency that it ran into issues with.
Thoughts?

amueller: on the CI or locally? I think ti's fine as long as it works on CI, locally you seem to have some unrelated issues.

amueller (approved):

website/sphinx/apidoc.sh
bpkroth: I mean, I would prefer that the scripts work both locally and in CI, but agreed on there being something funny when the system python gets in the way.
As discussed, doing this all via docker/conda is probably the more supportable path forward.

PR #88, opened 2020-09-18
initial work to binplace binaries on linux

bpkroth:
This provides "make install" style targets to binplace binaries to target/bin/{Release,Debug}/ for easier execution.

Note that SettingsRegistry assemblies were already being placed there.

With the current Mlos.Agent lookup behavior, this means that we can now
run

```sh
tools/bin/dotnet target/bin/Release/Mlos.Agent.Server.dll target/bin/Release/Mlos.UnitTest
```

Instead of having to specify
1. the dotnet build output paths to the Mlos.Agent.Server assembly
2. the cmake build output path to the target executable
3. the MLOS_SETTINGS_REGISTRY_PATH environment variable for the build output path of the relevant SettingsRegistry dlls

To do this, we expand binplace dotnet rules.

Next, we hook that into the existing Makefile and cmake wrappers as
well.

TODO:

- tests
- documentation


grlap (commented):
Why do we need to call dotnet publish  ?

build/Mlos.Common.targets
```diff
@@ -4,4 +4,7 @@
     <MSBuild Projects="@(ProjectReference)" Targets="Restore" RunEachTargetSeparately="true" />
     <MSBuild Projects="@(ProjectFile)" Targets="Restore" RunEachTargetSeparately="true" />
   </Target>
+  <!-- An empty placeholder for other targets to reference with an AfterTargets property -->
+  <!-- We can then invoke this one via "dotnet build /t:PublishProjectBinary" -->
+  <Target Name="PublishProjectBinary" />
```
grlap, line 9: Publish is for NetCore, why we need it in common ?

bpkroth:
> 
> 
> Why do we need to call dotnet publish ?

We don't have to, it's optional.

It's mostly another quality of life improvement, partially intended for the SmartCache notebook.

Without that, we have to call the following to run the test:

```sh
export MLOS_SETTINGS_REGISTRY_PATH=out/dotnet/source/Mlos.UnitTest/Mlos.UnitTest.SettingsRegistry/obj/AnyCPU
tools/bin/dotnet out/dotnet/source/Mlos.Agent.Server/obj/AnyCPU/Mlos.Agent.Server.dll out/cmake/Release/source/Mlos.UnitTest/Mlos.UnitTest
```

With the publish it can be just this:

```sh
cd target/bin/Release
./Mlos.Agent.Server ./Mlos.UnitTest
```

It looks nicer and is easier to do interactively rather than hunting through two different build system's output layouts.

The `MLOS_LIBRARY_PATH` mechanism is still useful for other scenarios though.

The `dotnet publish` isn't strictly necessary for that though, we could simply binplace all of the *.dlls in it's output directory into the `target/bin/$(Configuration)` directory and rely on still calling it like this:

```sh
tools/bin/dotnet target/bin/Release/Mlos.Agent.Server.dll target/bin/Mlos.UnitTest
```

Thoughts?

build/Mlos.Common.targets
bpkroth: That's a no-op definition.  It allows calling the target from a `dirs.proj` to publish recursively.

grlap: I know it is no op. 
I do not understand why do we need to publish host depended binaries in the build system.
We had this before , and I had it removed to reduce nuget footprint.

grlap: What we need to have is the settings registry in one folder, right ?
Maybe let's just modify the binplacing target (or a new one)
Mlos.SettingsSystem.CodeGen.targets 
  
  <Target Name="BinplaceSettingsRegistry" AfterTargets="Build">
    <Copy SourceFiles="$(OutputPath)\$(TargetFileName)" DestinationFolder="$(BinplaceDir)" />
  </Target>
?

bpkroth: I didn't realize those were included in the nuget.  Seems like that should probably be more specialized as well, no?

Anyways, as I mentioned elsewhere, I thought about this somemore and agree - publishing is unnnecessary.  I will yank that out in favor of binplacing the AnyCPU instead and calling it with `dotnet`.

Modifying the `Mlos.SettingsSystem.CodeGen.targets` to place everything in a single folder only maybe helps us for things internal to the MLOS repo.

Once we start add external project tuning (later this week?), then we'll still need to support some sort of search path since different projects will need different output locations.

bpkroth: Reverted and turned them into binplace operations instead.

grlap (approved):

PR #90, opened 2020-09-21
SmartCache example docs

bpkroth:


source/Examples/SmartCache/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+> Note: these commands are given relative to the root of the MLOS repo.
+>
+> To move there, you can execute the following within the repository:
+>
+> `cd $(git rev-parse --show-toplevel)`
+
+`SmartCache` can be invoked separate, or by the `Mlos.Agent.Server` itself.
+
+Once started, `SmartCache` will attempt to register its component specific set of shared memory messages with the `Mlos.Agent` in the `Mlos.Agent.Server` using some `Mlos.Core` component registration messages.  That message includes the name of the `SettingsRegistry` assembly (`.dll`) corresponding to that component's settings/messages.
```
amueller, line 74: I'm not sure how to fix this, but right now I got confused by the order of things. You're giving a high level overview and then you're giving the actual commands, which is good. But I didn't understand that on the first read and so I was like "you expected me to start SmartCache already but I didn't know how to".
In particular the "Executing" part starts with a very low-level instruction before going to a highlevel overview.

Maybe skipping the high-level overview in this place is easiest and do it inline with the commands? But that's just a didactic thing, we can polish this later.

source/Examples/SmartCache/README.md
```diff
@@ -0,0 +1,115 @@
...
+In this case we populate it with the path to the `SmartCache.SettingsRegistry.dll`:
+
+```sh
+export MLOS_SETTINGS_REGISTRY_PATH="out/dotnet/source/Examples/SmartCache/SmartCache.SettingsRegistry/obj/AnyCPU:$MLOS_SETTINGS_REGISTRY_PATH"
+```
+
+Next, we can the `Mlos.Server.Agent` using the `dotnet` command:
+
+```sh
+tools/bin/dotnet out/dotnet/source/Mlos.Agent.Server/obj/AnyCPU/Mlos.Agent.Server.dll
```
amueller, line 87: Does this output anything? If so, should we give the expected output here?
If not, we should change it to output "waiting for connection" or something like that.

source/Examples/SmartCache/README.md
```diff
@@ -0,0 +1,123 @@
...
+docker exec -it mlos-build /bin/bash
+```
+
+Now, we can start `SmartCache` as follows:
+
+```sh
+out/cmake/Release/source/Examples/SmartCache/SmartCache
+```
+
+> To have `Mlos.Agent.Server` start `SmartCache` without having to start another shell in the docker container instance, add the path to the `SmartCache` binary as an argument to the `dotnet ... Mlos.Agent.Server` invocation.
```
amueller, line 105: wouldn't it be easier / shorter to just give that instruction? I guess that doesn't show off the separation of the executable and the agent that well. Maybe removing that comment for the default instructions will make it easier? not sure.

source/Examples/SmartCache/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Caveats
+
+- The system currently only supports one shared memory region and doesn't cleanup the shared memory after itself.
+
+    To help with this, we currently provide a helper script to remove previous incarnations of the shared memory regions:
```
amueller, line 111: all of this is within the docker (might be obvious?)

amueller (commented):
some minor nitpicks, we can leave the bigger polish for later

source/Examples/SmartCache/README.md
bpkroth: Ack.  I think I was struggling with how to show the two methods, and maybe that's not necessary here.  Perhaps it'd be better suited in a `README.md` for `Mlos.Agent.Server`.

What's your preference when driving this from a notebook?

1. Have a single invocation that handles both processes (e.g. `dotnet path/to/Mlos.Agent.Server.dll path/to/SmartCache), or
2. Have two invocations:
  1. To start Mlos.Agent.Server
  2. To (repeatedly) start SmartCache to submit new results to it.

source/Examples/SmartCache/README.md
bpkroth: Yeah, I can make a comment about that

source/Examples/SmartCache/README.md
bpkroth: Yeah, that's a good idea.  I'll add that too.

source/Examples/SmartCache/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Executing
+
+> Note: these commands are given relative to the root of the MLOS repo.
+>
+> To move there, you can execute the following within the repository:
+>
+> `cd $(git rev-parse --show-toplevel)`
+
+`SmartCache` can be invoked separate, or by the `Mlos.Agent.Server` itself.
```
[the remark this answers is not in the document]
bpkroth: In a previous incarnation, `Mlos.Agent.Server` could be invoked separately and then the target application (e.g. daemon or different instantiations of the microbenchmark) could be started to connect to it over and over.  I think the other intention of that was to allow the notebook to be able to drive experiments by connecting to the Mlos.Agent.Server via grpc and having it start something.

When I tried that again just now, it seems that the `Mlos.Agent.Server` exits after the first shared memory client connects and finishes, so I guess that mode doesn't make sense for the moment and we can just focus on the one where `Mlos.Agent.Server` starts the microbenchmark via a commandline arg and exits after completion.

source/Examples/SmartCache/README.md
amueller: I think the second is more conceptually clear?

source/Examples/SmartCache/README.md
bpkroth: Added a note at the bottom.  Works?

source/Examples/SmartCache/README.md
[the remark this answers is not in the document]
bpkroth: I tried to reorganize this with a more simplified "command first, then explain" format.  Please have another look and let me know what you think.

source/Examples/SmartCache/README.md
bpkroth: I tried to reorganize this with a more simplified "command first, then explain" format.  Please have another look and let me know what you think.

source/Examples/SmartCache/README.md
bpkroth: Ok, I did the first just now, but I can change it back.

amueller: either is fine :)

amueller (approved):

source/Examples/SmartCache/README.md
bpkroth: Cool, let's go with this for the moment.  We can change it later.

bpkroth:
Oops, this never got published.  Sorry about that.

Issue #74, opened 2020-09-11
Reevaluate use of `getcwd` in python code


bpkroth:
There are several places where `getcwd` is used to compose a path assuming that the script is executed from the `source/Mlos.Python` directory or else used to create a temporary file.

For the first, we should change it to be relative to the file referencing it so that the script can be executed from a different directory.

For the second, we should be using the system provided get temp files to avoid security issues.

bpkroth:
Here's an example:
https://github.com/microsoft/MLOS/commit/b47737aa3e610c722afe99ef5819d85bc8fd204d

PR #92, opened 2020-09-21
2020/september/optimizing predict performance

byte-sculptor:
This change improves runtime of the DecisionTreeRegressionModel.predict() function.

The main idea is to use column-wise operators so that pandas and numpy can vectorize them. Tracing reveals 2x-200x improvement in filtering out invalid rows depending on the complexity of the hypergrid and the number of rows in the dataframe.


source/Mlos.Python/mlos/Spaces/Hypergrid.py
```diff
@@ -50,13 +59,73 @@ def get_dimensions_for_point(self, point, return_join_dimensions=True):
...
     def random_dataframe(self, num_samples):
         config_dicts = [
             {dim_name: value for dim_name, value in self.random()}
             for _ in range(num_samples)
         ]
         return pd.DataFrame(config_dicts)
 
+    @trace()
+    def filter_out_invalid_rows(self, original_dataframe: pd.DataFrame, exclude_extra_columns=True) -> pd.DataFrame:
+        """ Returns a dataframe containing only valid rows from the dataframe.
```
byte-sculptor, line 72: from the original_dataframe

source/Mlos.Python/mlos/Spaces/unit_tests/TestFilteringOutInvalidRows.py
```diff
@@ -0,0 +1,92 @@
...
+            if not space.is_hierarchical():
+                # For flat spaces we can choose between the column-wise operators and the row-wise validation. This is to get the tracing data to see the
+                # perf difference, but also to validate correctness by computing the desired index in yet another way.
+                #
+                with traced(scope_name="faster_filtering"):
+                    faster_valid_rows_index = random_dataframe_with_invalid_rows[random_dataframe_with_invalid_rows.apply(
+                        lambda row: Point(**{dim_name: row[i] for i, dim_name in enumerate(space.dimension_names)}) in space,
+                        axis=1
+                    )].index
+                self.assertTrue(faster_valid_rows_index.equals(fast_valid_rows_index))
```
byte-sculptor, line 89: rename these indices to indicate how they were computed

source/Mlos.Python/mlos/Spaces/Hypergrid.py
```diff
@@ -50,13 +59,73 @@ def get_dimensions_for_point(self, point, return_join_dimensions=True):
...
+
+        Valid rows are rows with no NaNs and with values for all dimensions in the required ranges.
+
+        If there are additional columns, they will be preserved.
+
+        :param original_dataframe:
```
amueller, line 78: Do you want to use the numpydoc standard? Then it'll render properly on the sphinx website.

source/Mlos.Python/mlos/Spaces/Hypergrid.py
amueller, line 72: There should be no space after ``"""``: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

source/Mlos.Python/mlos/Spaces/Hypergrid.py
```diff
@@ -50,13 +59,73 @@ def get_dimensions_for_point(self, point, return_join_dimensions=True):
...
+
+        valid_rows_index = None
+        dataframe = original_dataframe[self.dimension_names]
+
+        if not self.is_hierarchical():
+            # Let's exclude any extra columns
+            #
+            valid_rows_index = dataframe.index[dataframe.notnull().all(axis=1)]
+
+            # Now for each column let's filter out the rows whose values are outside the allowed ranges.
```
amueller, line 92: Is there a reason not to throw an error if the data is not in the right space?

source/Mlos.Python/mlos/Spaces/Hypergrid.py
```diff
@@ -50,13 +59,73 @@ def get_dimensions_for_point(self, point, return_join_dimensions=True):
...
         ]
         return pd.DataFrame(config_dicts)
 
+    @trace()
+    def filter_out_invalid_rows(self, original_dataframe: pd.DataFrame, exclude_extra_columns=True) -> pd.DataFrame:
+        """ Returns a dataframe containing only valid rows from the dataframe.
+
+        Valid rows are rows with no NaNs and with values for all dimensions in the required ranges.
+
+        If there are additional columns, they will be preserved.
```
amueller, line 76: the default is ``exclude_extra_columns=True`` which would mean they are not preserved by default, right?

amueller (commented):
I think I'm not entirely clear on where this would be used. How would you get an invalid point, and why wouldn't you raise in that case?

source/Mlos.Python/mlos/Spaces/Hypergrid.py
byte-sculptor: Yeah, I'll do that from now on.

byte-sculptor: And I'll fix this PR.

amueller: Thanks! I think it would be great to move to it bit-by-bit whenever we work on docs.

source/Mlos.Python/mlos/Spaces/Hypergrid.py
byte-sculptor: Good point.

source/Mlos.Python/mlos/Spaces/Hypergrid.py
byte-sculptor: Random forest passes all the data to the decision trees and each tree has to filter out data that's within its space. We could do this filtering in the random forest class too, but the mechanics would be the same.

source/Mlos.Python/mlos/Spaces/Hypergrid.py
amueller: (I know this is a terrible nitpick but it's a standard style ;)

source/Mlos.Python/mlos/Spaces/Hypergrid.py
amueller: Hm I'm wondering if we give up on some input checking here. One of the benefits of having such an explicit system for dimensions is that we can check if everything lines up properly and I feel we're not making use of this here. I don't have a super strong opinion though. I guess it depends on where we call this function.

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctionOptimizers/unit_tests/TestUtilityFunctionOptimizers.py
```diff
@@ -100,7 +100,7 @@ def tearDownClass(cls) -> None:
         temp_dir = os.path.join(os.getcwd(), "temp")
```
bpkroth, line 100: This comment is not related to this PR, so we can take it for later, but ...
This pattern is sensitive to the current working directory the user happens to be in when they execute the scripts.  I made issue #74 to track it a while back because unit tests were failing if not run from a particular directory (b47737aa3e610c722afe99ef5819d85bc8fd204d).
Would it be better to dump these in a consistent location, regardless of where they were at that point.  For instance, in `out/` where most of the other non-source stuff goes?  Trouble in this case (unlike the unit test case) is it makes examining the debug output more cumbersome because it's not sitting in the directory you're already in.

source/Mlos.Python/mlos/Spaces/Hypergrid.py
bpkroth: Once those are more cleaned up we can also consider to add that as part of CI checks

source/Mlos.Python/mlos/Tracer.py
```diff
@@ -28,7 +28,7 @@ def wrapper(*args, **kwargs):
                 try:
                     result = wrapped_function(*args, **kwargs)
                     end_timestamp_ns = int(time.time() * 1000000)
-                    if end_timestamp_ns <= start_timestamp_ns:
+                    while end_timestamp_ns <= start_timestamp_ns:
```
bpkroth, line 31: Curious, is there some case where time.time() goes backwards for some reason that the if statement version doesn't work?

source/Mlos.Python/mlos/Spaces/unit_tests/TestFilteringOutInvalidRows.py
```diff
@@ -0,0 +1,92 @@
...
+                        valid_indices.append(idx)
+                slow_valid_rows_index = pd.Index(valid_indices)
+
+            print(f"{len(slow_valid_rows_index)}/{len(random_dataframe_with_invalid_rows.index)} rows are valid.")
+            self.assertTrue(0 < len(slow_valid_rows_index))
+            self.assertTrue(len(slow_valid_rows_index) < num_samples)
+
+            # Let's filter out invalid rows the fast way.
+            #
+            fast_valid_rows_index = space.filter_out_invalid_rows(original_dataframe=random_dataframe_with_invalid_rows, exclude_extra_columns=True).index
```
bpkroth, line 77: Do you want to also validate that you expect this to be faster than the slow method as a sort of performance regression test?  Might be prone to flakiness or just be too long?

bpkroth (approved):
Looks good to me.  Just a few comments from me, nothing major.  Main one is a question about the path to the temp/ dir used for debug/trace output, but that's a pattern beyond this PR, so we can take it up in #74.

amueller (approved):

source/Mlos.Python/mlos/Spaces/unit_tests/TestFilteringOutInvalidRows.py
byte-sculptor: No, this is a functional test.

source/Mlos.Python/mlos/Tracer.py
byte-sculptor: I've seen it go backwards a bunch of times. Never seemed to have been more than 100 but this seems more robust.

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctionOptimizers/unit_tests/TestUtilityFunctionOptimizers.py
byte-sculptor: Sure. Let's tackle it as part of #74 though.

Issue #95, opened 2020-09-22
test_lasso_hierarchical_categorical_predictions seems flaky


bpkroth:
@edcthayer adding an issue to track this here

Seen this a couple of times recently.  Sometimes with slightly different KeyErrors (e.g. `medium_quadratic_params` instead).
Rerunning it seems to make it go away.

```
2020-09-22T15:46:14.7210731Z [10 rows x 5 columns]
2020-09-22T15:46:14.7211123Z     raise_missing = True
2020-09-22T15:46:14.7211913Z     self = <pandas.core.indexing._LocIndexer object at 0x0000021609A396D8>
2020-09-22T15:46:14.7212932Z   File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\pandas\core\indexing.py", line 1646, in _validate_read_indexer
2020-09-22T15:46:14.7213796Z     raise KeyError(f"{not_found} not in index")
2020-09-22T15:46:14.7214363Z     ax = Index(['vertex_height', 'medium_quadratic_params.x_1',
2020-09-22T15:46:14.7214990Z        'medium_quadratic_params.x_2', 'low_quadratic_params.x_1',
2020-09-22T15:46:14.7215540Z        'low_quadratic_params.x_2'],
2020-09-22T15:46:14.7216039Z       dtype='object')
2020-09-22T15:46:14.7216430Z     axis = 1
2020-09-22T15:46:14.7216840Z     indexer = array([ 0,  3,  4,  1,  2, -1, -1], dtype=int64)
2020-09-22T15:46:14.7217460Z     key = Index(['vertex_height', 'low_quadratic_params.x_1', 'low_quadratic_params.x_2',
2020-09-22T15:46:14.7218181Z        'medium_quadratic_params.x_1', 'medium_quadratic_params.x_2',
2020-09-22T15:46:14.7218832Z        'high_quadratic_params.x_1', 'high_quadratic_params.x_2'],
2020-09-22T15:46:14.7219465Z       dtype='object')
2020-09-22T15:46:14.7219832Z     missing = 2
2020-09-22T15:46:14.7220333Z     not_found = ['high_quadratic_params.x_1', 'high_quadratic_params.x_2']
2020-09-22T15:46:14.7220882Z     raise_missing = True
2020-09-22T15:46:14.7221500Z     self = <pandas.core.indexing._LocIndexer object at 0x0000021609A396D8>
[...]

PR #100, opened 2020-09-24
added feature imputation to RERF predict()

edcthayer:
Should fix: test_lasso_hierarchical_categorical_predictions seems flaky #95

### DIFF
--- a/source/Mlos.Python/mlos/Optimizers/RegressionModels/RegressionEnhancedRandomForestModel.py
+++ b/source/Mlos.Python/mlos/Optimizers/RegressionModels/RegressionEnhancedRandomForestModel.py
@@ -204,6 +204,13 @@ class RegressionEnhancedRandomForestRegressionModel(RegressionModel):
         self.variance_estimate_ = None
         self.root_model_gradient_coef_ = None
         self.polynomial_features_powers_ = None
+        self.num_dummy_vars_ = None
+        self.num_categorical_dims_ = None
+        self.continuous_dim_col_names_ = None
+        self.categorical_dim_col_names_ = None
+        self.dummy_var_map_ = None
+        self.dummy_var_cols_ = None
+        self.categorical_zero_cols_idx_to_delete_ = None
 
     @trace()
     def fit(self, feature_values_pandas_frame, target_values_pandas_frame, iteration_number=0):
@@ -419,10 +426,8 @@ class RegressionEnhancedRandomForestRegressionModel(RegressionModel):
         #  if any are missing, impute NaN values
         missing_column_names = set.difference(set(self.input_dimension_names),
                                               set(feature_values_pandas_frame.columns.values))
-        if missing_column_names:
-            num_x = len(feature_values_pandas_frame.index)
-            for missing_column_name in missing_column_names:
-                feature_values_pandas_frame[missing_column_name] = np.NaN * num_x
+        for missing_column_name in missing_column_names:
+            feature_values_pandas_frame[missing_column_name] = np.NaN
         x_df = feature_values_pandas_frame[self.input_dimension_names]
         x_star = self.transform_x(x_df)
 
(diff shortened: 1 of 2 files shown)


amueller:
Thanks! Can you maybe add a test that explicitly tests this?

edcthayer:
The current test_lasso_hierarchical_categorical_predictions test is flaky b/c (my assessment) the size of the prediction X (inputs) is small (N=10 uniform randomly selected in the hierarchical HyperGrid).  At this size, some test executions fail to contain all the levels observed in the (larger) training data.   Would it suffice to modify the "flaky test" so it always fails without this imputation, e.g. N=1?

amueller:
If it still tests the thing it was originally designed to test then sure. Otherwise maybe add a separate test with N=1?

edcthayer:
Added a loop to the test in which the prediction is performed for a single random x.  The prediction results are collected from the loop so the original prediction accuracy could still be computed.

Fixes required retaining more information from the .fit() so as to properly "explode_x" on a row lacking the columns needed to infer the needed transformations (one hot encoding, creating the design matrix - and pruning it).

This makes me worry what will happen if the training features from hierarchical hypergrids fail to contain sufficient row diversity to construct the "full" required transformations during the .fit() call.  So need to sleep on how .fit() might be modified to ensure the explode_x is "universal" wrt the input space dimensions.

amueller:
I'm not entirely sure I understand the issue, but would it be possible to reuse a sklearn ColumnTransformer to preserve the mapping? Or maybe alternatively write an Adaptor that does something similar using the dimension information?

edcthayer:
I'm not sure I can explain this well enough, but will try - without getting into how the Lasso/Ridge design matrix is constructed.

RERF regression allows for polynomial basis functions.  But these make no sense for categorical dimensions, so it does some "dimension management" to know which dimensions should be expanded to polynomial functions.  RERF has it's own one hot encoding (OHE) that's applied to the "discovered" categorical dimensions.  RERF keeps track of which augmented feature dimensions contain OHE values so as to correctly apply the polynomial basis function expansion and construct the linear regression design matrix (passed into sklearn's Lasso Regression).  

The initial "dimension management" logic attempts to infer both the categorical dimensions (when present - let's call this concern C1) and the count of the number of OHE levels across the complete universe of the input space (call this C2).  The flakey test issue arose b/c the points passed to the .predict() method were too few to reconstruct the list of all the feature dimensions.  This happens b/c Hierarchical HyperGrid random point generation doesn't always return the full set of input space columns.  Since .predict() couldn't recreate the dimension discovery outcome that .fit() arrived at, it mistakenly projects the .predict() X into the wrong expanded space and the lasso model coefficient matrix (from .fit()) makes no sense (wrong dimension and column arrangement).  The PR as it currently exists "remembers" enough information from the .fit() method's discovery process to transform a single .predict() X row into the design matrix space auto detected in .fit().

This brings me to C2:  After addressing C1, I realized .fit() can also be fooled by a small sample of X from hierarchical hypergrids (even when C1 doesn't happen, i.e. all dimensions are present) b/c the complete list all possible levels of each categorical dimension may be missing.  Hence the .fit() "discovered" transformation my be incorrect (incomplete OHE) if the .predict() X contains a categorical level not found in the feature frame sent to .fit().

- C1 might be addressed in the Hierarchical HyperGrid by ensuring all dimensions in the input space are included even if their values are NaNs.
- C2 requires knowing all categorical levels spanning the input space before performing the .fit().  I'll explore "helper methods" for hypergrids to see if the hypergrid can generate a list of all categorical levels (presumably via some well chosen traversal method) w/o relying on randomly generating a large sample set.  
- Pushing the OHE into an adaptor and loosing track of which features coming out the adaptor represent OHE columns removes the RERF model's ability to fit distinct surfaces at each level of the flattened categorical subspace.  To see why, we'd need to dig into the math more deeply.
- Ignoring concerns about an OHE adaptor's output, I don't see how using an adaptor to perform the polynomial basis expansion would be ideal since the RERF model would be unable to construct the algebraic gradient to the Lasso fit.  While we're not using this "feature" of the RERF surrogate model in an optimzer, it was my hope that it might be of value as seen in the initial notebook evaluation of the RERF model in the presence of noise.

Hope this clarifies the issue some - though I think one will eventually need to dig into the design matrix construction for full clarity.  The CMU statistics course (relevant chapters listed below) is probably required reading if one needs a refresher on how OHE is then used in linear regressions with basis functions.
1. http://www.stat.cmu.edu/~cshalizi/mreg/15/lectures/lecture-16.pdf - intro to encoding categorical & ordinal dimensions.
2. http://www.stat.cmu.edu/~cshalizi/mreg/15/lectures/19/lecture-19.pdf - most closely describes what's done in RERF Lasso/Ridge regressions.

byte-sculptor (approved):

byte-sculptor:
> I'm not entirely sure I understand the issue, but would it be possible to reuse a sklearn ColumnTransformer to preserve the mapping? Or maybe alternatively write an Adaptor that does something similar using the dimension information?

We spoke with Ed, and agreed that one-hot-encoding would valuable to have as an Adapter. When choosing the implementation we should evaluate the ColumnTransformer and possibly build the Adapter on top of it. 

Additionally, I view this PR as a step in the right direction - it fixes intermittent test failures - and as such we should check it in. Refactoring into an adapter is a larger effort, and in my opinion deserves its own PR and tests.

PR #102, opened 2020-09-24
Adding Config Stores

byte-sculptor:
This replaces the XYZConfig classes with ComponentConfigStore objects.


source/Mlos.Python/mlos/Examples/SmartCache/CacheImplementations/LruCache.py
```diff
@@ -4,19 +4,15 @@
 #
 from mlos.Examples.SmartCache.CacheImplementations.XruCache import XruCache
 from mlos.Spaces import DiscreteDimension, Point, SimpleHypergrid
-from mlos.Spaces.Configs.DefaultConfigMeta import DefaultConfigMeta
+from mlos.Spaces.Configs.ComponentConfigStore import ComponentConfigStore
 
-class LruCacheConfig(metaclass=DefaultConfigMeta):
-    CONFIG_SPACE = SimpleHypergrid(
+LruCacheConfigStore = ComponentConfigStore(
```
byte-sculptor, line 9: Is this the right naming convention? Or should we go with lru_cache_config_store?

sergiy-k (approved):

source/Mlos.Python/mlos/Examples/SmartCache/CacheImplementations/LruCache.py
amueller: the Python way would be underscores, I think: https://www.python.org/dev/peps/pep-0008/#global-variable-names
But if consistency with C# style is a concern I think CamelCase wouldn't be too disruptive? Usually that's reserved for classes.

Issue #55, opened 2020-09-04
run notebooks when building website


amueller:
Right now the checked-in version is just converted with nbconvert, we should add the ``execute`` flag.

bpkroth:
Working on this along with #35

bpkroth:
There's stub work for this in #61, but it currently has some errors that need to get resolved.

PR #107, opened 2020-09-26
2020/september/extending grpc tests

byte-sculptor:
All tests in TestBayesianOptimizer now test both a local optimizer and an optimizer created via gRPC. 

These tests ended up being pretty slow since we didn't have an API to register observations in bulk over gRPC. So I added that, along with a bunch of tracing to diagnose the problem and validate the improvement.

Remaining work:
- [x] Make sure that notebooks continue to work.
- [x] Linter


source/Mlos.Notebooks/SmartCacheOptimization.ipynb
```diff
@@ -18,7 +18,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
```
bpkroth, line 21: @amueller to comment further, but I think we want to the output for this one for the webpages that get published.
At least until we can fix automatically rendering them during website generation.

source/Mlos.Proto/OptimizerService.proto
```diff
@@ -67,6 +71,14 @@ message RegisterObservationRequest
     Observation Observation = 2;
 };
 
+// A request to add a observations to the optimizer's data set.
+//
+message RegisterObservationsRequest
+{
+    OptimizerHandle OptimizerHandle = 1;
+    Observations Observations = 2;
```
bpkroth, line 79: I think we talked about this already at one point, but can you please remind me why we need the difference for a single Observation and and Observations instead of an array of Observation objects (e.g. using `repeated`)?

source/Mlos.Python/mlos/Optimizers/BayesianOptimizerFactory.py
```diff
@@ -24,12 +25,28 @@ class BayesianOptimizerFactory:
...
         self.logger = logger if logger is not None else create_logger("BayesianOptimizerFactory")
         self._grpc_channel = grpc_channel
-        self._optimizer_service_stub = OptimizerServiceStub(channel=self._grpc_channel)
+        self._optimizer_service_stub = None
+        if self._grpc_channel is not None:
+            self._optimizer_service_stub = OptimizerServiceStub(channel=self._grpc_channel)
 
-    def create_remote_optimizer(self, optimization_problem: OptimizationProblem, optimizer_config: Point = None) -> BayesianOptimizerProxy:
+
+    def create_local_optimizer(self, optimization_problem: OptimizationProblem, optimizer_config: Point = None) -> OptimizerBase:
```
bpkroth, line 36: Should the type returned by the BayesianOptimizerFactory be a BayesianOptimizer, or does that mess with the class hierarchy too much.

bpkroth:
Batch observation registration makes sense.
I think my main thought is that it would be nice if it didn't need a separate API and could just be the default (i.e. drop the singleton case if possible).  Might be future work though.

source/Mlos.Proto/OptimizerService.proto
```diff
@@ -34,6 +34,10 @@ service OptimizerService
     //
```
grlap, line 34: Could you add a simple Echo ?

grlap (approved):

source/Mlos.Proto/OptimizerService.proto
byte-sculptor: In short: column vs. row layout. Observations end up being a dictionary of lists, rather than a list of dictionaries and as a result are faster to (de)serialize, occupy less space and are more easily ingested by pandas.

source/Mlos.Python/mlos/Optimizers/BayesianOptimizerFactory.py
byte-sculptor: Yeah, that will work.

source/Mlos.Proto/OptimizerService.proto
byte-sculptor: Sure.

byte-sculptor:
> Batch observation registration makes sense.
> I think my main thought is that it would be nice if it didn't need a separate API and could just be the default (i.e. drop the singleton case if possible). Might be future work though.

Might be a good idea. However, right now I'm trying to minimize changes to gRPC until we have really good tests of the C# - Python interactions. Once we have them, we can move fast.

source/Mlos.Notebooks/SmartCacheOptimization.ipynb
byte-sculptor: OK, I'll re-render these.

byte-sculptor: We should have tests for this.

byte-sculptor: Added an issue: #111.

bpkroth: We also have #55 already

source/Mlos.Proto/OptimizerService.proto
bpkroth: Should probably add some comments to that effect.

PR #108, opened 2020-09-26
Grpc linux fixups

bpkroth:
Apply a oneline fix to the `dotnet.env` allow the C# --> Python optimizer unencrypted grpc connection to work again (35944295ebb6b64f990c9030cb7d64848896368d).

Make the `SmartCache` unit test run with the optimizer.
- This requires a shell script helper to start/stop the optimizer in the background.
  (currently only done for Linux - should add something similar for Windows)
- Also had to fix the `start_mlos_optimizer_service.py` signal handler to deal with `SIGTERM` as well (the default `kill` signal), else it leaves the child process around.

Include some additional fixups developed during that effort:
- more out of the box vscode debugger configs
- a `make dotnet-build-quick` target that skips the restore and dependencies checks.  This is similar to the `cmake ...-fast` target and is mostly just nice for quicker dev/build/test cycles when you know you've only changed a single file.


build/Mlos.Cpp.UnitTest.cmake
```diff
@@ -39,6 +39,17 @@ set_tests_properties(CheckForMlosSharedMemories PROPERTIES
 #    FIXTURES_CLEANUP MlosSharedMemoriesChecks)
 # Now other test can add "MlosSharedMemoriesChecks" to their
 # "FIXTURES_REQUIRED" property to invoke these scripts before/after themselves.
+add_test(NAME StartMlosOptimizerService
+    COMMAND ${MLOS_ROOT}/build/CMakeHelpers/BackgroundProcessHelper.sh
+        start /tmp/mlos_optimizer_microservice.pid /tmp/mlos_optimizer_microservice.log
+        ${PYTHON3} ${MLOS_ROOT}/source/Mlos.Python/mlos/start_optimizer_microservice.py launch --port 54321)
```
bpkroth, line 45: FIXME: This requires a `pip install -e source/Mlos.Python` for CI

source/Mlos.Agent.Server/MlosAgentServer.cs
```diff
@@ -191,7 +191,7 @@ public static void Main(string[] args)
                 }
             }
 
-            // Print any exceptions if occured.
+            // Print any exceptions if occurred.
```
amueller, line 194:
```suggestion
            // Print any exceptions that occurred.

```

amueller (approved):

grlap (approved):

PR #109, opened 2020-09-26
Initial work to switch to system installed tools

bpkroth:
This is just a draft for now.

Initial fixups following efforts for #108.

Originally, we had scripts to fetch some tools into the local repo.

This has some benefits wrt to lowering the environment setup cost for users.  However, then the `PATH` environment needs to be setup to point at those.  In build scripts this is easy, but for local debugging it's less awesome.

This initially just _adds_ support for just using the system installed tools if they're new enough.

It does not yet, entirely remove the old path.  The main reason for this is the current CI pipelines still depend upon them.

As we transition to Docker for everything I will add that too and then we can cleanup the scripts and environment variables considerably.

For instance, I don't think most of the `DOTNET_` env variables are necessary anymore so that mess of wrappers can go away.

At best we could simply keep a script that checks that all of the installed versions of tools are new enough for us and if not throw a warning.

Not sure how we'd handle forced downgrades, but if we encounter something like that we could probably resurrect the current  fetching mechanism at that point.


scripts/install.cmake.sh
```diff
@@ -0,0 +1,19 @@
...
+if ! [ -x /usr/bin/cmake ]; then
+    wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | gpg --dearmor - | sudo tee /etc/apt/trusted.gpg.d/kitware.gpg >/dev/null
+    set -x
+    sudo apt-add-repository "deb https://apt.kitware.com/ubuntu/ `lsb_release -c -s` main"
+    sudo apt-get update
+    sudo apt-get --no-install-recommends -y install cmake
+    set +x
+else
+    echo '/usr/bin/cmake is already available.  Use "apt-get update && apt-get upgrade" to update it.'
+fi
```
bpkroth, line 19: Should add notes about these scripts to the setup docs if we go the route of ditching the "fetch local" method.

bpkroth:
Pulled this into #128.

PR #110, opened 2020-09-27
Porting ADO changes to GitHub 2020-09-26

bpkroth:


source/Mlos.Agent.Server/MlosAgentServer.cs
```diff
@@ -104,10 +104,15 @@ public static void Main(string[] args)
                 MlosContext.OptimizerFactory = new MlosOptimizer.BayesianOptimizerFactory(optimizerAddressUri);
             }
 
-            // Create (or open) the circular buffer shared memory before running the target process.
+            // In the active learning mode, create a new shared memory map before running the target process.
```
amueller, line 107: I know this is from upstream but I think we really should rename this mode, maybe to "agent-controlled". Though that's also a bit weird, because 'agent' also has a specific meaning in ML that's different from our meaning (in reinforcement learning speak, the target program we're optimizing would be the agent).

bpkroth: That's fine with me.  I would add a separate commit for that though so that we can easily pull it back the other direction.
Question is do we do that in this PR or another?

amueller: different PR, just a general comment.

grlap (approved):

PR #118, opened 2020-09-28
add instructions for lab2

amueller:


source/Mlos.Notebooks/SmartCacheCPP.ipynb
```diff
@@ -462,16 +462,9 @@
    "metadata": {},
    "source": [
     "# Going Further\n",
-    "\n",
-    "1. Implement another workload in the SmartCache executable (`Workloads.cpp`) and compile again. How does a different workload impact the cache hit rate for the different configurations?"
+    "1. Instead of cache hit rate, use a metric based on runtime (e.g. latency, throughput, etc) as performance metric. Runtime can be measured for example using ``chrono::high_precision::now()``, also see [this blog post](https://bduvenhage.me/performance/2019/06/22/high-performance-timer.html). How does the signal from the runtime based metric compare to the application specific one (hit rate)? How consistent are the runtime results across multiple runs?\n",
```
bpkroth, line 465:
```suggestion
    "1. Instead of cache hit rate, use a metric based on runtime (e.g. latency, throughput, etc) as performance metric. Runtime can be measured for example using ``chrono::high_precision_clock::now()``, also see [this blog post](https://bduvenhage.me/performance/2019/06/22/high-performance-timer.html). How does the signal from the runtime based metric compare to the application specific one (hit rate)? How consistent are the runtime results across multiple runs?\n",
```

bpkroth:
```suggestion
    "1. Instead of cache hit rate, use a metric based on runtime (e.g. latency, throughput, etc) as performance metric. Environment (context) sensitive metrics can also be measured (e.g. [time](https://bduvenhage.me/performance/2019/06/22/high-performance-timer.html). How does the signal from the runtime based metric compare to the application specific one (hit rate)? How consistent are the runtime results across multiple runs?\n",
```

bpkroth (approved):
Minor tweaks as discussed.

PR #122, opened 2020-09-29
Fixup comment word wrap settings for vscode vim

bpkroth:
Tweaks to vscode plugins to reflow comments at a width that matches the rest of the code style.

Noticed by @grlap


bpkroth:
Actually, I'm not sure we actually enforce line length with stylecop right now:
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/2302

grlap (approved):

Issue #113, opened 2020-09-28
Execute Python long haul tests on a schedule


bpkroth:
@byte-sculptor observed that we are currently missing the Python long haul tests.

We don't necessarily want these in the CI pipelines (they take too long), but we do want to execute them periodically.

To do that, we'll need a separate `.github/workflows/scheduled.yml` sort of file for the scheduled tasks.

Another task that we'll want to put in there are periodic docker image rebuilds (e.g. to catch security patches).  See Also: #36 

See Also: https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#scheduled-events

bpkroth:
Seems that didn't work.

bpkroth:
@byte-sculptor can you look into just decreasing the length of those long haul runs to say 5 hours 55 minutes instead?

Issue #105, opened 2020-09-25
Improve docker development integration with vscode


bpkroth:
We should add a `.devcontainer/` json config to allow easier development integration between vscode and docker:
https://code.visualstudio.com/docs/remote/containers

Probably best to do this after #36

Issue #87, opened 2020-09-17
build instructions: docker is not alternative path


amueller:
Looking at the build docs it looks like there are three paths, docker, linux and windows:
https://microsoft.github.io/MLOS/documentation/02-Build/#docker

But docker is actually "just" how to install docker, so you still need to run the linux installation afterwards.
I think we should restructure the docs to make this more clear, or just say "now do the linux build" after the docker install.

Issue #83, opened 2020-09-16
Move the CI to conda


bpkroth:
Do we have an issue to move the CI to conda?

_Originally posted by @amueller in https://github.com/microsoft/MLOS/pull/79#issuecomment-693479238_

amueller:
the basic idea is to have CI download miniconda (or potentially miniforge) and create a clean conda env for anything that's running python, and try to use pip as little as possible.

bpkroth:
Can also consider to combine this with efforts to publish docker images (#36) via CI as well.

PR #128, opened 2020-10-05
Docker image publishing work

bpkroth:
Lots of build/pipeline cleanup with the goal of making and publishing standard Docker images the for Linux developers and the CI pipelines (https://github.com/microsoft/MLOS/issues/36).

- Most implementation notes are in `.github/workflows/main.yml`, but here's a brief summary:
  - Nightly we build fresh docker images for Ubuntu {16.04, 18.04, 20.04}.  They are only published if all of the usual build/test jobs pass (using the same logic as for PRs).  That also helps prime the local caches for the following:
  - On each PR, we 
    1. (re)build the image from a local buildx cache that's keyed on the sha of the PR, and optionally will also pull from the latest "main" build (since it presumably worked).  That is initially only stored in a local cache.  Part of the goal here is to avoid the need to push unvetted docker images (e.g. tagged by sha) to the public registry and then have to clean them up later (which is not always even possible).
    2. For each build/test job, we again "rebuild" the image, using that cache from step 1, so that it goes quickly.  Then we load the image into the local runner's docker daemon and start an instance to run our tests.  This reload is necessary because individual jobs are assigned to new runners so we can't reuse state between jobs.
        Jobs are organized into dependency graphs to account for this cache reuse.
        At the moment, all of the tests run on all versions of Ubuntu, which makes the CI pipeline a bit slow.
        - [ ] TODO: We could consider only running 16.04 and 20.04 in the CI pipelines and then running 18.04 in the nightly scheduled run or else just drop support for it entirely.
    3. Once all of the tests pass, and the PR is merged, we execute a final job to update the published version of the job by retagging it and pushing it to various registries.  The current registries are all Github services (for convenience to get started):
      - docker.pkg.github.com: this is nice because it supports the built in `GITHUB_TOKEN` secret so doesn't require any additional work for a forked repo to use for testing (they push to their own path though).  However, it doesn't support anonymous pulls and is being deprecated by ...
      - ghcr.io: this the newly announced Github Container Registry service.  It requires a special set of secrets to be assigned to the repo, so needs some additional work to make use of.  Also, as of now, the Microsoft org doesn't seem to have it enabled, so for now I've set it to push to the Microsoft-CISL org.

Additional work:
- Reorg the Dockerfile to use multi-stage builds to allow skipping copying the source code into it so that we only distribute the build tools.  We can also optionally build slimmer "python-only" images this way.  In a future PR we'll take on the work to switch from straight pip to conda (https://github.com/microsoft/MLOS/issues/83)
  - [ ] TODO: Could optionally prime the `/src/MLOS` folder in the image so that one could `git pull` immediately to get the latest code if they wanted to without a local copy as a bind mount.
- Added a `make docker-image` target to simplify repeated local docker image build invocations.
- Switch to using tools (cmake, dotnet) installed in the system rather than fetched during build (https://github.com/microsoft/MLOS/pull/109)
  - These scripts are also nice reusable elements for users to install in their own environments if they wish.
    - [ ] TODO: Update the documentation to reference them instead of relying on the old "fetch" method.
    - [ ] TODO: There is work in ADO that provides a similar `install.python.sh` script to setup the `python3.7` install.  Will port that over later as well.
  - Cleanup _most_ of the build system to match (I have not yet removed the old setup scripts, but they are no longer automatically called).
- Added vscode integration support for developing inside the container (https://github.com/microsoft/MLOS/issues/105, https://github.com/microsoft/MLOS/issues/2)
  - Added support for running as a non-root user so that bind mounts of the repo source tree work without re-chowning files on the host system as root.
    This required a fair bit of fixups, but I think should cut down on some issues with students/others and makes the vscode integration easier as well.
- Add nightly run scheduling of Python long haul tests (https://github.com/microsoft/MLOS/issues/113)
- [ ] TODO: Fixup docker related documentation (https://github.com/microsoft/MLOS/issues/87)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -17,5 +17,10 @@
     "forwardPorts": [8080, 50051],
     "image": "ghcr.io/microsoft-cisl/mlos/mlos-build-ubuntu-20.04:latest",
     "workspaceMount": "source=${localWorkspaceFolder},target=/src/MLOS,type=bind,consistency=cached",
-    "workspaceFolder": "/src/MLOS"
+    "workspaceFolder": "/src/MLOS",
+    "remoteUser" : "mlos-docker",
+    "mounts" : [
+        "source=mlos-build-ubuntu-20.04-vscode-exts,target=/home/mlos-docker/.vscode-server/extensions,type=volume",
+        "source=mlos-build-ubuntu-20.04-vscode-exts,target=/home/mlos-docker/.vscode-server-insiders/extensions,type=volume"
+    ]
 }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,6 +1,19 @@
 {
     "name": "mlos-build-ubuntu-20.04",
-    "extensions": [],
+    "extensions": [
+        "ms-vscode.cmake-tools",
+        "ms-vscode.cpptools",
+        "ms-dotnettools.csharp",
+        "ms-python.python",
+        "zxh404.vscode-proto3",
+        "editorconfig.editorconfig",
+        "davidanson.vscode-markdownlint",
+        "yzhang.markdown-all-in-one",
+        "shd101wyy.markdown-preview-enhanced",
+        "bpruitt-goddard.mermaid-markdown-syntax-highlighting",
+        "streetsidesoftware.code-spell-checker",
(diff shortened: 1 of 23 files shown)


grlap (approved):

amueller (approved):

sergiy-k:
@bpkroth, this is an impressive piece of work! Very cool!

PR #130, opened 2020-10-07
Adding OneHotEncoder HypergridAdapter

edcthayer:
OneHotEncoder Hypergrid Adapter allows categorical dimensions to be transformed to dummy variables using the sklearn OneHotEncoder.  By default, each categorical dimension will produce its own collection of OneHotEncoded dummy variables, but one may also generate a OneHotEncoding of the cross product of all categorical dimension categories using the `merge_all_categorical_dimensions` argument (as used in the RERF surrogate model).  

Once this adapter is approved, it will be integrated into the regression-enhanced random forest regression (RERF) model.


source/Mlos.Python/mlos/Spaces/Point.py
```diff
@@ -120,7 +120,7 @@ def from_json(cls, json_str):
     def to_dict(self):
         return_dict = {}
         for param_name, value in self:
-            if isinstance(value, Number) and int(value) == value:
+            if isinstance(value, Number) and int(value) == value and not isinstance(value, bool):
```
edcthayer, line 123: This was added to ensure bools were not being cast as ints.

source/Mlos.Python/mlos/Spaces/HypergridAdapters/HypergridAdapter.py
```diff
@@ -113,7 +113,7 @@ def _project_point(self, point: Point) -> Point:
         :return:
         """
         original_dataframe = point.to_dataframe()
-        projected_dataframe = self.project_dataframe(original_dataframe)
+        projected_dataframe = self._project_dataframe(original_dataframe)
```
byte-sculptor, line 116: This is another bug-fix. Thank you for including it in this PR. A few words of explanation for the benefit of other reviewers:
* project_dataframe (public, without leading underscore) recursively invokes all adapters in the stack. For example: it might first invoke HierarchicalToFlatHypergridAdapter, then the CategoricalToOneHotEncodedHypergridAdapter. 
* _project_dataframe (protected, with leading underscore) executes only the current adapter's projection. 

Similar logic applies to: project_point and _project_point.

Thus, inside of the `_project_point()` function, we wish to only perform the projection defined by `self`, which is accomplished by calling transforming a point to a dataframe, performing self._project_dataframe(), and then recovering the point from the dataframe.

source/Mlos.Python/mlos/Spaces/HypergridAdapters/unit_tests/TestCategoricalToOneHotEncodedHypergridAdapter.py
```diff
@@ -0,0 +1,173 @@
...
+        adapter = CategoricalToOneHotEncodedHypergridAdapter(adaptee=adaptee, merge_all_categorical_dimensions=True, drop='first')
+        self._test_projecting_dataframe_categorical_to_one_hot_encoding_point_from_adaptee(adapter=adapter, adaptee=adaptee, num_random_points=1000)
+
+    # Helper functions
+    def _test_projecting_dataframe_categorical_to_one_hot_encoding_point_from_adaptee(self, adaptee, adapter, num_random_points: int):
+        original_df = adaptee.random_dataframe(num_samples=num_random_points)
+        projected_df = adapter.project_dataframe(df=original_df, in_place=False)
+        self.assertTrue(id(original_df) != id(projected_df))
+        for column in adapter.get_one_hot_encoded_column_names():
+            self.assertTrue(projected_df[column].between(0, 1).all())
```
byte-sculptor, line 143: Would it be good to check if `projected_df[column].isin([0,1]).all()` instead of `between(0, 1)`?

source/Mlos.Python/mlos/Spaces/HypergridAdapters/unit_tests/TestCategoricalToOneHotEncodedHypergridAdapter.py
```diff
@@ -0,0 +1,173 @@
...
+#
+from numbers import Number
+
+import unittest
+
+from mlos.Spaces import SimpleHypergrid, CategoricalDimension, ContinuousDimension, DiscreteDimension, OrdinalDimension
+from mlos.Spaces.HypergridAdapters import CategoricalToOneHotEncodedHypergridAdapter
+
+
+class TestCategoricalToOneHotEncodedHypergridAdapter(unittest.TestCase):
```
byte-sculptor, line 13: The number of tests is really awesome :D

source/Mlos.Python/mlos/Spaces/HypergridAdapters/unit_tests/TestCategoricalToOneHotEncodedHypergridAdapter.py
```diff
@@ -0,0 +1,173 @@
...
+
+        # Now let's make sure that we can translate a bunch of points successfully.
+        #
+
+        for _ in range(num_random_points):
+            original_point = adaptee.random()
+            projected_point = adapter.project_point(original_point)
+            # Now let's ascertain that each coordinate in the projected point is in fact a number.
+            #
+            self.assertTrue(all(isinstance(dim_value, Number) for dim_name, dim_value in projected_point))
```
byte-sculptor, line 168: Could we add a check to make sure that:
1) all categorical dimensions were dropped,
2) all dummy_variables are in (0,1)?

source/Mlos.Python/mlos/Spaces/HypergridAdapters/CategoricalToOneHotEncodedHypergridAdapter.py
```diff
@@ -0,0 +1,260 @@
...
+        self._adaptee: Hypergrid = adaptee
+        self._merge_all_categorical_dimensions = merge_all_categorical_dimensions
+        self._one_hot_encoder_kwargs = {
+            'drop': drop,
+            'dtype': np.float64,
+            'sparse': False,
+            'handle_unknown': 'error'
+        }
+        self._all_one_hot_encoded_target_dimension_names = []
+        self._adaptee_to_target_data_dict = {}
```
byte-sculptor, line 66: A type annotation and a comment describing the purpose of this dictionary would help a lot.

byte-sculptor (approved):

source/.pylintrc
```diff
@@ -145,6 +145,7 @@ disable=print-statement,
         invalid-name,  # Eventually want to turn this on, just have a lot of errors currently, so didn't want to clutter things up.
         bare-except,
         len-as-condition,
+        pointless-string-statement   # To allow multi-line string comments
```
bpkroth, line 148: Where does this show up?

source/Mlos.Python/mlos/Spaces/HypergridAdapters/CategoricalToOneHotEncodedHypergridAdapter.py
```diff
@@ -0,0 +1,268 @@
...
+from pandas import DataFrame
+from sklearn.preprocessing import OneHotEncoder
+from mlos.Spaces import CategoricalDimension, DiscreteDimension, Hypergrid, SimpleHypergrid
+from mlos.Spaces.HypergridAdapters.HypergridAdapter import HypergridAdapter
+from mlos.Spaces.HypergridAdapters.HierarchicalToFlatHypergridAdapter import HierarchicalToFlatHypergridAdapter
+from mlos.Spaces.HypergridAdapters.CategoricalToDiscreteHypergridAdapter import CategoricalToDiscreteHypergridAdapter
+
+
+class CategoricalToOneHotEncodingAdapteeTargetMapping:
+    """ Retains the list of target Hypergrid's (one hot encoded) dimensions
```
bpkroth, line 16: @amueller to comment more, but I think these are supposed to not have the space separation in order to be processed by `sphinx` for API publication.

source/Mlos.Python/mlos/Spaces/HypergridAdapters/CategoricalToOneHotEncodedHypergridAdapter.py
```diff
@@ -0,0 +1,268 @@
...
+    @property
+    def adaptee(self) -> Hypergrid:
+        return self._adaptee
+
+    @property
+    def target(self) -> Hypergrid:
+        return self._target
+
+    def get_original_categorical_column_names(self):
+        return self._adaptee_to_target_data_dict.keys()
```
bpkroth, line 109: Could see also providing access to the the map for lookup instead of just the array of names, but that could always be added later when the need arises.

source/Mlos.Python/mlos/Spaces/HypergridAdapters/CategoricalToOneHotEncodedHypergridAdapter.py
```diff
@@ -0,0 +1,268 @@
...
+            dimensions=None,
+            random_state=self._adaptee.random_state
+        )
+
+        """ Details about construction of the target hypergrid:
+           1) Moving non-categorical dimensions to target, while collecting needed info about adaptee categorical dimensions
+           2) Since sklearn's OHE will handle both project and unproject dataframe transforms, prepare the OHE class.
+              This requires constructing the 'categories' argument for OHE (all categorical dims or 1 cross product dim).
+              The dimension's .linspace() method provides the order list of values but doesn't include possible np.NaN values,
+               hence that list is augmented to include the string 'nan' which pandas.DataFrame.apply(map(str)) will produce from a np.NaN value.
```
bpkroth, line 207:
```suggestion
              hence that list is augmented to include the string 'nan' which pandas.DataFrame.apply(map(str)) will produce from a np.NaN value.

```
?

bpkroth: Similar below.

source/.pylintrc
byte-sculptor: This lets us use multiline strings as comments.

bpkroth: Got that part.  I just didn't initially see where they were being used not as a function or class description.

For the record, here's an example:
https://github.com/microsoft/MLOS/pull/130/files#diff-72efa0a3dd17279337c95252efb4376bc4f6bb6efc8cc1897dacae6b4a1c87c7R202

bpkroth (approved):
I don't have any substantial feedback.
I think there are some comment styling things that will show up when we try to publish the API docs here: https://microsoft.github.io/MLOS/python_api/
I don't think we have instructions written up on testing that yourself yet.
The rough outline goes something like this:
```sh
cd website
# generate the Python APIs to be served at http://localhost:8080/MLOS/python_api/
make sphinx-site # repeat this one as necessary while the docker container is running
# generate the rest of the http://localhost:8080/MLOS site content
make hugo-site # optional
# start a local webserver
docker run -d --name mlos-website-link-checker -v $PWD:/src/MLOS/website -v $PWD/nginx.conf:/etc/nginx/conf.d/mlos.conf -p 8080:8080 nginx:latest
# now point your browser at http://localhost:8080/MLOS/python_api/ to double check things
```
Note: the instructions assume a Linux environment atm (e.g. via WSL2)

bpkroth:
> 
> 
> I don't have any substantial feedback.
> I think there are some comment styling things that will show up when we try to publish the API docs here: https://microsoft.github.io/MLOS/python_api/
> I don't think we have instructions written up on testing that yourself yet.
> The rough outline goes something like this:
> 
> ```shell
> cd website
> # generate the Python APIs to be served at http://localhost:8080/MLOS/python_api/
> make sphinx-site # repeat this one as necessary while the docker container is running
> # generate the rest of the http://localhost:8080/MLOS site content
> make hugo-site # optional
> # start a local webserver
> # borrowed from https://github.com/microsoft/MLOS/blob/main/website/test_site_links.sh#L48
> docker run -d --name mlos-website-link-checker -v $PWD:/src/MLOS/website -v $PWD/nginx.conf:/etc/nginx/conf.d/mlos.conf -p 8080:8080 nginx:latest
> # now point your browser at http://localhost:8080/MLOS/python_api/ to double check things
> ```
> 
> Note: the instructions assume a Linux environment atm (e.g. via WSL2)

Actually, now that I look at it, we seem to have neglected to transfer some API doc comments to some of the public calls after we added the `_` underscore prefix.

For example:
https://microsoft.github.io/MLOS/python_api/api/mlos.Spaces.HypergridAdapters.HypergridAdapter.html
https://github.com/Microsoft/MLOS/blob/732b9d7/source/Mlos.Python/mlos/Spaces/HypergridAdapters/HypergridAdapter.py#L98

We should fix that, though I don't think it needs to be done in this PR.

PR #133, opened 2020-10-12
Fixups for libstdc++-10 support

bpkroth:
Recently we discovered the following error in CI for Linux builds:

```txt
Scanning dependencies of target Mlos.Core
make[3]: Leaving directory '/home/runner/work/MLOS/MLOS/out/cmake/Debug'
make[3]: Entering directory '/home/runner/work/MLOS/MLOS/out/cmake/Debug'
[ 20%] Building CXX object source/Mlos.Core/CMakeFiles/Mlos.Core.dir/ArenaAllocator.cpp.o
In file included from /home/runner/work/MLOS/MLOS/source/Mlos.Core/ArenaAllocator.cpp:16:
In file included from /home/runner/work/MLOS/MLOS/source/Mlos.Core/Mlos.Core.h:90:
In file included from /home/runner/work/MLOS/MLOS/source/Mlos.Core/PropertyProxy.h:18:
In file included from /home/runner/work/MLOS/MLOS/source/Mlos.Core/ObjectSerialization.h:19:
/home/runner/work/MLOS/MLOS/source/Mlos.Core/StringTypes.h:89:38: error: use of undeclared identifier 'wcslen'
        Length = string != nullptr ? wcslen(string) : 0;
                                     ^
/home/runner/work/MLOS/MLOS/source/Mlos.Core/StringTypes.h:105:61: error: use of undeclared identifier 'wcsncmp'
    return lhs.Length == rhs.Length && (lhs.Length == 0 || (wcsncmp(lhs.Data, rhs.Data, lhs.Length) == 0));
                                                            ^
2 errors generated.
```

Not entirely sure how this snuck in since I can't find anything on the C++ side which has changed recently and tests were passing a few days ago, but here we are.

It seems that we weren't including the wide character string functions in the standard libs previously.  This small change should fix that.

### DIFF
--- a/source/Examples/SmartCache/Main.cpp
+++ b/source/Examples/SmartCache/Main.cpp
@@ -54,8 +54,8 @@ void ThrowIfFail(HRESULT hr)
 int
 __cdecl
 main(
-    __in int argc,
-    __in char* argv[])
+    _In_ int argc,
+    _In_ char* argv[])
 {
     UNUSED(argc);
     UNUSED(argv);
--- a/source/Examples/SmartSharedChannel/Main.cpp
+++ b/source/Examples/SmartSharedChannel/Main.cpp
@@ -51,8 +51,8 @@ void CheckHR(HRESULT hr)
 int
 __cdecl
 main(
-    __in int argc,
-    __in char* argv[])
+    _In_ int argc,
+    _In_ char* argv[])
 {
     UNUSED(argc);
     UNUSED(argv);
--- a/source/Mlos.Core/Mlos.Core.h
+++ b/source/Mlos.Core/Mlos.Core.h
@@ -56,7 +56,6 @@ constexpr int32_t INVALID_FD_VALUE = -1;
 
 #define _Check_return_
 #define _Out_
-#define __in
 #define _In_
 #define _In_z_
 
--- a/source/Mlos.Core/Mlos.Core.h
+++ b/source/Mlos.Core/Mlos.Core.h
@@ -24,6 +24,7 @@
 #include <functional>
 #include <type_traits>
 #include <string.h>
+#include <cwchar>
 
 // Undefine MIN MAX macros.
 //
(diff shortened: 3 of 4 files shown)


bpkroth:
Discussing with @grlap, this one-line fix isn't going to work due to the fact that `wchar_t` is of different lengths for Windows/Linux.

Some options to consider to use an explicit lengthed type (e.g. `char16_t`) or force it to a particular length with a compile flag (e.g. `-fshort-wchar`).

Still unclear what changed however.  None of the C++ code referencing this has changed since August.  I suspect something in the Github Ubuntu runners config has changed but I haven't seen what yet.  Can't reproduce the issue locally either.

bpkroth:
Figured it out: Github added `libstdc++-10-dev` to the runners:

https://github.com/actions/virtual-environments/commit/207f49ed9343b65a01c0750a2e552d05a3bd985b#diff-29b87ca86e8166171be2d4cf761d6e8f

bpkroth:
With the first fix we were getting type of errors.

```txt
/usr/bin/../lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/bits/basic_string.tcc:1501:63: error: expected expression
              const __ctype_type& __ct = use_facet<__ctype_type>(__in.getloc());
```

This was because we had cases of `__in` SAL notations in use that we were `#define`ing away on Linux.
However, for GCC-10's libstdc++ they use that token in some code generation so we were introducing errors by doing that.

https://github.com/microsoft/MLOS/pull/133/commits/6d318b0f37426776fa7568426eb6df473479ea8a fixes that by switching to `_In_` in the few remaining places.

That still doesn't address the `wchar_t` width issues though.

grlap (approved):

bpkroth:
For the record, the `wstrlen` issue showed up in the first place because of some more strict header requirements introduced in the new libstdc++ version:
https://gcc.gnu.org/gcc-10/porting_to.html

PR #143, opened 2020-10-22
2020/october/bug fixes

byte-sculptor:
This fixes a number of bugs and typos:
- ThreeLevelQuadratic: used only x_1 instead both x_1 and x_2
- BayesianFactory was not passing a logger to BayesianOptimizer
- idx_max() -> idxmax()
- checking containment in Enums used to throw a warning
- validating optima in TestBayesianOptimizer would randomly fail
- Point would cast bools to ints (Ed's change probably fixed it already)
- Trying to use an uninitialized Tracer used to cause a crash
- Global singletons could be overwritten

### DIFF
--- a/source/Mlos.Python/mlos/Optimizers/unit_tests/TestBayesianOptimizer.py
+++ b/source/Mlos.Python/mlos/Optimizers/unit_tests/TestBayesianOptimizer.py
@@ -472,5 +472,25 @@ class TestBayesianOptimizer(unittest.TestCase):
             # At the very least we can assert the ordering. Note that the configs corresponding to each of the below confidence bounds can be different, as confidence intervals
             # change width non-linearily both with degrees of freedom, and with prediction variance.
             #
-            assert lcb_99_ci_optimum.lower_confidence_bound <= lcb_95_ci_optimum.lower_confidence_bound <= lcb_90_ci_optimum.lower_confidence_bound <= predicted_optimum.predicted_value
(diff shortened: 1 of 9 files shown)


grlap (approved):

source/Mlos.Python/mlos/Optimizers/unit_tests/TestBayesianOptimizer.py
```diff
@@ -480,5 +480,25 @@ def validate_optima(self, optimizer: OptimizerBase):
             # At the very least we can assert the ordering. Note that the configs corresponding to each of the below confidence bounds can be different, as confidence intervals
             # change width non-linearily both with degrees of freedom, and with prediction variance.
             #
-            assert lcb_99_ci_optimum.lower_confidence_bound <= lcb_95_ci_optimum.lower_confidence_bound <= lcb_90_ci_optimum.lower_confidence_bound <= predicted_optimum.predicted_value
-            assert predicted_optimum.predicted_value <= ucb_90_ci_optimum.upper_confidence_bound <= ucb_95_ci_optimum.upper_confidence_bound <= ucb_99_ci_optimum.upper_confidence_bound
+            if not (lcb_99_ci_optimum.lower_confidence_bound <= lcb_95_ci_optimum.lower_confidence_bound <= lcb_90_ci_optimum.lower_confidence_bound <= predicted_optimum.predicted_value):
+                # If the the prediction for predicted_value has too few degrees of freedom, it's impossible to construct a confidence interval for it.
```
bpkroth, line 484: Mentioned elsewhere, but this comment seemed at odds with the if statement below (hard to have fewer degrees of freedom than 0, yet that the one where we check whether or not it's possible).

PR #144, opened 2020-10-22
Made Opitimizer Serializable

byte-sculptor:
Implementing __getstate__ and __setstate__ on Point makes all instances of Point serializable and by extension the entire optimizer.

### DIFF
--- a/source/Mlos.Python/mlos/Optimizers/unit_tests/TestBayesianOptimizer.py
+++ b/source/Mlos.Python/mlos/Optimizers/unit_tests/TestBayesianOptimizer.py
@@ -4,6 +4,7 @@
 #
 import math
 import os
+import pickle
 import random
 import unittest
 import warnings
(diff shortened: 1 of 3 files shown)


grlap (approved):

bpkroth:
Are pickle dumps stable?  My vague recollection from a while ago was that they wouldn't necessarily work across machines/versions?

More generally, is the eventual point here to store the optimizers? (e.g. MLFlow :)

byte-sculptor:
@bpkroth: I believe you are correct - pickling doesn't work across Python versions and I wouldn't be surprised if it breaks even within the same version but between machines etc.

So for storage in MLFlow we will have to evaluate other options. However, pickling is pretty fundamental and useful so it's good to restore this functionality. For example: recently Neha had a problem with a trained optimizer. It would have been useful for her to pickle the entire thing and send it to both Andreas and I for further debugging. 

Lastly, pickling will allow us to cache optimizers in the OptimizerMicroservice and to run multiple instances of the same optimizer repeatedly. The use case for this is parallelizing predictions. Today, to render a single Tomograph grid we must repeatedly query the same instance of the optimizer without changing its state. We could instead instantiate n_cpus optimizers and parallelize that entire read-only workload.

PR #151, opened 2020-10-26
Small feature additions to tracer and to logger

byte-sculptor:
Really small features that I've had to add to Tracer and to Logger. Basically, when after each optimization run, I want to have an EvaluationReport that contains the tracer events, and all the log records (along with goodness of fit metrics over time etc.). I also want to be able to pickle/unpickle them and to dump them to files. 

Furthermore, the evaluation uses multiprocessing, so capturing the PID in the Traces helps disambiguate events from different runs and render them pretty.

Tracer:
* we capture thread id and PID with traced events

Logger:
* we save the record as is, instead of turning it into a dictionary. This is a minor perf improvement, but it also allows us to use the formatter specified by the user when dumping records to file.
* we can dump traces to file now.

### DIFF
--- a/source/Mlos.Python/mlos/Logger.py
+++ b/source/Mlos.Python/mlos/Logger.py
@@ -23,22 +23,21 @@ class BufferingHandler(logging.StreamHandler):
 
     def emit(self, record):
         if self.level <= record.levelno:
-            event = {
-                'timestamp': record.asctime,
-                'level': record.levelname,
-                'filename': record.filename,
-                'line': record.lineno,
-                'function': record.funcName,
-                'message': record.message.replace("'", '"'),
-                'exception_text': record.exc_text.replace("'", '"') if record.exc_text is not None else None
-            }
-            self.buffered_log_records.append(event)
+            self.buffered_log_records.append(record)
 
     def get_records(self, clear_buffer=False):
         records = self.buffered_log_records
         if clear_buffer:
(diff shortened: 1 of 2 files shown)


bpkroth:
Just to be clear, this is initial work to capture some additional execution context trace to file that we can later on upload to some sort of github actions archive to help with unit test failure analysis.  However, it's not actually invoking the dump_to_file, nor the archive save yet, correct?

byte-sculptor:
> Just to be clear, this is initial work to capture some additional execution context trace to file that we can later on upload to some sort of github actions archive to help with unit test failure analysis. However, it's not actually invoking the dump_to_file, nor the archive save yet, correct?

Correct, this functionality would be used both by the scenario you describe (to help debug test failures) and by optimizer evaluation tools. And yes - the dump_to_file is not invoked anywhere.

bpkroth:
> 
> 
> > Just to be clear, this is initial work to capture some additional execution context trace to file that we can later on upload to some sort of github actions archive to help with unit test failure analysis. However, it's not actually invoking the dump_to_file, nor the archive save yet, correct?
> 
> Correct, this functionality would be used both by the scenario you describe (to help debug test failures) and by optimizer evaluation tools. And yes - the dump_to_file is not invoked anywhere.

Thanks.  I was looking it over before there was a description :)

bpkroth (approved):
LGTM

PR #154, opened 2020-10-31
use pytest to run python tests

amueller:
I still need to install pytest probably which I still have to figure out.

## Caveat
Right now the long haul test match is not equivalent. Future long haul tests might be missed.
Unfortunately there's no easy way to change the file matching pattern from the command line. We could always match all the tests and then select a subset with ``-k`` from the command line.
Or we could use marks (like a longhaul mark).
Either of these would mean that by default (i.e. without specifying additional command line arguments and just running pytest), the long haul tests would be included in the test run.

Another option would be to overwrite the config file in the CI script, but that seems a bit ugly as well?


.github/workflows/nightly.yml
```diff
@@ -119,7 +119,7 @@ jobs:
       timeout-minutes: 700
       run: |
         docker exec --workdir /src/MLOS/source/Mlos.Python mlos-build-ubuntu-${{ matrix.UbuntuVersion }} \
-          python3.7 -m unittest discover -p LongHaulTest*.py --verbose
+          pytest -sv mlos/Optimizers/long_haul_test/*
```
bpkroth, line 122: Do we can which version of python that runs?

Dockerfile
```diff
@@ -170,7 +170,7 @@ RUN /bin/bash /tmp/MLOS/scripts/install.python.sh && \
 
 RUN python3.7 -m pip install pip && \
     python3.7 -m pip install --upgrade pip && \
-    python3.7 -m pip install setuptools wheel
+    python3.7 -m pip install setuptools wheel pytest
```
bpkroth, line 173: Can you please add pylint here as well?  I hadn't realized we didn't include it yet (probably already in the base somehow).

pytest.ini
```diff
@@ -0,0 +1,2 @@
+[pytest]
+python_files = Test*.py
```
bpkroth, line 2: Should we include lower case here as well?

bpkroth (approved):

bpkroth:
> 
> 
> I still need to install pytest probably which I still have to figure out.
> ## Caveat
> 
> Right now the long haul test match is not equivalent. Future long haul tests might be missed.
> Unfortunately there's no easy way to change the file matching pattern from the command line. We could always match all the tests and then select a subset with `-k` from the command line.
> Or we could use marks (like a longhaul mark).
> Either of these would mean that by default (i.e. without specifying additional command line arguments and just running pytest), the long haul tests would be included in the test run.
> 
> Another option would be to overwrite the config file in the CI script, but that seems a bit ugly as well?

Can we add a different `pytest.ini` (e.g. `pytest.longhaul.ini`) somewhere else and reference it in the `pytest` invocation?

scripts/run-python-tests.sh
```diff
@@ -10,9 +10,4 @@ set -eu
 scriptdir=$(readlink -f "$(dirname "$0")")
 cd "$scriptdir/.."
 
-. "$scriptdir/util.sh"
-pythoncmd=`getPythonCmd`
-
-# Linux filesystems are case-sensitive, so we need to tell the python unit test
-# scanner to look for upper-case files as well:
-$pythoncmd -m unittest discover --verbose --locals --failfast -s source/Mlos.Python -p "[Tt]est*.py" $*
+pytest -svx source\Mlos.Python
```
bpkroth, line 13:
```suggestion
pytest -svx source/Mlos.Python
```

.github/workflows/nightly.yml
amueller: Do you mean care? Well it's the one that installed pytest. Generally I'm highly skeptical about not using conda/virtualenv as it's really hard to control what is run. We might be able to run via python3.7 but then if that's not where we installed it earlier, it's not gonna work? Generally we should make sure that pytest and mlos where installed in the same environment, and then use that same environment.

amueller:
Ah, right we can probably use a different pytest.ini and pass it. will fix that up.

.github/workflows/nightly.yml
bpkroth: Agreed, but I think we'd tabled switching to conda for a future PR (that hasn't bubbled to the surface yet).
In the meantime, I've been trying to just explicitly reference python3.7 everywhere to avoid those types of confusions.
I think we can do that here with `python3.7 -m pytest` right?
Anyways, just a thought, I have no strong convictions on this atm given we recognize that some sort of virtual environment is the better long term approach (and as long as what we use works for now).

pytest.ini
amueller: do we have both? (and if so, why ;)

amueller:
Actually, it doesn't look like you can set the config file from the command line.
We could do this:
https://docs.pytest.org/en/stable/example/simple.html#control-skipping-of-tests-according-to-command-line-option

which requires us to mark all long haul / slow tests, as I said before, but disable them by default.

pytest.ini
bpkroth: Probably laziness/cross-platform issues early on.
I was originally thinking better to catch accidental mistakes, but I think you're probably right, better to not let them proliferate.  I withdraw my comments :)

scripts/run-python-tests.sh
```diff
@@ -10,9 +10,4 @@ set -eu
 scriptdir=$(readlink -f "$(dirname "$0")")
 cd "$scriptdir/.."
 
-. "$scriptdir/util.sh"
-pythoncmd=`getPythonCmd`
-
-# Linux filesystems are case-sensitive, so we need to tell the python unit test
-# scanner to look for upper-case files as well:
-$pythoncmd -m unittest discover --verbose --locals --failfast -s source/Mlos.Python -p "[Tt]est*.py" $*
+pytest -svx source/Mlos.Python
```
bpkroth, line 13:
```suggestion
pytest -svx source/Mlos.Python $*
```

If we keep the trailing `$*` we can continue to pass `-k someTest` to through the script to be able to limit the set of tests it runs.

Given the command is much shorter/easier now, this is probably less of a concern as it was before though.

amueller:
I think we can leave the long haul test as they are right now. They could certainly be simplified, but I think this is a reasonable start :)

.github/workflows/nightly.yml
bpkroth: Actually, given other comments, should this continue to run with `unittest` for now?

amueller: I think this is fine, do you think there's an issue? I had thought about making this more elegant instead of using the particular file match, but I can live with this for now, and maybe we can later refactor the test.

bpkroth: Yeah, it's fine.  In a previous comment I thought you'd said you didn't fixup the LongHaul tests to use pytest instead of unittest yet, but I think that was out of date.

source/Mlos.Python/mlos/OptimizerEvaluationTools/unit_tests/TestOptimizerEvaluator.py
```diff
@@ -102,58 +101,58 @@ def test_defaults(self):
...
         with open(os.path.join(self.temp_dir, "goodness_of_fit.pickle"), "rb") as in_file:
             unpickled_gof = pickle.load(in_file)
 
         gof_df = unpickled_gof.get_goodness_of_fit_dataframe()
-        self.assertTrue(len(gof_df.index) > 0)
-        self.assertTrue(all((col_name in gof_df.columns.values or col_name == "data_set_type") for col_name in GoodnessOfFitMetrics._fields))
-        self.assertTrue(all(
+        assert len(gof_df.index) > 0
+        assert all((col_name in gof_df.columns.values or col_name == "data_set_type") for col_name in GoodnessOfFitMetrics._fields)
+        assert (all(
```
byte-sculptor, line 121: nit: you removed parentheses elsewhere. Is there a reason to keep this pair?

source/Mlos.Python/mlos/Optimizers/long_haul_tests/LongHaulTestBayesianOptimizer.py
```diff
@@ -304,19 +304,19 @@ def run_optimization(optimizer):
                 run_optimization(optimizer)
             best_config_point, best_objective = optimizer.optimum()
             print(f"Optimum config: {best_config_point}, optimum objective: {best_objective}")
-            self.assertLessEqual(sign * best_objective['function_value'], -5.5)
+            assert sign * best_objective['function_value'] < -5.5
```
byte-sculptor, line 307: <=

source/Mlos.Python/mlos/Spaces/unit_tests/TestIntervalTree.py
```diff
@@ -147,5 +146,5 @@ def enumerate_possibly_overlapping_continuous_intervals(self, interval, gap_widt
 
 
 
-class TestIntervalTreeWithDiscreteDimension(unittest.TestCase):
```
byte-sculptor, line 150: newline at the end of file

byte-sculptor:
Nice work!

byte-sculptor (approved):

source/Mlos.Python/mlos/OptimizerEvaluationTools/unit_tests/TestOptimizerEvaluator.py
amueller: Different regex and in general with linebreaks they might be necessary. Here they indeed can be removed.

Issue #33, opened 2020-09-02
Add code coverage checks


bpkroth:
Breaking out from #6:

It would be good to add code coverage checks and badges for that to the repo landing pages.

amueller:
using https://github.com/codecov/codecov-action
and https://www.codecov.io/ ?

bpkroth:
That works for me.  It's what hummingbird uses as well:
https://github.com/microsoft/hummingbird/blob/b8332d12fc82c1c9137a09e69776a1555da0176a/.github/workflows/pythonapp.yml#L75

amueller:
I wonder who recommended it to them ;)

PR #172, opened 2020-11-11
Add hypergrid adapter exposing SkLearn's PolynomialFeatures for adaptee continuous dimensions

edcthayer:
Bring SkLearn's PolynomialFeatures transform into MLOS as hypergrid adapter acting on continuous dimensions in the adaptee.

### DIFF
--- a/source/Mlos.Python/mlos/Spaces/HypergridAdapters/ContinuousToPolynomialBasisHypergridAdapter.py
+++ b/source/Mlos.Python/mlos/Spaces/HypergridAdapters/ContinuousToPolynomialBasisHypergridAdapter.py
@@ -14,6 +14,8 @@ from mlos.Spaces.HypergridAdapters.HierarchicalToFlatHypergridAdapter import Hie
 class ContinuousToPolynomialBasisHypergridAdapter(HypergridAdapter):
     """ Adds polynomial basis function features for each continuous dimension in the adaptee hypergrid using
         https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.PolynomialFeatures.html.
+        All non-continuous adaptee dimensions will be present in the target hypergrid.
+        Beware: Because HierarchicalHypergrids may have NaN values for some points, these NaNs will be replaced by zeros.
 
         Parameters
         ----------
@@ -43,7 +45,7 @@ class ContinuousToPolynomialBasisHypergridAdapter(HypergridAdapter):
         self._polynomial_features_kwargs = {
             'degree': degree,
             'interaction_only': interaction_only,
-            'include_bias': True,
+            'include_bias': False,
             'order': 'C'
         }
         self._target: Hypergrid = None
@@ -60,13 +62,20 @@ class ContinuousToPolynomialBasisHypergridAdapter(HypergridAdapter):
         self._num_dimensions_to_transform = len(self._adaptee_dimension_names_to_transform)
         self._adaptee_contains_dimensions_to_transform = self._num_dimensions_to_transform > 0
 
+        # see definition of _get_polynomial_feature_names() for usage
+        self._internal_feature_name_terminal_char = '_'
+
+        # Since sklearn PolynomialFeatures does not accept NaNs and these may appear in data frames from hierarchical hypergrids,
+        # the NaNs will be replaced with an imputed (finite) value.  The following sets the value used.
+        self._nan_imputed_finite_value = 0
+
         # instantiate sklearn's polynomial features instance
         self._polynomial_features = PolynomialFeatures(**self._polynomial_features_kwargs)
         # because the exact number of additional dimensions that will be added depends on the parameters to sklearn's PF,
         # *and* the sklearn PF instance above doesn't determine this information until after the .fit() method is called (requiring a dataframe),
         # *and* the target hypergrid can not be constructed without knowing the resulting number of continuous dimensions,
         # a trivial dataframe is constructed (all 1s) and .fit_transform() of _polynomial_features instance is called.
-        trivial_continuous_dim_x = np.array(np.ones((1, self._num_dimensions_to_transform))).reshape(1, -1)
+        trivial_continuous_dim_x = np.ones((1, self._num_dimensions_to_transform))
         trivial_polynomial_features_y = self._polynomial_features.fit_transform(trivial_continuous_dim_x)
         self._polynomial_features_powers = self._polynomial_features.powers_
         self._num_polynomial_basis_dimensions_in_target = trivial_polynomial_features_y.shape[1]
@@ -80,7 +89,8 @@ class ContinuousToPolynomialBasisHypergridAdapter(HypergridAdapter):
             random_state=self._adaptee.random_state
         )
 
-        # since linear terms will always be included in the polynomial basis functions, add all dimensions
+        # Add all adaptee dimensions to the target
+        # This aligns with this adapter's goal since the linear terms will always be included in the polynomial basis functions
         for adaptee_dimension in self._adaptee.dimensions:
             self._target.add_dimension(adaptee_dimension.copy())
 
@@ -90,10 +100,10 @@ class ContinuousToPolynomialBasisHypergridAdapter(HypergridAdapter):
         # add new dimensions to be created by sklearn PolynomialFeatures
 
         # construct target dim names using adaptee dim names and polynomial feature powers matrix
-        # The polynomial feature "feature names" look like: ['1', 'x0', 'x1', 'x0^2', 'x0 x1', 'x1^2']
-        # and this is independent of the adaptee dimension names (since the sklearn class doesn't take the x feature names as inputs).
-        poly_feature_dim_names = self._polynomial_features.get_feature_names().copy()
-
+        # This logic is worked out explicitly here so we have control over the derived dimension names.
+        # Currently, the code only substitutes adaptee feature names into the default feature_names produced by
+        # sklearn's PolynomialFeatures .get_feature_names() method.
+        poly_feature_dim_names = self._get_polynomial_feature_names()
         for i, poly_feature_name in enumerate(poly_feature_dim_names):
             ith_terms_powers = self._polynomial_features_powers[i]
 
@@ -101,18 +111,18 @@ class ContinuousToPolynomialBasisHypergridAdapter(HypergridAdapter):
                 # the constant term is skipped and the linear terms are already included in _target
                 continue
             else:
-                # replace adaptee dim names for poly feature name {x0, x1, ...} representatives
+                # replace adaptee dim names for poly feature name {x0_, x1_, ...} representatives
                 target_dim_name = poly_feature_name
                 for j, adaptee_dim_name in enumerate(self._adaptee_dimension_names_to_transform):
                     adaptee_dim_power = ith_terms_powers[j]
                     if adaptee_dim_power == 0:
                         continue
                     if adaptee_dim_power == 1:
-                        poly_feature_adaptee_dim_name_standin = f'x{j}'
+                        poly_feature_adaptee_dim_name_standin = f'x{j}{self._internal_feature_name_terminal_char}'
                         adaptee_dim_replacement_name = adaptee_dim_name
                     else:
                         # power > 1 cases
-                        poly_feature_adaptee_dim_name_standin = f'x{j}^{adaptee_dim_power}'
+                        poly_feature_adaptee_dim_name_standin = f'x{j}{self._internal_feature_name_terminal_char}^{adaptee_dim_power}'
                         adaptee_dim_replacement_name = f'{adaptee_dim_name}^{adaptee_dim_power}'
 
                     target_dim_name = target_dim_name.replace(poly_feature_adaptee_dim_name_standin, adaptee_dim_replacement_name)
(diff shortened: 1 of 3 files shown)


source/Mlos.Python/mlos/Spaces/HypergridAdapters/unit_tests/TestContinuousToPolynomialBasisHypergridAdapter.py
```diff
@@ -0,0 +1,210 @@
...
+
+        cls.balanced_hierarchical_hypergrid = ThreeLevelQuadratic().parameter_space
+
+    @staticmethod
+    def n_choose_k(n, k):
+        return math.factorial(n) / math.factorial(k) / math.factorial(n - k)
+
+    ## simple_hypergrid tests
+    # dataframe projections
+    def test_simple_hypergrid_degree_two_all_terms_dataframe(self):
```
amueller, line 59: You might simplify some of these tests with ``pytest.mark.parametrize`` instead of making them different test functions, like
```python
@pytest.mark.parametrize("degree", [2, 3])
@pytest.mark.parametrize("interaction_only", [TrueFalse])
@pytest.mark.parmaetrize("adaptee", [self.simple_hypergrid, self.unbalanced_hierarchical_hypergrid, adaptee=self.balanced_hierarchical_hypergrid])
def test_dataframe_projection(degree, interaction_only, adaptee):
    # code of test_dataframe_projection
```

and similar for test_point_projection.

amueller (commented):
looks good, some ideas to simplify maybe.

source/Mlos.Python/mlos/Spaces/HypergridAdapters/unit_tests/TestContinuousToPolynomialBasisHypergridAdapter.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def _test_dataframe_projection(self, adaptee, adapter_kwargs, num_random_points):
+        num_adaptee_continuous_dims = 0
+        for adaptee_dim in adaptee.dimensions:
+            if isinstance(adaptee_dim, ContinuousDimension):
+                num_adaptee_continuous_dims += 1
+
+        # count the number of polynomial terms expected excluding the constant term
+        num_target_continuous_dims_expected = self.n_choose_k(adapter_kwargs['degree'] + num_adaptee_continuous_dims, num_adaptee_continuous_dims) - 1
+        if adapter_kwargs['interaction_only']:
```
byte-sculptor, line 170: if `adapter_kwargs['interaction_only']` is true, we are overwriting the results of computation in 169. Can we use an if-else?

source/Mlos.Python/mlos/Spaces/HypergridAdapters/ContinuousToPolynomialBasisHypergridAdapter.py
```diff
@@ -0,0 +1,167 @@
...
+        return list(set.union(set(self._adaptee_dimension_names_to_transform), set(self._target_polynomial_feature_map.keys())))
+
+    def get_polynomial_feature_powers_table(self):
+        return self._polynomial_features_powers
+
+    def _project_dataframe(self, df: DataFrame, in_place=True) -> DataFrame:
+        if not in_place:
+            df = df.copy(deep=True)
+
+        # replace NaNs with zeros
```
byte-sculptor, line 145: Please document this somewhere. Perhaps in the class docstring.

source/Mlos.Python/mlos/Spaces/HypergridAdapters/ContinuousToPolynomialBasisHypergridAdapter.py
```diff
@@ -0,0 +1,167 @@
...
+            # add target dimension
+            # min and max are placed at -Inf and +Inf since .random() on the target hypergrid is generated on the original
+            # hypergrid and passed through the adapters.
+            self._target.add_dimension(
+                ContinuousDimension(name=target_dim_name, min=-math.inf, max=math.inf)
+            )
+            self._target_polynomial_feature_map[target_dim_name] = i
+
+    @property
+    def adaptee(self) -> Hypergrid:
```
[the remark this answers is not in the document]
byte-sculptor: adaptee is an abstract property in the HypergridAdapter base class. Basically forces all subclasses to have this field.

source/Mlos.Python/mlos/Spaces/HypergridAdapters/ContinuousToPolynomialBasisHypergridAdapter.py
```diff
@@ -0,0 +1,167 @@
...
+        return self._adaptee
+
+    @property
+    def target(self) -> Hypergrid:
+        return self._target
+
+    def get_column_names_for_polynomial_features(self):
+        return list(set.union(set(self._adaptee_dimension_names_to_transform), set(self._target_polynomial_feature_map.keys())))
```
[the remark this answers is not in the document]
byte-sculptor: I actually like the union better. Being explicit makes reading easier.

byte-sculptor:
Nice work!! :) 

The only substantial comment I have is that it would be good to write a little more tests :) Specifically, I'd recommend asserting that the values that come out of the adapter are actually computed correctly. Like making sure that if `x1 = 2` then `x1^2=4`. Also that the dimensions that shouldn't have been modified were left intact.

source/Mlos.Python/mlos/Spaces/HypergridAdapters/ContinuousToPolynomialBasisHypergridAdapter.py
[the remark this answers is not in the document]
edcthayer: How does one get SkLearn's PolynomialFeatures to accommodate NaNs?

source/Mlos.Python/mlos/Spaces/HypergridAdapters/ContinuousToPolynomialBasisHypergridAdapter.py
```diff
@@ -0,0 +1,167 @@
...
+            self._target.add_dimension(adaptee_dimension.copy())
+
+        if not self._adaptee_contains_dimensions_to_transform:
+            return
+
+        # add new dimensions to be created by sklearn PolynomialFeatures
+
+        # construct target dim names using adaptee dim names and polynomial feature powers matrix
+        # The polynomial feature "feature names" look like: ['1', 'x0', 'x1', 'x0^2', 'x0 x1', 'x1^2']
+        # and this is independent of the adaptee dimension names (since the sklearn class doesn't take the x feature names as inputs).
```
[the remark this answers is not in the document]
edcthayer: When Adam and I spoke about the derived dimension names, I heard concern about including white space or potentially reserved characters in these names.  So wanted to have the logic defined to perform the substitutions ourselves.  As it is written, we're using '*', '^', ' ' since that's what PolynomialFeatures uses, but this can be changed easily enough now.  Thanks for pointing out that get_feature_names accepts the original field names though.  Great to know for the future.

source/Mlos.Python/mlos/Spaces/HypergridAdapters/ContinuousToPolynomialBasisHypergridAdapter.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def _build_simple_hypergrid_target(self) -> None:
+        self._target = SimpleHypergrid(
+            name=self._adaptee.name,
+            dimensions=None,
+            random_state=self._adaptee.random_state
+        )
+
+        # since linear terms will always be included in the polynomial basis functions, add all dimensions
+        for adaptee_dimension in self._adaptee.dimensions:
```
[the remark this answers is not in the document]
edcthayer: Correct.  This adapter will propagate all dimensions present in the adaptee *and* add the polynomial features for any adaptee continuous dimensions.  I'll make that more clear in the comments here and in the adapter's documentation.  Thanks.

source/Mlos.Python/mlos/Spaces/HypergridAdapters/unit_tests/TestContinuousToPolynomialBasisHypergridAdapter.py
edcthayer: Happy to use this strategy going forward.  @byte-sculptor do you agree?

source/Mlos.Python/mlos/Spaces/HypergridAdapters/ContinuousToPolynomialBasisHypergridAdapter.py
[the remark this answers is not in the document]
amueller: a matter of taste I guess.

source/Mlos.Python/mlos/Spaces/HypergridAdapters/ContinuousToPolynomialBasisHypergridAdapter.py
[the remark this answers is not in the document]
amueller: ah, thanks for the explanation. I think often that's not programatically enforces in Python. If you want to enforce it, I think the standard way is metaclasses. But no strong opinion.

source/Mlos.Python/mlos/Spaces/HypergridAdapters/ContinuousToPolynomialBasisHypergridAdapter.py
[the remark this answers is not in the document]
amueller: Does it raise an error now? hm I thought we fixed that, but maybe we haven't. Then this makes a lot of sense.

source/Mlos.Python/mlos/Spaces/HypergridAdapters/ContinuousToPolynomialBasisHypergridAdapter.py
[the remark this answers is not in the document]
byte-sculptor: In my experience so far, the interpreter errors out when you try to instantiate an object that fails to define all abstract methods and properties. This of course doesn't prove it's universally true. 

Metaclasses sound like a good idea, but they are definitely outside the scope of this PR.

source/Mlos.Python/mlos/Spaces/HypergridAdapters/unit_tests/TestContinuousToPolynomialBasisHypergridAdapter.py
byte-sculptor: Looks quite neat :) I wouldn't block this PR on it, but down the road we should incorporate it. I think it can improve our test coverage.

Speaking of: @amueller what would it take to enable code coverage checks in GCI?

bpkroth: #33 has some notes on that

edcthayer: @amueller, @bpkroth , Am getting `NameError: name 'pytest' is not defined`.  How's this fixed?

edcthayer: @byte-sculptor helped on this question, so please ignore my question. :)

source/Mlos.Python/mlos/Spaces/HypergridAdapters/unit_tests/TestContinuousToPolynomialBasisHypergridAdapter.py
```diff
@@ -208,3 +223,32 @@ def _test_point_projection(adaptee, adapter_kwargs, num_random_points):
             projected_point = adapter.project_point(original_point)
             unprojected_point = adapter.unproject_point(projected_point)
             assert original_point == unprojected_point
+
+    @staticmethod
```
byte-sculptor, line 227: Thank you!

byte-sculptor (approved):

source/Mlos.Python/mlos/Spaces/HypergridAdapters/ContinuousToPolynomialBasisHypergridAdapter.py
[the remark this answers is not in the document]
amueller: Oh yes it does that. I think people are usually fine with erroring when any method tries to access a non-existing attribute. That means that the error is even later, but also you might get some functionality from the parent class without implementing all of the interface. Python code is often very casual about interfaces and inheritance. Not saying that's necessarily a good thing ;)

source/Mlos.Python/mlos/Spaces/HypergridAdapters/unit_tests/TestContinuousToPolynomialBasisHypergridAdapter.py
amueller: locally you can just set do ``pytest --cov-report html --cov=mlos``, the slightly trickier part is having the results hosted as a website, see the issue for a solution.

amueller (approved):

source/Mlos.Python/mlos/Spaces/HypergridAdapters/unit_tests/TestContinuousToPolynomialBasisHypergridAdapter.py
amueller: I wouldn't necessarily block on it, but it makes the tests much shorter and easier to read in this instance and reduces the potential for mistakes by being much more consise.

source/Mlos.Python/mlos/Spaces/HypergridAdapters/ContinuousToPolynomialBasisHypergridAdapter.py
```diff
@@ -0,0 +1,189 @@
...
+        # This aligns with this adapter's goal since the linear terms will always be included in the polynomial basis functions
+        for adaptee_dimension in self._adaptee.dimensions:
+            self._target.add_dimension(adaptee_dimension.copy())
+
+        if not self._adaptee_contains_dimensions_to_transform:
+            return
+
+        # add new dimensions to be created by sklearn PolynomialFeatures
+
+        # construct target dim names using adaptee dim names and polynomial feature powers matrix
```
amueller, line 102: I'm not sure why this is necessary when passing the adaptee feature names to ``get_feature_names``.

PR #175, opened 2020-11-12
Registering multiple objectives

byte-sculptor:
This PR is a first step towards multi-objective optimization.

It's really simple: the user can now define an objective space that's multi-dimensional. Subsequently the user can register observations that contain values for one or more of those objectives. 

Still, only one objective is optimized. Effectively, all those extra objectives are remembered, but ignored.

### DIFF
--- a/source/Mlos.Python/mlos/Optimizers/BayesianOptimizer.py
+++ b/source/Mlos.Python/mlos/Optimizers/BayesianOptimizer.py
@@ -48,7 +48,7 @@ class BayesianOptimizer(OptimizerBase):
         OptimizerBase.__init__(self, optimization_problem)
 
         # Since the optimization_problem.objective_space can now be multi-dimensional (as a milestone towards multi-objective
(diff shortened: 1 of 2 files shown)


source/Mlos.Python/mlos/Optimizers/BayesianOptimizer.py
```diff
@@ -41,11 +41,23 @@ def __init__(
...
             logger = create_logger("BayesianOptimizer")
         self.logger = logger
+
         # Let's initialize the optimizer.
         #
         assert len(optimization_problem.objectives) == 1, "For now this is a single-objective optimizer."
         OptimizerBase.__init__(self, optimization_problem)
 
+        # Since the optimization_problem.objective_space can now be multi-dimensional (as a milestone towards multi-objective
+        # optimization), we have to prepare a smaller search space for the surrogate model.
```
byte-sculptor, line 51: smaller objective space *

edcthayer (approved):

Issue #23, opened 2020-09-01
Enable gcc support


bpkroth:
Currently to build C++ code generation code we rely on `clang`s support for MSVC attributes (e.g. to ignore duplicate definitions at link time).

There are multiple ways around this that would help enable `gcc` support including reorganizing the code generation output or using some macros to add `#ifdef` wrappers around the attributes usages.

Consider this as icing/wishlist for now.

PR #176, opened 2020-11-13
Initial work to add gcc support

bpkroth:
Initial work to enable gcc support #23 

Started as a hack to workaround cmake's FetchContent not appearing to pass the CMAKE_COMPILER_ID through when building with CXX=clang++-10

This also fixes a few minor issues that gcc complained about.

- [x] Enable CI pipeline tests

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -227,9 +227,6 @@ jobs:
             cc: clang-10
           - cxx: g++
             cc: gcc
-        exclude:
-          - UbuntuVersion: 16.04
-            cxx: g++
     steps:
     - name: Checkout
       uses: actions/checkout@v2
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -213,7 +213,7 @@ jobs:
         #load: true
 
   linux-build-test:
-    name: Ubuntu ${{ matrix.UbuntuVersion }} ${{ matrix.configuration }} build/test C++/C#
+    name: Ubuntu ${{ matrix.UbuntuVersion }} ${{ matrix.configuration }} build/test C++/C# (${{ matrix.cxx }})
     runs-on: ubuntu-latest
     needs: [prep-vars, docker-image-cached-build]
     timeout-minutes: 30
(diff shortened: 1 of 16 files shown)


source/Examples/SmartCache/Main.cpp
```diff
@@ -52,7 +52,7 @@ void ThrowIfFail(HRESULT hr)
 }
 
 int
-__cdecl
```
bpkroth, line 55: See if we can remove this

bpkroth:
Per discussion with @grlap, the other (better) approach that I was originally thinking to remove the duplicate definitions by splitting the the declaration from the definition should also be possible.
Originally, I was concerned that that made for a more complicated build/integration scenario, however we already require the developer to `#include Mlos.Platform.h` in a single place (e.g. `main.cpp`).
We would essentially do the same thing for the `SettingsProvider_gen_callbacks.h` file.
Requires a bit more changes to the codegen first though.

.github/workflows/main.yml
```diff
@@ -213,14 +213,23 @@ jobs:
...
         configuration: [Debug, Release]
+        cxx: [clang++-10, g++]
+        include:
+          - cxx: clang++-10
+            cc: clang-10
+          - cxx: g++
+            cc: gcc
+        exclude:
+          - UbuntuVersion: 16.04
+            cxx: g++
```
bpkroth, line 232: g++ on Ubuntu 16.04 is quite old and fails with some additional errors.
Rather than debug them, for the moment, I simply disabled it.

bpkroth: @grlap says we can actually enable this too by fixing `constexpr`

bpkroth: @grlap I pulled some of your changes in and added a few others to fix some `std::array` initialization syntax that `gcc` on 16.04 was complaining about.

bpkroth:
Note: since this adds new build jobs to the pipelines, we'll need to change the CI requirements to let this through.

grlap (approved):
LGTM

PR #179, opened 2020-11-17
Increase Linux cmake-test timeouts in pipeline

bpkroth:


bpkroth:
I saw another long stall (hours in duration) elsewhere.
There might be an actual deadlock introduced somewhere here ...

bpkroth:
Haven't seen this issue for a bit, and the Python ones seem to be more pressing.  Closing for now.

PR #181, opened 2020-11-18
Computing Pareto

byte-sculptor:
This PR introduces a simple algorithm for computing a pareto frontier, given an objectives dataframe and a corresponding OptimizationProblem. 

You will find the algorithm in ParetoFrontier. This class will later be extended to maintain a cache of the latest pareto, and to iteratively add to it as more observations roll in. It will be owned by a BayesianOptimizer.

I also wrote some tests.


source/Mlos.Python/mlos/Optimizers/ParetoFrontier.py
```diff
@@ -0,0 +1,95 @@
...
+        objectives_df: pd.DataFrame
+    ):
+        assert len(features_df.index) == len(objectives_df.index)
+        if features_df is not None:
+            assert all(column in optimization_problem.feature_space.dimension_names for column in features_df.columns)
+
+        if objectives_df is not None:
+            assert all(column in optimization_problem.objective_space.dimension_names for column in objectives_df.columns)
+
+        # TODO: maybe prealloate larger arrays...
```
byte-sculptor, line 42: Remove the TODO

source/Mlos.Python/mlos/Optimizers/unit_tests/TestParetoFrontier.py
```diff
@@ -0,0 +1,282 @@
...
+
+        We want to be able to do that for an arbitrary number of dimensions so as to extract maximum coverage from this simple test.
+
+
+        How the test works?
+        -------------------
+        For N objectives we will specify the following parameters:
+            1. radius - distance of a point from origin.
+            2. theta0, theta1, ..., theta{i}, ..., theta{N-1} - angle between the radius segment and the and the hyperplane containing
+                unit vectors along y0, y1, ..., y{i}
```
byte-sculptor, line 105:
```suggestion
                unit vectors along y0, y1, ..., y{i-1}

```

source/Mlos.Python/mlos/Optimizers/ParetoFrontier.py
byte-sculptor, line 42:
```suggestion

```

source/Mlos.Python/mlos/Optimizers/ParetoFrontier.py
```diff
@@ -0,0 +1,91 @@
...
+        2. Objective function values
+        3. Possibly context values
+
+    A point belongs to a pareto frontier if it is not dominated by any other point. So if two points have the exact same values
+    for all objectives (but possibly different configurations), we will consider both of them to be pareto efficient.
+
+
+    """
+
+    def __init__(
```
byte-sculptor, line 28: None of the state is in use at the moment. I will populate it once I add this to the BayesianOptimizer.

source/Mlos.Python/mlos/Optimizers/unit_tests/TestParetoFrontier.py
```diff
@@ -0,0 +1,288 @@
...
+        objectives_df = pd.DataFrame()
+
+        for i in range(num_output_dimensions-1):
+            objectives_df[f'y{i}'] = prefix * np.cos(random_params_df[f'spherical_coordinates.theta{i}'])
+            prefix = prefix * np.sin(random_params_df[f'spherical_coordinates.theta{i}'])
+
+        # Conveniently, by the time the loop exits, the prefix is the value of our last objective.
+        #
+        if minimize == "all":
+            # Must flip the prefix first, since an even number of sin factors contributed to it.
```
byte-sculptor, line 233:
```suggestion
            # Must flip the prefix first, since there was no negative cosine to do it for us.

```

source/Mlos.Python/mlos/Optimizers/unit_tests/TestParetoFrontier.py
```diff
@@ -0,0 +1,288 @@
...
+        objectives_df = pd.DataFrame()
+
+        for i in range(num_output_dimensions-1):
+            objectives_df[f'y{i}'] = prefix * np.cos(random_params_df[f'spherical_coordinates.theta{i}'])
+            prefix = prefix * np.sin(random_params_df[f'spherical_coordinates.theta{i}'])
+
+        # Conveniently, by the time the loop exits, the prefix is the value of our last objective.
+        #
+        if minimize == "all":
+            # Must flip the prefix first, since there was no negative cosine to do it for us.
```
byte-sculptor, line 233:
```suggestion
            # Must flip the prefix first, since there was no negative cosine to do it for us.
```

source/Mlos.Python/mlos/Optimizers/ParetoFrontier.py
```diff
@@ -0,0 +1,91 @@
...
+        #
+        pareto_df = objectives_df.copy(deep=True)
+
+        # First, let's turn it into a maximization problem, by flipping the sign of all objectives that are to be minimized.
+        #
+        for objective in optimization_problem.objectives:
+            if objective.minimize:
+                pareto_df[objective.name] = -pareto_df[objective.name]
+
+        # By presorting we guarantee, that all dominated points are below the currently considered point.
```
amueller, line 69: That's only true if no objective value appears twice, right?

source/Mlos.Python/mlos/Optimizers/unit_tests/TestParetoFrontier.py
```diff
@@ -0,0 +1,287 @@
...
+        points on the perimeter. Any point that is between that secant and the perimeter is not dominated and would thus be pareto
+        efficient as well. (Actually even more points are pareto efficient, but this subset is easiest to explain in text).
+
+
+        We want to test scenarios where:
+            1) all objectives are maximized,
+            2) all objectives are minimized,
+            3) some objectives are maximized and some are minimized.
+
+        We want to be able to do that for an arbitrary number of dimensions so as to extract maximum coverage from this simple test.
```
amueller, line 94: Just as a side note, in higher dimensions, nearly all the points will be on the pareto frontier I think.

source/Mlos.Python/mlos/Optimizers/unit_tests/TestParetoFrontier.py
```diff
@@ -0,0 +1,287 @@
...
+
+            # Let's also make sure that no point on the pareto is dominated by any other point there.
+            #
+            other_rows = pareto_df.index.difference([i])
+            assert (pareto_df.loc[other_rows] > row).all(axis=1).sum() == 0
+
+
+    @pytest.mark.parametrize("minimize", ["all", "none", "some"])
+    @pytest.mark.parametrize("num_output_dimensions", [2, 10])
+    @pytest.mark.parametrize("num_points", [100, 1000])
```
amueller, line 74: nice!

amueller (commented):
Can you please add a test with duplicated objective values, say by using integers?

source/Mlos.Python/mlos/Optimizers/ParetoFrontier.py
```diff
@@ -0,0 +1,91 @@
...
+        :return:
+        """
+
+        assert all(column in optimization_problem.objective_space.dimension_names for column in objectives_df.columns)
+
+        # Let's copy it, since we are going to mess it up.
+        #
+        pareto_df = objectives_df.copy(deep=True)
+
+        # First, let's turn it into a maximization problem, by flipping the sign of all objectives that are to be minimized.
```
byte-sculptor, line 63:
```suggestion
        # First, let's turn it into a maximization maximization problem, by flipping the sign of all objectives that are to be minimized.

```

source/Mlos.Python/mlos/Optimizers/ParetoFrontier.py
byte-sculptor: At the time you made this comment, the sort was (erroneously) using only the first objective. Nice catch :)

I believe that sorting on all objectives fixes the problem. I got it to produce this:

image


@edcthayer: during our conversation yesterday we considered the green points:

On the one hand they are on the frontier line, and we will know the corresponding configuration.

On the other hand they are dominated in that we could improve one of the objectives, without sacrificing another. Excluding them, would produce a smaller dataset for the user, though I really expect this to be a rare occurrence as most objectives will likely be continuous and a probability of getting the exact same float value twice are slim.

So, should we include these or not?

byte-sculptor: Talked with Ed. We decided that including these points is better for 3 reasons:
1. If they were jittered only a bit to the north-east we would want to include them.
2. The code to compute this is more efficient.
3. It's a very rare situation in real life, so if the user doesn't appreciate these points, chances are that they will never see them :)

source/Mlos.Python/mlos/Optimizers/unit_tests/TestParetoFrontier.py
byte-sculptor: Yeah.

byte-sculptor:
> Can you please add a test with duplicated objective values, say by using integers?

@amueller: done

source/Mlos.Python/mlos/Optimizers/unit_tests/TestParetoFrontier.py
```diff
@@ -0,0 +1,336 @@
...
+            ]
+        )
+
+        expected_pareto_df = pd.DataFrame(
+            [
+                [1, 2],
+                [1, 2],
+                [2, 1],
+                [0.5, 2],
+                [1, 1],
```
amueller, line 317: I'm not sure I understand this. Shouldn't [1, 2] and [2, 1] dominate all the other points?

source/Mlos.Python/mlos/Optimizers/ParetoFrontier.py
amueller: Yes, I agree this is an edge case for sure, though we should make sure to handle it semi-reasonable :)

I wonder if these are weak pareto points? https://en.wikipedia.org/wiki/Pareto_efficiency
I need to reread the wikipedia article. Maybe adding a comment to the test and the class might be good?

amueller (approved):

source/Mlos.Python/mlos/Optimizers/ParetoFrontier.py
byte-sculptor: Actually, I think sorting only on the first dimension would work again, but I'll leave it in place in case we ever want to exclude the "maybe" points.

amueller:
It looks like generally the dominated points are not included in the "pareto frontier": https://en.wikipedia.org/wiki/Pareto_efficiency (going by the definitions in the first paragraph).
However, I think it's still nice to include them here, maybe add to the docs that the weak pareto optimal points are included?

byte-sculptor:
> It looks like generally the dominated points are not included in the "pareto frontier": https://en.wikipedia.org/wiki/Pareto_efficiency (going by the definitions in the first paragraph).
> However, I think it's still nice to include them here, maybe add to the docs that the weak pareto optimal points are included?

Sure

PR #185, opened 2020-11-19
Avoid some broken cmake package versions in Docker image build

bpkroth:
At the moment the cmake apt repo we rely on has published newer cmake packages that specify a dependency that doesn't exist.  This breaks a clean cmake install (as we do during docker image build).

For the moment, we're adding that version to a rejectlist to workaround the issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -175,7 +175,7 @@ docker-image:
 	    --build-arg=http_proxy=${http_proxy} \
 	    -t mlos-build-ubuntu-$(UbuntuVersion)
 	@ echo Finished building mlos-build-ubuntu-$(UbuntuVersion) image.
-	@ echo Run "docker run -v $$PWD:/src/MLOS --name mlos-build mlos-build-ubuntu-$(UbuntuVersion)" to start an instance.
+	@ echo "Run 'docker run -v $$PWD:/src/MLOS --name mlos-build mlos-build-ubuntu-$(UbuntuVersion)' to start an instance."
 
 # Cleanup the outputs produced by cmake.
 .PHONY: cmake-distclean
--- a/Makefile
+++ b/Makefile
@@ -169,7 +169,7 @@ endif
 .PHONY: docker-image
 docker-image:
 	docker pull ghcr.io/microsoft-cisl/mlos/mlos-build-ubuntu-$(UbuntuVersion):latest
-	docker build . --target $(MlosBuildImageTarget) \
+	docker build . $(DOCKER_BUILD_ARGS) --target $(MlosBuildImageTarget) \
 	    --build-arg=MlosBuildBaseArg=$(MlosBuildBaseArg) \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -169,13 +169,13 @@ endif
 .PHONY: docker-image
 docker-image:
 	docker pull ghcr.io/microsoft-cisl/mlos/mlos-build-ubuntu-$(UbuntuVersion):latest
-	docker build . --target $(MlosBuildImageTarget) \
+	docker build . $(DOCKER_BUILD_ARGS) --target $(MlosBuildImageTarget) \
```
bpkroth, line 172: A convenience variable to allow me to pass `--pull` to the `docker build` command for forced refresh of the underlying local image caches.

Makefile
```diff
@@ -169,13 +169,13 @@ endif
...
 	    --build-arg=MlosBuildBaseArg=$(MlosBuildBaseArg) \
 	    --build-arg=UbuntuVersion=$(UbuntuVersion) \
 	    --build-arg=http_proxy=${http_proxy} \
 	    -t mlos-build-ubuntu-$(UbuntuVersion)
 	@ echo Finished building mlos-build-ubuntu-$(UbuntuVersion) image.
-	@ echo Run "docker run -v $$PWD:/src/MLOS --name mlos-build mlos-build-ubuntu-$(UbuntuVersion)" to start an instance.
+	@ echo "Run 'docker run -v $$PWD:/src/MLOS --name mlos-build mlos-build-ubuntu-$(UbuntuVersion)' to start an instance."
```
bpkroth, line 178: quote the command the status message outputs for easier copy/paste

scripts/install.cmake.sh
```diff
@@ -10,6 +10,17 @@ set -eu
 if ! [ -x /usr/bin/cmake ] || [ "${1:-}" == '--force-upgrade' ]; then
     wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | gpg --dearmor - | sudo tee /etc/apt/trusted.gpg.d/kitware.gpg >/dev/null
     set -x
+
+    sudo tee /etc/apt/preferences.d/avoid-broken-cmake-repo-package-versions <<'APT_PREFERENCES'
+Package: cmake
+Pin: version 3.19.0-0kitware1ubuntu20.04.1
```
bpkroth, line 16: This version doesn't actually exist in the upstream repos yet.
When it does, this problem will go away.
In the meantime however, we still want to install the previous version of cmake (3.18.4...).  For that to work we need to rejectlist the higher version of `cmake-data`.
With that in place, when the version above of `cmake` appears it will depend on the corresponding version of `cmake-data`, which we would have rejectlisted, causing a new but related error.
Thus, for now, we simply rejectlist both of them.

amueller (approved):

PR #187, opened 2020-11-19
first shot at adding ~coveralls~ codecov integration

amueller:


Dockerfile
```diff
@@ -175,6 +175,7 @@ RUN python3.7 -m pip install pip && \
 COPY ./source/Mlos.Python/requirements.txt /tmp/
 RUN python3.7 -m pip install -r /tmp/requirements.txt
 
+
```
bpkroth, line 178:
```suggestion
```

bpkroth: Or do we need `pip install pytest-cov`?

amueller: yes

amueller:
right now the PR coverage diff isn't working because there's no coverage for the base branch. So I think we should merge so that we can get coverage for that?

I have to see how to enable it as a CI check, but we already have file-level coverage for this PR: https://codecov.io/gh/microsoft/MLOS/pull/187/tree?path=source%2FMlos.Python%2Fmlos

amueller:
Not sure if you can see it, but in the settings here:
https://app.codecov.io/gh/microsoft/MLOS/settings

it tells me to enable something in the MS org which I don't have access to :-/

bpkroth:
> 
> 
> Not sure if you can see it, but in the settings here:
> https://app.codecov.io/gh/microsoft/MLOS/settings
> 
> it tells me to enable something in the MS org which I don't have access to :-/

No, I get a "sorry you don't have access to this", but I didn't try very hard ...

We should check with @ksaur about what they did for the hummingbird repo.

bpkroth:
Related: #33 (though perhaps not all of it)

ksaur:
@bpkroth  Hmm....I don't remember having to enable something special.  Ping me tomorrow and we can try to troubleshoot.  (I just checked that I can verify that link above if I `s/MLOS/hummingbird/g`, but I'm not sure what I did to enable it.)

amueller:
btw looks like we don't need to have a yaml file:
https://docs.codecov.io/docs/codecov-yaml

When I go to the config page, at the top it says "Github Integration is installed. However, this repository is not enabled.".

amueller:
I still think we should merge this as a first pass btw.

bpkroth:
Do you want to add a badge to the front page, or leave that for later?
https://app.codecov.io/gh/microsoft/MLOS/settings/badge

.github/workflows/main.yml
```diff
@@ -536,6 +536,11 @@ jobs:
     - name: Run Python unit tests (Ubuntu ${{ matrix.UbuntuVersion }})
       timeout-minutes: 30
       run: docker exec mlos-build-python-${{ matrix.UbuntuVersion }} make python-test
+    - name: Upload coverage to codecov
+      uses: codecov/codecov-action@v1
+      with:
+        file: ./coverage.xml
+        flags: unittests
```
bpkroth, line 543:
```suggestion
        flags: unittests
        fail_ci_if_error: true
```

.github/workflows/main.yml
```diff
@@ -536,6 +536,11 @@ jobs:
     - name: Run Python unit tests (Ubuntu ${{ matrix.UbuntuVersion }})
       timeout-minutes: 30
       run: docker exec mlos-build-python-${{ matrix.UbuntuVersion }} make python-test
+    - name: Upload coverage to codecov
```
bpkroth, line 539:
```suggestion
    - name: Upload coverage to codecov
      if: ${{ github.repository == 'microsoft/mlos' }}
```

bpkroth: To prevent uploads when the actions are run from a forked repo during testing.

amueller:
added the badge: https://github.com/microsoft/MLOS/blob/97249bc5a7ef467f8eeaeedf8cc262e9bb1e3e56/README.md

.github/workflows/main.yml
bpkroth: Although, would those just go to their own codecov collection?  If so, maybe we don't care and that's useful for testing as well, so nevermind.

.github/workflows/main.yml
bpkroth: Undecided about this.  Seems more useful if we're providing a badge, but then we're dependent upon an external service.

bpkroth (approved):
Thanks!

PR #188, opened 2020-11-19
be more loosy-goosy with the test thresholds for less flaky tests

amueller:
cc @bpkroth @byte-sculptor current tests rely too much on randomness, we can
a) change the thresholds (this PR)
b) run the optimizer for more iterations (how many? takes longer)
c) fix random seeds

I think this is the easiest fix to keep it somewhat meaningful and fast and not have master failing a lot.


bpkroth:
> 
> 
> cc @bpkroth @byte-sculptor current tests rely too much on randomness, we can
> a) change the thresholds (this PR)
> b) run the optimizer for more iterations (how many? takes longer)
> c) fix random seeds
> 
> I think this is the easiest fix to keep it somewhat meaningful and fast and not have master failing a lot.

Seems fine for now.  I think this is not the only place for this unfortunately.

One other thing we discussed in chat was to keep the randomness to get better coverage, but to print the seed used out prior to execution so that we could at least hack together a reproduction locally for debugging if need be.

bpkroth (approved):

PR #190, opened 2020-11-20
Remove quotes from a docker label

bpkroth:
Should fix this error that recently showed up in our docker/build-push-action step:

"Error: Invalid Opening Quote: a quote is found inside a field at line 2"


bpkroth:
Here's an example of the error:
https://github.com/microsoft/MLOS/runs/1428242412?check_suite_focus=true#step:6:26

amueller (approved):

PR #197, opened 2020-12-02
Comparison of two different ways to summarize decision tree predictions

edcthayer:
Since April, after Adam point me to Hutter et al.'s summary statistics for their random forest surrogate model implementation, I've had the nagging feeling some classical statistical methods from pooled variance and specifically the Welch-Satterthwaite equation for pooled variance degrees of freedom might be an improvement.

In this notebook, the two summarization/pooling methods are explored using simulated data from three different "true" populations.  The simulated data is assigned randomly to N subpopulations that are "standing in for" the leaves in our HomogeneousRandomForestRegressionModel (HRF).  Comparisons use the relative percentage change from the mean across all observations prior to "leaf" assignment. 

In all cases, the Welch-Satterthwaite based summarization produce less biased, more robust, and more accurate estimates of sample variance, prediction variance, and (hence) 95% confidence interval radii on both sample and prediction estimates.  The expected effect would be to reduce uncertainty in both prediction and sample estimates.

Reference links are provided in the notebook to Hutter et al.'s paper and sources for the Welch-Satterthwaite based summarization formulas.

Once/if consensus on these changes is achieved, I'll move the new summarizations into the HRF prediction class.


edcthayer:
Sorry! There are two obsolete "helper" functions that I forgot to cut from the notebook: `pooled_stats` and `single_normal_distribution`.  I'll remove these while addressing any comments in the next commit.

byte-sculptor (approved):

byte-sculptor:
As per our discussions - before we enable this improvement in Homogeneous Random Forest, we should use the Optimizer Evaluator to validate your findings across a large number of objective functions.

PR #203, opened 2020-12-08
Fixup cake cmake config

bpkroth:


build.cake
```diff
@@ -236,14 +236,17 @@ Task("Run-MSBuild-UnitTests")
 
 Task("Generate-CMake")
     .WithCriteria(() => IsRunningOnUnix())
-    .IsDependentOn("Build-NetCore")
```
bpkroth, line 239: `cmake` also handles building the dependent netcore components on its own.

build.cake
```diff
@@ -236,14 +236,17 @@ Task("Run-MSBuild-UnitTests")
...
     .Does(() =>
     {
         var cmakeSettings = new CMakeSettings
         {
             Generator = "Unix Makefiles",
             OutputPath = $"{CMakeBuildDir}",
             SourcePath = ".",
+            Options = new[]
+            {
+                $"-DCMAKE_BUILD_TYPE={CMakeConfiguration}",
```
bpkroth, line 248: Copied from the `Makefile`.
The config dependent `OutputPath` was already being used by the `Build-CMake` and `Test-CMake` targets below.

grlap (approved):

PR #206, opened 2020-12-10
Parallelizing pytest

byte-sculptor:
This PR:
1. Invokes pytest with a `-n auto` argument to make multiple tests run in parallel (num parallel jobs = num cores)
2. Installs pytest-xdist in all our pipelines to make the above possible
3. Shortens the duration of TestOptimizerEvaluator
4. Modifies the tests that use gRPC to attempt to start the service on 100 different ports before giving up. This is the easiest way to make sure that all tests requiring gRPC can run in parallel. The alternative would be to have them all talk to a single instance, but it would turn an easy parallelization problem into a hard one as we'd need to manage the lifetime of that single instance.
5. Relaxes the check in TestSmartCacheWithRemoteOptimizerV3.py as the current one was a bit too ambitious and lead to some flakiness.
6. Puts a band-aid on test_optimization_with_context. #207 hints at a long-term fix.

### DIFF
--- a/source/Mlos.Python/mlos/Optimizers/unit_tests/TestBayesianOptimizer.py
+++ b/source/Mlos.Python/mlos/Optimizers/unit_tests/TestBayesianOptimizer.py
@@ -345,7 +345,7 @@ class TestBayesianOptimizer:
         )
 
         random_state = random.Random()
-        num_restarts = 10
+        num_restarts = 5
         for restart_num in range(num_restarts):
             # Let's set up random seeds so that we can easily repeat failed experiments
             #
--- a/source/Mlos.Python/mlos/Optimizers/unit_tests/TestBayesianOptimizer.py
+++ b/source/Mlos.Python/mlos/Optimizers/unit_tests/TestBayesianOptimizer.py
@@ -70,7 +70,9 @@ class TestBayesianOptimizer:
 
     @classmethod
     def teardown_class(cls) -> None:
-        cls.server.stop(grace=None)
+        cls.server.stop(grace=None).wait(timeout=1)
+        cls.server.wait_for_termination(timeout=1)
+        cls.optimizer_service_channel.close()
 
         cls.temp_dir = os.path.join(os.getcwd(), "temp")
         if not os.path.exists(cls.temp_dir):
@@ -499,8 +501,8 @@ class TestBayesianOptimizer:
         with pytest.raises(ValueError):
             optimizer.register(input_df, only_invalid_outputs_df)
 
-
-    def test_optimization_with_context(self):
+    @pytest.mark.parametrize('run', [i for i in range(100)])
+    def test_optimization_with_context(self, run):
         # Gaussian blob in x with position dependent on context variable y.
         def f(parameters, context):
             if isinstance(parameters, pd.DataFrame):
@@ -523,7 +525,7 @@ class TestBayesianOptimizer:
         )
 
         # create some data points to eval
-        n_samples = 100
+        n_samples = 5000
         parameter_df = input_space.random_dataframe(n_samples)
         context_df = context_space.random_dataframe(n_samples)
 
--- a/source/Mlos.Python/mlos/Optimizers/unit_tests/TestBayesianOptimizer.py
+++ b/source/Mlos.Python/mlos/Optimizers/unit_tests/TestBayesianOptimizer.py
@@ -47,7 +47,8 @@ class TestBayesianOptimizer:
         global_values.tracer = Tracer(actor_id=cls.__name__, thread_id=0)
         cls.logger = create_logger(logger_name=cls.__name__)
 
-        # Start up the gRPC service. Try a bunch of ports, before giving up.
+        # Start up the gRPC service. Try a bunch of ports, before giving up so we can do several in parallel.
+
         #
         max_num_tries = 100
         num_tries = 0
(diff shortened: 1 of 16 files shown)


source/Mlos.Python/mlos/unit_tests/TestSmartCacheWithRemoteOptimizerV3.py
```diff
@@ -36,19 +36,29 @@ def setup_method(self, method):
...
+        # Start up the gRPC service. Try a bunch of times before giving up.
         #
-        self.server = OptimizerMicroserviceServer(port=50051, num_threads=10)
-        self.server.start()
-
-        self.optimizer_service_grpc_channel = grpc.insecure_channel('localhost:50051')
-        self.bayesian_optimizer_factory = BayesianOptimizerFactory(grpc_channel=self.optimizer_service_grpc_channel, logger=self.logger)
+        max_num_tries = 100
+        num_tries = 0
+        for port in range(50051, 50051 + max_num_tries):
```
bpkroth, line 43: Hmm, questions ...
Is this only necessary when running tests in parallel?
Or is it not cleaning up after itself?  Or maybe some delay after the teardown is necessary?
Can a single remote optimizer not handle requests from multiple clients?

bpkroth: In fd8ef0884dc6aa9c0a03144f1027520683557431 I think we had to add some additional code to the `teardown_method`:
https://github.com/microsoft/MLOS/pull/154/commits/e29697b82de38418ab7ab7b8d1da456037e50c9e

byte-sculptor: It's so that multiple services can run in parallel. Having one handle multiple requests would work too.

byte-sculptor: *one service - each test instantiates its own optimizer.

bpkroth: Thanks, makes sense.
The teardown changes (specifically closing the client channel explicitly) may still be necessary in the other one.  I think I only made the change in the one that was throwing errors at the time.

byte-sculptor: Hmmm I thought that the client could just crash... what happens if we don't close the connection on the client?

source/Mlos.Python/mlos/Optimizers/unit_tests/TestBayesianOptimizer.py
```diff
@@ -47,12 +47,22 @@ def setup_class(cls):
         global_values.tracer = Tracer(actor_id=cls.__name__, thread_id=0)
         cls.logger = create_logger(logger_name=cls.__name__)
 
-        # Start up the gRPC service.
+        # Start up the gRPC service. Try a bunch of ports, before giving up.
```
bpkroth, line 50:
```suggestion
        # Start up the gRPC service. Try a bunch of ports, before giving up so we can do several in parallel.

```

source/Mlos.Python/mlos/Optimizers/unit_tests/TestBayesianOptimizer.py
```diff
@@ -47,12 +47,22 @@ def setup_class(cls):
...
+        num_tries = 0
+        for port in range(50051, 50051 + max_num_tries):
+            num_tries += 1
+            try:
+                cls.server = OptimizerMicroserviceServer(port=port, num_threads=10)
+                cls.server.start()
+                cls.optimizer_service_channel = grpc.insecure_channel(f'localhost:{port}')
+                break
+            except:
+                cls.logger.info(f"Failded to create OptimizerMicroserviceServer on port {port}.")
```
bpkroth, line 62:
```suggestion
                cls.logger.info(f"Failed to create OptimizerMicroserviceServer on port {port}.")

```

source/Mlos.Python/mlos/unit_tests/TestSmartCacheWithRemoteOptimizerV3.py
bpkroth: If I recall correctly it leaves a socket open that causes conflicts later on when it goes to reuse it.

bpkroth (approved):
Couple minor suggestions.

source/Mlos.Python/mlos/unit_tests/TestSmartCacheWithRemoteOptimizerV3.py
bpkroth: If I recall correctly it leaves a socket open that causes conflicts later on when it goes to reuse it.

byte-sculptor: I thought it would only happen on the server side. Is there any repro that would tell us when we've fixed the bug?

bpkroth: https://github.com/microsoft/MLOS/compare/main...bpkroth:demonstrate-missing-grpc-client-close-bug
https://github.com/bpkroth/MLOS/commit/cadc29c8b50cf110b3325c36431cc92e2d2c1189

amueller (approved):
didn't look into the issue @bpkroth pointed out in detail but otherwise looks good.

source/Mlos.Python/mlos/unit_tests/TestSmartCacheWithRemoteOptimizerV3.py
byte-sculptor: awesome, thanks

bpkroth:
> 
> 
> didn't look into the issue @bpkroth pointed out in detail but otherwise looks good.

@amueller it was just another place to apply the grpc client close issue you found and I helped debug when converting to `pytest` (exactly references are in the thread).

PR #208, opened 2020-12-14
Implementing a multi-objective regression model

byte-sculptor:
This PR introduces a multi-objective random forest model, which is essentially a collection of single-objective random forests - each predicting a different objective. 

This model is not yet exposed as part of the optimizer.

### DIFF
--- a/source/Mlos.Python/mlos/Optimizers/RegressionModels/MultiObjectiveRegressionModel.py
+++ b/source/Mlos.Python/mlos/Optimizers/RegressionModels/MultiObjectiveRegressionModel.py
@@ -1,0 +1,41 @@
+#
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+#
+from abc import ABC, abstractmethod
+
+import pandas as pd
+
+from mlos.Spaces import Hypergrid, Point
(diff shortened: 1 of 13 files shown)


source/Mlos.Python/mlos/Optimizers/RegressionModels/MultiObjectiveRegressionModel.py
```diff
@@ -0,0 +1,45 @@
...
+import pandas as pd
+
+from mlos.Spaces import Hypergrid, Point
+from mlos.Optimizers.RegressionModels.GoodnessOfFitMetrics import DataSetType
+from mlos.Optimizers.RegressionModels.MultiObjectiveGoodnessOfFitMetrics import MultiObjectiveGoodnessOfFitMetrics
+from mlos.Optimizers.RegressionModels.MultiObjectivePrediction import MultiObjectivePrediction
+
+
+class MultiObjectiveRegressionModel(ABC):
+    """An interface for all multi-objective regression models to implement."""
```
byte-sculptor, line 16: A base class for ...

grlap (approved):

amueller:
FYI the scikit-learn decision trees and random forests support multiple targets natively, which allows sharing some structure of the tree, instead of modeling them independently.

PR #211, opened 2021-01-05
Workaround an issue uncovered by recent versions of sphinx

### LINKED ISSUE #210: Sphinx Python API docs generation broken in recent nightly CI runs
For example: <https://github.com/microsoft/MLOS/runs/1635132574?check_suite_focus=true>

bpkroth:
With sphinx 3.4.0-3.4.2 (released in late December 2020), sphinx generates
asserts by the Point module due to invalid keys (e.g. `__sphinx_mock__`) in the Point.

```txt
Exception occurred:
  File "/src/MLOS/source/Mlos.Python/mlos/Spaces/Point.py", line 75, in __getitem__
    raise KeyError(f"This Point does not have a value along dimension: {dimension_name}")
KeyError: 'This Point does not have a value along dimension: __sphinx_mock__'
```

Fixes #210 (at least enough that we can submit PRs and run the nightly CI jobs again)

### DIFF
--- a/source/Mlos.Python/mlos/Spaces/Point.py
+++ b/source/Mlos.Python/mlos/Spaces/Point.py
@@ -58,7 +58,7 @@ class Point:
         try:
             return self[dimension_name]
         except KeyError:
-            return None
+            raise AttributeError(f"This Point does not have a {dimension_name} attribute.")
 
     def __setattr__(self, name, value):
         if name == "dimension_value_dict":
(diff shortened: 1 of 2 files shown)


bpkroth:
This initially identified a workaround (avoid recent versions of sphinx).

However, I think I've identified the actual issue:
- Recent versions of sphinx make use of the `hasattr()` function to check for the `__sphinx_mock__` attribute on objects.
- The MLOS `Point` class overloads `__getattr__` and uses it to call `__getitem__` (indirectly via a `self[__sphinx_mock__]` indexing operation).
- This triggers the `KeyError` in `__getitem__` for invalid dimension keys.

Now, what to do about it ...

We could just catch the `KeyError` in the `Point` `__getattr__` overload and return `None` instead.
@byte-sculptor or @amueller
Thoughts?

byte-sculptor (approved):

amueller:
lgtm

PR #212, opened 2021-01-13
Integrating multi objective random forest into the optimizer framework

byte-sculptor:
This is the next batch of code getting us towards multi-objective optimization.

PAST: Previously we have been able to register multiple objectives with the optimizer, but the optimizer wouldn't use them at all.
PRESENT: With this PR the optimizer will instantiate a regression model for each of the objectives, but it doesn't yet use the predictions for anything. 
FUTURE: The next steps will be to write a few multi-objective utility functions so that the optimizer can use them to suggest new configurations.

The chief goal of this PR is to integrate the MultiObjectiveHomogeneousRandomForest into the framework. 

Note that there is no change to the Optimizer's interface. The prediction of the first objective is returned, and all the others are thrown out.

The reason for this is that changing the Optimizer Interface requires changing all the consumers:
* gRPC
* Tomograph
* Notebooks

We should do it in a separate PR.


grlap (approved):

source/Mlos.Python/mlos/Spaces/Point.py
```diff
@@ -52,13 +52,14 @@ def __iter__(self):
...
-            return False
         try:
             return self[dimension_name]
         except KeyError:
-            raise AttributeError(f"This Point does not have a {dimension_name} attribute.")
+            pass
+
+        # Raise the exception outside the 'except' block to not make it look like a nested error.
+        #
+        raise AttributeError(f"This Point does not have a {dimension_name} attribute.")
```
bpkroth, line 62: Good catch.

PR #222, opened 2021-02-25
Making glow worm optimizer use the context

byte-sculptor:
I was working on Random Near Incumbent Optimizer and I noticed that GlowWormSwarmOptimizer was doing something really wrong with the context features. I don't really understand how/if it ever worked with context.

So this is a quick fix, where we have the worms keep track of the parameter values (not feature values) and only constructing the features dataframe when calling the utility function. This way, the glow worm swarm optimizer can suggest optimal parameters for a given context, instead of suggesting optimal parameters and context. I really don't know how it ever worked. I suspect it was never able to compute the utility value, and always bailed on the first 'if' producing a de-facto random suggestion.


grlap (approved):

byte-sculptor:
I just debugged this code path before and after my change in this branch. Ever since context was added, GlowWarmSwarmOptimizer essentially a really bad random search optimizer. I will add more tests with Random Near Incumbent to make sure we catch such regressions in the future.

Issue #221, opened 2021-02-22
Input parameter space random state is not retained


kkanellis:
When defining an input space it is possible to pass a `random_state` argument. For example:

```python
# Define random state
RANDOM_SEED = 42
random_state = random.Random()
random_state.seed(RANDOM_SEED)

# Define spaces
input_dim_1 = DiscreteDimension('dim1', 1, 1000)
input_dim_2 = DiscreteDimension('dim2', 1, 1000)
input_space = SimpleHypergrid(
  name='input', dimensions=[input_dim_1, input_dim_2], random_state=random_state)
```

You can then use the `input_space` hypergrid to initialize a simple optimization problem, and eventually get (initially random) suggestions from the underlying Bayesian Optimizer, as follows:

```python

output_dim = ContinuousDimension(name="throughput", min=0, max=10**8)
output_space = SimpleHypergrid(name="objective", dimensions=[output_dim])

# Define optimization problem
optimization_problem = OptimizationProblem(
    parameter_space=input_space,
    objective_space=output_space,
    objectives=[Objective(name="throughput", minimize=False)]
)

# Initialize optimizer
optimizer_config = bayesian_optimizer_config_store.default
optimizer_factory = BayesianOptimizerFactory()
optimizer = optimizer_factory.create_local_optimizer(
    optimization_problem=optimization_problem,
    optimizer_config=optimizer_config
)

suggestion = optimizer.sample()
```

From my understanding, each time that the above code is invoked, the sample point should be the *same*, since we have fixed the input parameter space random state. However, this is not the case -- each time a different random point is being generated. It seems that for some reason the random state is not being retained.

One possible solution that I've found is to explicitly replace the parameter space random, *after* the optimizer has been constructed, which seems to solve this problem:

```python
# Replace the random state of all input parameter space (sub)dimensions
optimizer.experiment_designer   \
        .optimization_problem   \
        .parameter_space        \
        .random_state = random_state
```

Moreover, I've found that in `ExperimentDesigner` class, a random number generator is being created, which does not adhere to global random seeds: 

https://github.com/microsoft/MLOS/blob/dd56ab4125542cad0e68d9754ca5a6021953e63e/source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/ExperimentDesigner.py#L118

[...]

PR #226, opened 2021-03-04
Random Near Incumbent Optimizer

byte-sculptor:
This PR introduces a third utility function optimizer: RandomNearIncumbent. Its implementation is described in detail in RandomNearIncumbentOptimizer.py.

The interface between the UtilityFunctionOptimizer subclasses and the Experiment designer has been changed. Now, if a given UtilityFunctionOptimizer is unable to produce a guided suggestion, it raises an Exception and the ExperimentDesigner can decide how to handle it and perhaps explicitly generate a random suggestion. All existing utility function optimizers have been updated and TestUtilityFunctionOptimizers.py has been extended to cover this new case.

Testing: I ran pyest with code coverage and got 100% coverage for both RandomSearchOptimizer and GlwoWormSwarmOptimizer. For RandomNearIncumbentOptimizer the score is 93%, but the only lines missing are really defensive checks, that raise an exception if something is wrong at the end of the suggest() method. 

I also temporarily, made the RandomNearIncumbent the default optimizer and ran all unit tests with it locally. It is considerably slower than RandomSearchOptimizer so I reverted the to the old default having verified that it works.


source/Mlos.Python/mlos/Exceptions.py
```diff
@@ -18,3 +18,22 @@ class InvalidPointException(MlosException):
...
+class UnableToProduceGuidedSuggestionException(MlosException):
+    """Thrown by any utility function optimizer unable to produce a guided suggestion.
+
+    This is to explicitly instruct the experiment designer that a random suggestion should be produced. This can be reflected in
+    the suggestion's metadata. Furthermore, having this information we can now assert that a portion of suggestions is non-random.
+    """
+
+
+class UtilityValueUnavailableException(UnableToProduceGuidedSuggestionException):
+    """Thrown by utility function optimizer unable to produce a guided suggestion due to utility function productin no values.
```
byte-sculptor, line 32: producing

source/Mlos.Python/mlos/Logger.py
```diff
@@ -65,10 +65,11 @@ def create_logger(logger_name, create_console_handler=True, create_file_handler=
     formatter.datefmt = '%m/%d/%Y %H:%M:%S'
 
     if create_console_handler:
```
byte-sculptor, line 67: This is to avoid duplicated log messages in the console

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctionOptimizers/RandomNearIncumbentOptimizer.py
```diff
@@ -0,0 +1,451 @@
...
+        # We will cache good configs from past invocations here.
+        #
+        self._good_configs_from_the_past_invocations_df = None
+
+    @trace()
+    def suggest(self, context_values_dataframe: pd.DataFrame = None):
+        """ Returns the next best configuration to try.
+
+        The idea is pretty simple:
+            1. We start with all configs on the pareto frontier, plus some good points from previous calls to suggest plus some random configs.
```
byte-sculptor, line 157: some* configs on the pareto frontier

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctionOptimizers/RandomNearIncumbentOptimizer.py
```diff
@@ -0,0 +1,451 @@
...
+            2. For each point we generate random neighbors and optionally adjust them using our velocity.
+            3. We compute utility for all neighbors and select a new incumbent.
+            4. We update the velocity.
+            5. We repeat until we ran out of iterations or until velocity falls below some threshold.
```
byte-sculptor, line 161: run

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctionOptimizers/RandomNearIncumbentOptimizer.py
```diff
@@ -0,0 +1,451 @@
...
+            all_neighbors_df = all_neighbors_df.loc[neighbors_utility_df.index]
+            all_neighbors_df['utility'] = neighbors_utility_df['utility']
+            all_neighbors_df['utility_gain'] = all_neighbors_df['utility'] - all_neighbors_df['incumbent_utility']
+
+            # We can filter out all rows with negative utility gain.
+            #
+            all_neighbors_df = all_neighbors_df[all_neighbors_df['utility_gain'] >= 0]
+            self.logger.info(f"{len(all_neighbors_df.index)} have positive utility gain.")
+
+            # The series below has best neighbor's index as value and the incumbent_config_idx as key.
```
byte-sculptor, line 277: and the incumbent_config_idx as index

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctionOptimizers/RandomNearIncumbentOptimizer.py
```diff
@@ -0,0 +1,451 @@
...
+            self.logger.info(error_message)
+            raise UtilityValueUnavailableException(error_message)
+
+        if num_iterations == 0:
+            error_message = f"{self.__class__.__name__} performed 0 iterations."
+            self.logger.info(error_message)
+            raise UnableToProduceGuidedSuggestionException(error_message)
+
+        if incumbents_df.dtypes['utility'] != np.float:
+            self.logger.info(f"The type of incumbents_df['utility'] is {incumbents_df.dtypes['utility']}. Utility function: {self.utility_function.__class__.__name__}, incumbents_df lendth: {len(incumbents_df.index)}")
```
byte-sculptor, line 335: length

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctionOptimizers/RandomNearIncumbentOptimizer.py
```diff
@@ -0,0 +1,451 @@
...
+
+        initial_points_pareto_fraction = self.optimizer_config.initial_points_pareto_weight / total_weights
+        initial_points_cached_good_fraction = self.optimizer_config.initial_points_cached_good_params_weight / total_weights
+
+        num_initial_points = self.optimizer_config.num_starting_configs
+
+        # Let's start with the pareto points.
+        #
+        num_desired_pareto_points = math.floor(num_initial_points * initial_points_pareto_fraction)
+        num_existing_pareto_points = len(self.pareto_frontier.params_for_pareto_df.index) if self.pareto_frontier.params_for_pareto_df is not None else 0
```
byte-sculptor, line 379: This incurs two copies of params_for_pareto_df. Rewrite to copy only once.

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctionOptimizers/unit_tests/MultiObjectivePassThroughModelForTesting.py
```diff
@@ -0,0 +1,150 @@
...
+        objective_function_config=objective_function_config_store.get_config_by_name("multi_objective_waves_3_params_2_objectives_half_pi_phase_difference")
+    )
+)
+
+
+class MultiObjectivePassThroughModelForTesting(MultiObjectiveRegressionModel):
+    """Used for testing. Exposes the MultiObjectiveRegressionModel interface, but is in fact not a model.
+
+        This dummy wraps around any of the synthetic functions and when it's time to make a prediction it simply evaluates that function,
+    adds some preconfigured noise of its own and returns. For now register() is a noop, but we could conceivably make the observations
```
byte-sculptor, line 80: noise -> uncertainty

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctionOptimizers/RandomNearIncumbentOptimizer.py
```diff
@@ -0,0 +1,472 @@
...
+
+    **Starting Points**
+
+        Utility function optimizers are invoked repeatedly to find maxima of utility functions that change only slightly between
+    subsequent iterations. So we should be able to utilize learnings from prior iterations to speed up search for utility function
+    maxima in subsequent calls. We do this by:
+        * using points on the pareto frontier as some of the starting points,
+        * using good points from past iterations as some of the starting points.
+
+        Additionally, we use some random points as starting points too to balance the exlpore-exploit tradeoff.
```
bpkroth, line 112:
```suggestion
        Additionally, we use some random points as starting points too to balance the explore-exploit tradeoff.

```

bpkroth:
Regarding "considerably slower", that's just the `suggest()` call though, correct?  Do we have data on whether it converges to a good/better optima faster?
Future work could be to add perf regression tests for that.  Probably deserves fixing #221 first though so that the test is reproducible.

bpkroth:
Another thought: how does this compare with BlendSearch techniques from https://github.com/microsoft/flaml

byte-sculptor:
> Regarding "considerably slower", that's just the `suggest()` call though, correct? Do we have data on whether it converges to a good/better optima faster?
> Future work could be to add perf regression tests for that. Probably deserves fixing #221 first though so that the test is reproducible.

Right, basically every iteration in RandomNearIncumbent is comparable in duration to the entire RandomSearchOptimizer run. Anecdotally, it does a lot better than RandomNearIncumbent (from a few runs I have done locally), but I have not yet benchmarked it properly yet.

byte-sculptor:
> Another thought: how does this compare with BlendSearch techniques from https://github.com/microsoft/flaml

Not sure yet. but we could wrap it and see :). A quick look at it makes me think that their current implementation is not terribly parallelizable, but we could change that. Also, I'm not sure how well we meet their assumptions. There is in general no difference in cost of evaluating one config over another for us - they are all just predictions made by the model and - as it stands now - it's all the same cost more or less.

grlap (approved):

PR #233, opened 2021-03-29
Bug fix in GridPlot

byte-sculptor:
This is to allow grid plot to display an dataset where the we optimize only a subset of objectives (i.e. one or more of the objectives is observed but not actively optimized).


grlap (approved):

sergiy-k (approved):

byte-sculptor:
Already checked in with a different PR

PR #235, opened 2021-04-05
Added test of stacking PolyFeat and OHE to assess RERF stacking logic

edcthayer:


edcthayer:
Didn't mean to create this - closing it.

PR #240, opened 2021-04-08
Moving Optimizer Store out of Optimizer Microservice

byte-sculptor:
The goal of the upcoming PRs is to split gRPC into two Microservices:
1. OptimizerMicroservice - implementing only very few interfaces so that we can have optimizer clients in multiple languages.
2. OptimizerMonitoringMicroservice - implementing all of the other APIs currently defined in the .proto file.

gRPC forces us to make two separate Service classes but their objects will need to be able to reference the same optimizers. So step 1 is to move the OptimizerStore to its own class so that it could be referenced by both services. Down the road, we should evolve the OptimizerStore to use a Database or filesystem or MLFlow. For now it stays in memory.

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
+    <!-- Only use public nuget sources for OSS builds. -->
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+  </packageSources>
+  <disabledPackageSources>
+    <clear />
+  </disabledPackageSources>
+</configuration>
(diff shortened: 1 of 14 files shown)


bpkroth:
> The goal of the upcoming PRs is to split gRPC into two Microservices:
> 
> 1. OptimizerMicroservice - implementing only very few interfaces so that we can have optimizer clients in multiple languages.
> 2. OptimizerMonitoringMicroservice - implementing all of the other APIs currently defined in the .proto file.

Regarding the "we can have optimizer clients in multiple languages" comment.  Just to be clear on my understanding, the split is to reduce the implementation burden of adding additional languages to only those APIs that likely involve actual optimization, and not the presumably less used monitoring of APIs, correct?

bpkroth (approved):
LGTM

byte-sculptor:
> > The goal of the upcoming PRs is to split gRPC into two Microservices:
> > 
> > 1. OptimizerMicroservice - implementing only very few interfaces so that we can have optimizer clients in multiple languages.
> > 2. OptimizerMonitoringMicroservice - implementing all of the other APIs currently defined in the .proto file.
> 
> Regarding the "we can have optimizer clients in multiple languages" comment. Just to be clear on my understanding, the split is to reduce the implementation burden of adding additional languages to only those APIs that likely involve actual optimization, and not the presumably less used monitoring of APIs, correct?

Well put. 

The clients only need to be able to:
1. Create an optimizer (or connect to an existing one)
2. Solicit suggestions
3. Register observations

PR #243, opened 2021-04-12
Splitting gRPC into two services

byte-sculptor:
The goal of this PR is to split the current OptimizerService into two separate services:
1. OptimizerService
2. OptimizerMonitoringService

The OptimizerService can be used by clients in any language to instantiate an optimizer, and use it to drive optimization. It's surface area has been dramatically reduced to:
1. Reduce implementation burden for all those clients
2. Reduce the complexity of version management - the OptimizerService API should be much more stable than the OptimizerMonitoringService API, and now they can evolve separately.

The OptimizerMonitoringService is used to query and visualize the optimization data, as well as models predictions. It is meant to be consumed primarily by a Python client and should not be implemented by clients in other languages. 

To achieve the above goals:
1. I split the OptimizerService.proto into 3 files: MlosCommonMessageTypes.proto, OptimizerService.proto, OptimizerMonitoringService.proto
2. Generated the required _pb2.py and _pb2_grpc.py files (and modified the scripts).
3. Split the OptimizerMicroservice class into: OptimizerService, OptimizerMonitoringService
4. Registered both Services' servicers with the server (on the same grpc channel).

Additionally, I fixed an np.float deprecation issue (we should be using float, instead of np.float).

This is a refactoring exercise, thus no new tests were needed. I added no new functionality.


byte-sculptor:
This needs more work.

PR #244, opened 2021-04-13
2021/april/separating grpc services

byte-sculptor:
The goal of this PR is to split the gRPC service into two services that can evolve separately.

The OptimizerService now has a reduced surface and in subsequent PRs we will evolve it further. The goal is to make it easy to implement clients in different languages, and to reduce the cost of maintaining a public-facing API. 

The OptimizerMonitoringService contains many of the interfaces likely to evolve faster. In general we shouldn't need to worry about backwards compatibility of this service as it is meant to be shipped with its client (the Tomograph, etc.).

Note, the goal of this PR is to simply separate the two services - no new functionality was added, nor was any removed. Thus existing tests should cover these changes.

Implementation

Splitting the services required creating a second .proto file and specifying the OptimizerMonitoringService there. There is some code duplication, which is the price we pay for decoupling of the two services. This duplication should be reduced as the two services evolve their separate ways.

In order to properly serialize/deserialize the OptimizationProblem, I wrote the OptimizerServiceEncoder/Decoder classes and OptimizerMonitoringServiceEncoder/Decoder classes. These will be further developed to encode/decode other objects too, as we move away from the JSON strings inside protobuf messages.


source/Mlos.Python/mlos/Optimizers/RegressionModels/Prediction.py
```diff
@@ -113,7 +113,14 @@ def validate_dataframe(self, dataframe: pd.DataFrame):
...
         if mean_variance_col in self.expected_column_names:
             if dataframe[mean_variance_col].notnull().any():
-                assert (dataframe[dataframe[mean_variance_col].notnull()][mean_variance_col] >= 0).all()
+                if not (dataframe[dataframe[mean_variance_col].notnull()][mean_variance_col] >= 0).all():
+                    violated_rows_df = dataframe[dataframe[dataframe[mean_variance_col].notnull()][mean_variance_col] < 0]
+                    with pd.option_context('display.max_rows', None, 'display.max_columns', None):
+                        print(violated_rows_df)
+                    print(f"Num invalid rows: {len(violated_rows_df.index)}")
+                    print(f"Index: {violated_rows_df.index}")
+                    print(f"{mean_variance_col}: {violated_rows_df[mean_variance_col]}")
```
byte-sculptor, line 122: This assert has failed several times last week, but had insufficient information to diagnose the problem, so adding a little extra information to help debug this in the future.

source/Mlos.Proto/OptimizerMonitoringService.proto
```diff
@@ -0,0 +1,199 @@
+syntax = "proto3";
+package mlos.optimizer_monitoring_service;
+
+option csharp_namespace = "Mlos.OptimizerMonitoringService";
```
grlap, line 4: Not required

grlap (approved):

PR #247, opened 2021-04-20
2021/april/reducing test duration

byte-sculptor:
This PR reduces the number of iterations in some really expensive tests. It does slightly reduce our test coverage, but mainly in the evaluation tools. Given that our Linux pipeline has been timing out about 50% of the time, I think the increase in productivity is worth the reduction in coverage.


source/Mlos.Python/mlos/OptimizerEvaluationTools/unit_tests/TestEnvelopedWaves.py
```diff
@@ -39,15 +39,15 @@ def test_enveloped_waves(self):
...
             assert ((objectives_df['y'] <= (num_params + vertical_shift)) & (objectives_df['y'] >= -num_params + vertical_shift)).all()
 
-    def test_random_configs(self):
-        for _ in range(100):
-            function_config = enveloped_waves_config_space.random()
-            objective_function = EnvelopedWaves(function_config)
-            random_params_df = objective_function.parameter_space.random_dataframe(100)
-            objectives_df = objective_function.evaluate_dataframe(random_params_df)
-            assert objective_function.output_space.get_valid_rows_index(objectives_df).equals(objectives_df.index)
+    @pytest.mark.parametrize('i', [i for i in range(10)])
```
byte-sculptor, line 42: This took a long time, but it's just a synthetic function test that has not failed since November. Reducing num iterations.

edcthayer (approved):

PR #249, opened 2021-06-07
Created rustc flag optimizer example

zaporter:
This adds the rustc flag optimizer example to the source/Mlos.Python/mlos/Examples directory. I had thought of putting it in the notebooks directory or the source/Examples directory but it didn't seem to fit in either. If you have further thoughts, let me know.


bpkroth:
I think I would prefer to see them in the `source/Mlos.Notebooks/` tree since that's where we point a lot of our other examples to.

source/Mlos.Python/mlos/Examples/RustcFlagOptimization/bevybench/Cargo.lock
```diff
@@ -0,0 +1,2602 @@
+# This file is automatically @generated by Cargo.
```
bpkroth, line 1: Should probably remove this file and add a `.gitignore` rule to prevent future inclusions as well.

source/Mlos.Python/mlos/Examples/RustcFlagOptimization/bevybench/target/.rustc_info.json
```diff
@@ -0,0 +1 @@
+{"rustc_fingerprint":2256235523725413435,"outputs":{"1164083562126845933":["rustc 1.47.0 (18bf6b4f0 2020-10-07)\nbinary: rustc\ncommit-hash: 18bf6b4f01a6feaf7259ba7cdae58031af1b7b39\ncommit-date: 2020-10-07\nhost: x86_64-unknown-linux-gnu\nrelease: 1.47.0\nLLVM version: 11.0\n",""],"4476964694761187371":["___\nlib___.rlib\nlib___.so\nlib___.so\nlib___.a\nlib___.so\n/home/zack/.rustup/toolchains/stable-x86_64-unknown-linux-gnu\ndebug_assertions\nproc_macro\ntarget_arch=\"x86_64\"\ntarget_endian=\"little\"\ntarget_env=\"gnu\"\ntarget_family=\"unix\"\ntarget_feature=\"fxsr\"\ntarget_feature=\"sse\"\ntarget_feature=\"sse2\"\ntarget_os=\"linux\"\ntarget_pointer_width=\"64\"\ntarget_vendor=\"unknown\"\nunix\n",""]},"successes":{}}
```
bpkroth, line 1: Same here.  Remove and add a `.gitignore`.

source/Mlos.Python/mlos/Examples/RustcFlagOptimization/raw_data.csv
```diff
@@ -0,0 +1,447 @@
+,codegen-units,embed-bitcode,opt-level,force-frame-pointers,force-unwind-tables,linker-flavor,lto,no-vectorize-loops,no-vectorize-slp,overflow-checks,target-cpu,relocation-model,panic,compile-time-s,run-time-s,binary-size-mb,is_pareto,class
```
bpkroth, line 1: I didn't see this file referenced on my first skim.  Is it relevant to include or can it be reproduced by the notebook or some other scripts that weren't included?

bpkroth (changes requested):
Hi, thanks for your PR!  Love to see this applied to rust.

I left a few other comments in places about files that don't need to be included.  I think the html rendering can also be included in that set of files to remove.  The github rendering of the notebook (https://github.com/microsoft/MLOS/blob/b21ba167bc8500a95fa7073357b0e642d39baf42/source/Mlos.Python/mlos/Examples/RustcFlagOptimization/RustcFlagOptimization.ipynb) doesn't seem to include the bokeh graphs, but it also doesn't render the html, so I'm not sure it adds lots of value for us to include it (one more thing to maintain).
One other thing you could do is add a short README.md describing the purpose/contents of the RustcFlagOptimization folder.
Thanks!

byte-sculptor:
> Hi, thanks for your PR! Love to see this applied to rust.
> 
> I left a few other comments in places about files that don't need to be included. I think the html rendering can also be included in that set of files to remove. The github rendering of the notebook (https://github.com/microsoft/MLOS/blob/b21ba167bc8500a95fa7073357b0e642d39baf42/source/Mlos.Python/mlos/Examples/RustcFlagOptimization/RustcFlagOptimization.ipynb) doesn't seem to include the bokeh graphs, but it also doesn't render the html, so I'm not sure it adds lots of value for us to include it (one more thing to maintain).
> One other thing you could do is add a short README.md describing the purpose/contents of the RustcFlagOptimization folder.
> Thanks!

Brian, I'd recommend that we keep the html file, as it forms the snapshot and the report. Basically, the notebook will eventually stop working, but the html report will live there forever and will be a useful, interactive example on how to use the optimizer. Same goes for the raw data. This entire package basically forms an example of how one can use the MLOS tools to setup an optimization study, execute it, and analyze the results.

bpkroth:
> > Hi, thanks for your PR! Love to see this applied to rust.
> > I left a few other comments in places about files that don't need to be included. I think the html rendering can also be included in that set of files to remove. The github rendering of the notebook (https://github.com/microsoft/MLOS/blob/b21ba167bc8500a95fa7073357b0e642d39baf42/source/Mlos.Python/mlos/Examples/RustcFlagOptimization/RustcFlagOptimization.ipynb) doesn't seem to include the bokeh graphs, but it also doesn't render the html, so I'm not sure it adds lots of value for us to include it (one more thing to maintain).
> > One other thing you could do is add a short README.md describing the purpose/contents of the RustcFlagOptimization folder.
> > Thanks!
> 
> Brian, I'd recommend that we keep the html file, as it forms the snapshot and the report. Basically, the notebook will eventually stop working, but the html report will live there forever and will be a useful, interactive example on how to use the optimizer. Same goes for the raw data. This entire package basically forms an example of how one can use the MLOS tools to setup an optimization study, execute it, and analyze the results.

Works for me.  We just don't generally do that for the others currently.  Maybe include a timestamp in the file/data or README.md to note when/how it was generated.

zaporter:
Alright, let me know if this works. I think I fixed all of the issues you highlighted.

bpkroth (commented):
LGTM, I only made a small change to the README.md for markdown syntax lint purposes.

bpkroth:
@byte-sculptor thoughts on including this notebook in the docs rendering at https://microsoft.github.io/MLOS/notebooks?

It might take a little tweaking since we already have html for that and they don't need to be rendered, so it can probably be a separate PR as well if we want.

bpkroth (approved):

byte-sculptor:
> > > Hi, thanks for your PR! Love to see this applied to rust.
> > > I left a few other comments in places about files that don't need to be included. I think the html rendering can also be included in that set of files to remove. The github rendering of the notebook (https://github.com/microsoft/MLOS/blob/b21ba167bc8500a95fa7073357b0e642d39baf42/source/Mlos.Python/mlos/Examples/RustcFlagOptimization/RustcFlagOptimization.ipynb) doesn't seem to include the bokeh graphs, but it also doesn't render the html, so I'm not sure it adds lots of value for us to include it (one more thing to maintain).
> > > One other thing you could do is add a short README.md describing the purpose/contents of the RustcFlagOptimization folder.
> > > Thanks!
> > 
> > 
> > Brian, I'd recommend that we keep the html file, as it forms the snapshot and the report. Basically, the notebook will eventually stop working, but the html report will live there forever and will be a useful, interactive example on how to use the optimizer. Same goes for the raw data. This entire package basically forms an example of how one can use the MLOS tools to setup an optimization study, execute it, and analyze the results.
> 
> Works for me. We just don't generally do that for the others currently. Maybe include a timestamp in the file/data or README.md to note when/how it was generated.

Agreed!

zaporter:
I misunderstood Cargo.lock. Purpose is to ensure that when other people use the same repo, they download the same library versions. Ensures a static build.

bpkroth:
> I misunderstood Cargo.lock. Purpose is to ensure that when other people use the same repo, they download the same library versions. Ensures a static build.

Seems better handled in the dependencies files themselves, but :shrug:

byte-sculptor (approved):

PR #251, opened 2021-06-09
Split CI workflows for easier retry on failures

### LINKED ISSUE #250: Split CI workflows
CI workflows can be flaky, even for reasons not under our control (e.g. slow runner).
Right now retrying one failed portion requires the entire pipeline to be rerun, which is a bit annoying.
It would be nice if we could split the pipelines up a bit so that we can retry only the portions (and maybe their dependency parts) that failed.

bpkroth:
Fixes #250 

Basically uses a simple shell script and string replacement to template out separate workflows for each of the Ubuntu version items and splits the Windows workflow off so that they can each be retried independently.

A downside is that if say Ubuntu 16.04 fails but 20.04 succeeds, then we will push a new docker image for one but not the other rather than an all or nothing approach.

I have an example of this working here:
https://github.com/bpkroth/MLOS/pull/18/checks

I think the workflows could probably be cleaned up a bit more (e.g. remove unnecessary matrix usage now that they're separatead), but it seems functional to move us along for now.

We'd have to change the PR checkin requirements to require jobs from each workflow though if we switch to that.


sergiy-k:
Thank you! That's a lot of work. The overall change looks good to me but did not drill into details.

sergiy-k (approved):

.github/workflows/ubuntu.yml.tmpl
```diff
@@ -0,0 +1,409 @@
+# MLOS continuous integration pipelines:
```
bpkroth, line 1: Most of the Linux portions of `main.yml` got moved into this one.

.github/workflows/ubuntu-20.04.yml.extra
```diff
@@ -0,0 +1,284 @@
+# vim: set ft=yaml:
```
bpkroth, line 1: This file has the few portions of the `main.yml` pipelines that ran on Linux but only for one version.
It gets appended to only one of the templated out files (ubuntu-20.04.yml)

.github/workflows/ubuntu-20.04.yml
```diff
@@ -444,46 +451,17 @@ jobs:
         # so we instruct git-diff to ignore those when considering changes.
         git --no-pager diff --ignore-cr-at-eol --exit-code
 
-  # FIXME: These timeouts are egregiously long right now.
```
bpkroth, line 447: The windows portions from `main.yml` got moved to `windows.yml`

.github/workflows/ubuntu-20.04.yml
```diff
@@ -332,9 +314,31 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: [prep-vars]
+    if: needs.prep-vars.outputs.UbuntuVersionMatrix != '["SKIP_ME"]'
+    steps:
+    - uses: actions/checkout@v2
+    - name: Run the gen-workflows.sh script and check for changes that are not checked in
+      shell: bash
+      run: |
+        ./.github/workflows/gen-workflows.sh
+        git --no-pager diff --exit-code
```
bpkroth, line 335: This is a new check to make sure that the workflow templating script is run and we don't find any differences.
This checks two things:
1. That not individual `ubuntu-{version}.yml` has been edited directly.
2. That the individual `ubuntu-{version}.yml` files that are checked in match what the script/templates say they should.

.github/workflows/ubuntu-20.04.yml
```diff
@@ -107,22 +81,28 @@ jobs:
...
+      # (though we retain the matrix rules for convenience for now).
+      run: |
+        if [ "$GITHUB_REF" == 'refs/heads/main' ] || [ "$GITHUB_EVENT_NAME" == 'schedule' ]; then
+          echo '::set-output name=UbuntuVersionMatrix::["20.04"]'
+        elif ! echo '20.04' | egrep -q '^18.04$'; then
+          echo '::set-output name=UbuntuVersionMatrix::["20.04"]'
         else
-          echo '::set-output name=UbuntuVersionMatrix::[16.04, 20.04]'
+          echo '::set-output name=UbuntuVersionMatrix::["SKIP_ME"]'
         fi
```
bpkroth, line 98: Previously we used dynamically generated lists of versions (based on whether we were in PR mode or CI mode) in a matrix for the later rules to use.

Now we template those out one for each version, but that means we need a non-empty value to check for when we have nothing to do during PR mode (e.g. skip Ubuntu 18.04 checks).
This also means lots of `if: ` conditional job checks on the remainder of the job templates.

.github/workflows/ubuntu-20.04.yml
```diff
@@ -24,35 +27,6 @@ on:
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
-  # Run the debug/retail msbuild jobs separately so they can be executed in parallel.
-
-  windows-msbuild:
```
bpkroth, line 29: This got moved to `windows.yml` from `main.yml`

.github/workflows/ubuntu-20.04.yml
```diff
@@ -1,20 +1,23 @@
...
+# - Build/Test on Windows (split to windows.yml)
 # - Build/Test Ubuntu 16.04, 18.04, 20.04
 #   (using docker images which also get published on success)
 # - Check for license headers
 # - Generate a static website (and link check it)
+#
+# Note: to allow separate retries, we split these into different workflows using the gen-workflows.sh script.
 
-name: CI
+name: CI-Ubuntu-20.04
```
bpkroth, line 12: This is a templated out version of this file for Ubuntu 20.04.  Git seems to match against it instead of `ubuntu.yml.tmpl` which is the source it's derived from.  🤷

bpkroth:
> Thank you! That's a lot of work. The overall change looks good to me but I did not drill into details.

Thanks.  I tried to annotate it with some comments a bit, but mostly it was some reorgs to new files and a few new conditional rules.  Let me know if you have any other questions/comments, else I'll push this through in a bit to hopefully help unblock some other PRs.

amueller (approved):

PR #255, opened 2021-07-25
Moved RERF from sklearn lasso to MLOS LassoCVRegressionModel

edcthayer:
Having previously added the LassoCVRegressionModel to the MLOS RegressionModel class, wanted to change the RERF root regression model option to the MLOS LassoCVRegressionModel.

The following changes were made so this could happen:
1) Simplified the unit tests for RERF models since the newer LassoCrossValidatedRegressionModel doesn't support the algebraic gradient feature in the original RERF implementation.
2) Moved the RERF configuration to the newer ConfigStore structure while modifying the boosting_root_model_name config to point to LassoCrossValidatedRegressionModel (instead of the sklearn lasso previously used).
3) Deleted (now) obsolete RegressionModels/SklearnLassoRegressionModelConfig.py.  Will delete the SklearnRidgeRegressionModelConfig.py once it is replaced with a RidgeCVRegressionModel.
4) Moved all changes except the .pipelines/.pipeline.user.linux.yml from https://msdata.visualstudio.com/Database%20Systems/_git/MLOS/pullrequest/619130?_a=files into this PR to fix failing tests.  The .pipelines change was specific to ADO.

One subtle change in RERF's _fit_root_regression() is to create a temporary SimpleHypergrid to serve as the input_space when calling the (embedded) root regression model (LassoCVRegressionModel in this case).  This was done to allow RERF to manage the design matrix produced by exploding the original input space to a polynomial basis functions and applying the classical one hot encoding for polynomial fits.  Since this new SimpleHypergrid no longer contains categorical dimensions, LassoCV will not project the one hot encoding one might expect in the presence of categorical input dimensions to a call to RERF .fit() or .predict().

Next steps are to:
1) Move both the LassoCrossValdatedRegressionModel and this updated RERF model to the list available to the optimizers.
2) Create a RidgeCrossValidatedRegressionModel to MLOS RegressionModel and make it available in RERF as an alternative root model.


source/Mlos.Python/mlos/Optimizers/RegressionModels/RegressionEnhancedRandomForestModel.py
```diff
@@ -376,7 +299,7 @@ def _execute_grid_search_for_random_forest_regressor_model(self, x_filtered_to_d
             'n_estimators': [10, 50, 100],
             'max_features': max_feature_param
         }
-
+        print(f'RERF::performing random forest grid search CV')
```
byte-sculptor, line 302: self.logger.info(f"Performing Random Forest Grid Search CV")

sergiy-k:
Awesome! @edcthayer, thank you for adding this!

source/Mlos.Python/mlos/Optimizers/RegressionModels/RegressionEnhancedRandomForestModel.py
edcthayer: Changed in latest commit

source/Mlos.Python/mlos/Optimizers/RegressionModels/RegressionEnhancedRandomForestModel.py
```diff
@@ -387,214 +310,133 @@ def _execute_grid_search_for_random_forest_regressor_model(self, x_filtered_to_d
...
+            objective_name=self.target_dimension_names[0],
+            predictor_outputs=self._PREDICTOR_OUTPUT_COLUMNS,
+            dataframe_index=valid_rows_index
+        )
+        prediction_dataframe = predictions.get_dataframe()
+
+        if valid_rows_index is not None and not valid_rows_index.empty:
+            prediction_dataframe[is_valid_input_col] = True
+
+            base_predictions_dataframe = self.base_regressor_.predict(model_design_matrix_dataframe).get_dataframe()
```
byte-sculptor, line 349: At any point do we validate that this design matrix, belongs to the base regressors feature space?  I mean, it seems like it should, but at least for debugging or somewhere in tests, an assert would be good. Even if only on column names being what the base regressor expects.

edcthayer: Maybe assert that the design_matrix column names are made up of polynomial terms and one hot encoding column names derived from the RERF input_space features?  I'm not sure that failing such an assert would mean anything.  For example, if Ed+1 changes the transform_x() method and destroys the feature names, but correctly captures the design, then asserting about column names may cause pause, but not anything more meaningful.  What we want to be certain of is that the base regressors don't encounter categorical dimensions and then invoke their OHE adapter because the OHE adapter output_space is combined with the polynomial adapter output_space is a very specific manner.  Constructing the design_matrix_hypergrid ensures the base regressor is initialized with the expected input_space needed to correspond to the exploded feature space.

byte-sculptor (approved):

PR #257, opened 2021-07-30
Add LassoCV and RegressionEnhancedRandomForest Regression Models to optimizers

edcthayer:
Changes needed to allow optimizers to use either LassoCrossValidated or RegressionEnhancedRandomForest regression models.


source/Mlos.Python/mlos/Optimizers/BayesianOptimizer.py
```diff
@@ -61,14 +63,20 @@ def __init__(
...
+            LassoCrossValidatedRegressionModel.__name__,
+            RegressionEnhancedRandomForestRegressionModel.__name__
         )
 
         # Note that even if the user requested a HomogeneousRandomForestRegressionModel, we still create a MultiObjectiveRegressionModel
         # with just a single RandomForest inside it. This means we have to maintain only a single interface.
         #
+        # TODO : Ask Adam why model_config assignment is .homogeneous_random_forest_regression_model_config
+        #        Expected the passed optimizer_config would specify the model_config to be used as the value for surrogate_model_implementation
+        print(f'self.optimizer_config: {self.optimizer_config}')
```
byte-sculptor, line 76: Essentially, the complete optimizer Hypergrid is composed, by successively joining subgrids for each model to the root hypergrid. These subgrids need to have unique names, which then form a prefix for the parameter.

So each regression model, needs to have a uniquely named Hypergrid, otherwise, keeping things straight would be quite difficult.

PR #258, opened 2021-08-04
Preliminary Zaporter/timeseries support

zaporter:
This PR is a mess. I will clean this up later. I messed up and continued to base this off of my previous commits. I will fix it tomorrow. I am just too tired to fight with git right now
 
This PR in "Files changed" has two sub-PRs that I will break out tomorrow:
1) Observation Deprecation: removes the observation gRPC message and switches to Observations. Doesn't change almost any code on the python side and changes ~10 lines on the C# side. Minimal change. 
Please don't comment on this or editing it. 

2) Addition of timeseries support. Please see Mlos.Notebooks/SeriesConvergenceDemo.ipynb to see a little bit of how this works. I need to clean it up and add comments. 
Current thoughts: 
- It currently just uses predicted_value to find utility. The error bounds on the predicted_value should be intelligently included into the calculations. 
- It has no unit tests. 
- It has mixed performance and has a hard time learning things like (a*x^2)sin(b*x) where x is the modulated variable. 
- It has a pretty clean interface and doesn't change the current API in any serious way. This might need to be altered a bit though
- It only supports single objective optimization. We will require multi-objective optimization but this will require a much more complicated class that I will work on as soon as I am happy with the single-objective optimization


source/Mlos.Model.Services.Client/OptimizerServiceEncoderDecoder.cs
```diff
@@ -47,13 +47,13 @@ public static OptimizerService.Objective EncodeObjective(OptimizationObjective o
                 Name = objective.Name,
             };
         }
-
-        public static OptimizerService.GuestSubgrid EncodeJoinedSubgrid(JoinedSubgrid joinedSubgrid)
+
+        public static OptimizerService.GuestSubgrid EncodeJoinedSubgrid(JoinedSubgrid subgridJoin)
```
byte-sculptor, line 51: ?

source/Mlos.Proto/GeneratePythonGrpcCode.cmd
```diff
@@ -4,6 +4,8 @@
 @echo off
 rem We need to recreate the structure of the python module in the input to grpc
 rem to have the imports be generated correctly. See https://github.com/grpc/grpc/issues/9575#issuecomment-293934506
+rem Note: We currently do not use imports in order to keep OptimizerMonitoringService
```
byte-sculptor, line 7: The comment above is about relative imports in the Python code, not one gRPC module importing another.

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/SeriesDifferenceUtilityFunction.py
```diff
@@ -0,0 +1,98 @@
...
+        ]
+    ),
+    default=Point(
+        utility_function_name="upper_confidence_bound_on_improvement",
+        alpha=0.01
+    )
+)
+
+
+class SeriesDifferenceUtilityFunction(UtilityFunction):
```
byte-sculptor, line 34: Could you please add a comment here describing how this works and what's the intent?

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/SeriesDifferenceUtilityFunction.py
```diff
@@ -0,0 +1,98 @@
...
+            f"context_space.{self.series_objective.series_modulation_dimension.name}": self.series_objective.series_modulation_dimension.linspace()
+        })
+        feature_values_pandas_frame_merged = feature_values_pandas_frame.merge(series_vals_df, how="cross")
+        multi_objective_predictions = self.surrogate_model.predict(features_df=feature_values_pandas_frame_merged)
+
+        predictions = multi_objective_predictions[0]
+        predictions_df = predictions.get_dataframe()
+
+        # TODO ZACK: Vectorize this. I'm sure pandas/numpy can do this faster. @Adam, any suggestions
+        resulting_series_arry = []
```
byte-sculptor, line 68: array?

byte-sculptor: What is resulting_series_arry? What does it hold? Also, why is series_modulation_dimension categorical?

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/SeriesDifferenceUtilityFunction.py
```diff
@@ -0,0 +1,98 @@
...
+        # Later they will diverge as SeriesDifferenceUtilityFunction should require additional logic surrounding probability
+        # of improvement. Perhaps this part should be refactored to be shared though.
+        #
+        self.logger.debug(f"Computing utility values for {len(feature_values_pandas_frame.index)} points.")
+
+        predicted_value_col = Prediction.LegalColumnNames.PREDICTED_VALUE.value
+        predicted_value_var_col = Prediction.LegalColumnNames.PREDICTED_VALUE_VARIANCE.value
+        dof_col = Prediction.LegalColumnNames.PREDICTED_VALUE_DEGREES_OF_FREEDOM.value
+
+        del feature_values_pandas_frame[f"context_space.{self.series_objective.series_modulation_dimension.name}"]
```
byte-sculptor, line 57: It looks like `del` modifies the input array... I'm not sure if it's prudent to mess with the caller's df like that.

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/SeriesDifferenceUtilityFunction.py
```diff
@@ -0,0 +1,98 @@
...
+        series_vals_df = pd.DataFrame({
+            f"context_space.{self.series_objective.series_modulation_dimension.name}": self.series_objective.series_modulation_dimension.linspace()
```
byte-sculptor, line 59: Modulation implies periodicity, no? How about series indexing dimension? Also, that dimension should really be a ContinuousDimension or maybe a DiscreteDimension. But according to all major religions, and philosophies, time is not Categorical :)

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/SeriesDifferenceUtilityFunction.py
```diff
@@ -0,0 +1,98 @@
...
+        })
+        feature_values_pandas_frame_merged = feature_values_pandas_frame.merge(series_vals_df, how="cross")
```
byte-sculptor, line 61: Make index a column of the feature_values_pandas_frame before the merge,. You can group by and filter on that column later.

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/SeriesDifferenceUtilityFunction.py
```diff
@@ -0,0 +1,98 @@
...
+        # TODO ZACK: Vectorize this. I'm sure pandas/numpy can do this faster. @Adam, any suggestions
+        resulting_series_arry = []
+        current_arry = []
+        for index, prediction in predictions_df.iterrows():
+            current_arry.append(prediction[predicted_value_col])
+            if ((index + 1) % len(self.series_objective.series_modulation_dimension)) == 0:
+                resulting_series_arry.append(np.array(current_arry))
+                current_arry=[]
+
+        # Didn't do bootstrapping yet because I think math would be computationally faster
```
byte-sculptor, line 76: The for-loop above is very brittle. Use the 'index column' to group the predictions by the features that created them. And remember to restore the index so that the caller can properly match the utility function value, to the features.

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/SeriesDifferenceUtilityFunction.py
byte-sculptor: Also, why use linspace here, instead of grabbing these values from target_series? That way, they will always be matched up...

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/SeriesDifferenceUtilityFunction.py
byte-sculptor: Got it, it's a predicted value series

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/SeriesDifferenceUtilityFunction.py
```diff
@@ -0,0 +1,98 @@
...
+        utility_function_values = []
+        for series in resulting_series_arry:
+            utility_function_values.append(self._sign * sum((self.series_objective.target_series-series)**2))
+
```
byte-sculptor, line 80: Why do we need the sign here? Aren't we always minimizing this error? Also, here we implicitly commit to sum-of-squared-errors as our metric.. Either specify this explicitly in the class name/class doc string, or add options :)

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/SeriesDifferenceUtilityFunction.py
```diff
@@ -0,0 +1,98 @@
...
+        # TODO ZACK : Something like this with some squared sum of variance term might work...
+        # NOTE:
+        # E[(x-y)^2] = E[X]^2 + 2E[X]E[Y] - E[Y]^2
```
byte-sculptor, line 83: There is an anomaly in this expansion.

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/SeriesDifferenceUtilityFunction.py
```diff
@@ -0,0 +1,98 @@
...
+        # Var[(x-y)^2] = (E[(X-Y)^2] - E[X-Y]^2)^2 (... I think)
```
byte-sculptor, line 84: Could you please point me to the source where you found this formula?

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/SeriesPredictedValueUtilityFunction.py
```diff
@@ -0,0 +1,67 @@
...
+from mlos.Optimizers.OptimizationProblem import SeriesObjective
+from mlos.Optimizers.ExperimentDesigner.UtilityFunctions.UtilityFunction import UtilityFunction
+from mlos.Optimizers.RegressionModels.MultiObjectiveRegressionModel import MultiObjectiveRegressionModel
+from mlos.Optimizers.RegressionModels.Prediction import Prediction
+from mlos.Spaces import SimpleHypergrid, ContinuousDimension, CategoricalDimension, Point
+from mlos.Spaces.Configs.ComponentConfigStore import ComponentConfigStore
+from mlos.Tracer import trace
+
+
+class SeriesPredictedValueUtilityFunction(UtilityFunction):
```
byte-sculptor, line 18: What is the point of this class?

source/Mlos.Python/mlos/Optimizers/OptimizationProblem.py
```diff
@@ -6,20 +6,53 @@
...
+class Objective:
+    def __init__(self, name: str, minimize: bool):
+        self.name = name
+        self.minimize = minimize
 
 
-def objective_from_dict(objective_dict):
-    return Objective(**objective_dict)
+class SeriesObjective(Objective):
+    def __init__(self, name: str, minimize: bool, target_series: np.array, series_modulation_dimension: Dimension):
```
byte-sculptor, line 21: Comments? Also, target_series cannot be an np.array - it has to be a dataframe with one column for time, and another for the target value at that time instance

byte-sculptor: Also, formatting :) Also, stuffing the target_series_df into the Objective, is an elegant solution. 

Some general recommendations:
1. Make both Objective and SeriesObjective a @dataclass 
2. Rename SeriesObjective to something a brand new user can understand, such as SeriesMatchingObjective, or even TimeSeriesMatchingObjective. Do we envision trying to fit any other series other than time series?

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
```diff
@@ -107,10 +113,30 @@ def optimum(self, optimum_definition: OptimumDefinition = OptimumDefinition.BEST
...
+    @trace()
+    def _optimum_with_series_difference(self, context: pd.DataFrame):
+        predicted_value_utility_function = SeriesPredictedValueUtilityFunction(
+            surrogate_model=self.surrogate_model,
+            objective=self.optimization_problem.objectives[0],
+            logger=self.logger
+        )
+        utility_optimizer = UtilityFunctionOptimizerFactory.create_utility_function_optimizer(
+            utility_function=predicted_value_utility_function,
+            optimizer_type_name=RandomSearchOptimizer.__name__,
```
byte-sculptor, line 131: Why RandomSearchOptimizer?

byte-sculptor: Also, at this point we definitely don't want the default config. We want some beefy config that would thoroughly explore the space for maxima.

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
```diff
@@ -107,10 +113,30 @@ def optimum(self, optimum_definition: OptimumDefinition = OptimumDefinition.BEST
...
+            optimizer_config=random_search_optimizer_config_store.default,
+            optimization_problem=self.optimization_problem
+        )
+        return utility_optimizer.suggest(context_values_dataframe=context)
+
     @trace()
     def _optimum_within_context(self, context: pd.DataFrame):
+        # TODO ZACK: Discuss the minimize=... line. Why only objectives[0]?
```
byte-sculptor, line 139: Since nobody ever added support for context to work in multi-objective settings.

source/Mlos.Python/mlos/Optimizers/OptimumDefinition.py
```diff
@@ -45,6 +45,11 @@ class OptimumDefinition(Enum):
 
     BEST_SPECULATIVE_WITHIN_CONTEXT = 'best_speculative_within_context'
 
+    # This mode works to select the points that given a function f(series) is able to output a utility value for that series. This will often be related to
+    # time series approximation however it shouldn't be limited to that.
+    #
+    BEST_SERIES_ERROR_FUNCTION_MINIMIZATION = 'best_series_error_function_minimization'
```
byte-sculptor, line 51: In the optimization problem the user specifies that they want to match time series. And the time-series utility functions compute the relevant goodness of fit metrics and their statistics, and the optimizer will try to spit out either the: best observation (series that most closely matched the target at least once), best predicted value, etc... I don't see why we need a new optimum type for this.

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/SeriesDifferenceUtilityFunction.py
zaporter: series_modulation_dimension was my best idea for how to name the dimension which is modulated across the time series. I chose categorical so that I could feed it a set of values that are of interest. For example, if kusto gives performance metrics at t=0.1, t=0.5, and t=1s, it is foolish to check any other value and try and compare the resulting series. The only dimension type that I thought of that would work in this case is categorical. Predicted_value_series seems like that is the result of the prediction when it is the opposite, it is the input to the prediction. 

Do you have any thoughts for the name now? I thought about just using something like time_dimension but as I will elaborate on later, I think that this functionality could be used for more complex things than just time-modulation and the time-convention doesn't really add that much readability in my mind. It is, to me, just the ability to test across a wide sample of values and converge on a specified series.

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/SeriesDifferenceUtilityFunction.py
zaporter: Good point. Will change.

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/SeriesDifferenceUtilityFunction.py
zaporter: Yes but values in time are.... well not categorical but close. As I wrote earlier, if kusto only gives reports at specified intervals, it seems weird to assume that those intervals are going to be evenly spaced. 

Also, according to all major religions and philosophies, the number of instances between t=0s and t=1s is infinite, and yet this sadly cannot evaluate infinite points (yet..) Putting the values as discrete (and flexible) time intervals was my solution to this problem. Thoughts?

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/SeriesDifferenceUtilityFunction.py
zaporter: Good idea.

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/SeriesDifferenceUtilityFunction.py
zaporter: Sign was added because... well it felt harsh to remove the option.. but I guess it is fairly useless

As per the sum-of-squared-errors, I started with an implementation where the user passed a function of type function(series)=>error, so this model would just pass the series to this arbitrary function and get back the error. I liked this method because of how flexible it was but I realized that it was really slow.

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/SeriesDifferenceUtilityFunction.py
zaporter: I fumbled this together from my memories of my loss models class. I would have to check more before I actually wrote it in code.

source/Mlos.Python/mlos/Optimizers/ExperimentDesigner/UtilityFunctions/SeriesPredictedValueUtilityFunction.py
zaporter: To copy the PredictedValueUtilityFunction class. When we are finding the optimum, we don't want anything to do with probability of improvement, we just want the basic result. 

Thoughts on a way to reduce code duplication?

source/Mlos.Python/mlos/Optimizers/OptimizationProblem.py
zaporter: Ah! This explains a great deal

I have been using series_modulation_dimension to store the time values, but you rather suggest to store it in the series itself... that is far smarter. I will do this tomorrow. 

> Do we envision trying to fit any other series other than time series? > No reason to lock ourselves in right now I think

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
zaporter: Thank you for pointing this out! That was a typo / bug

source/Mlos.Python/mlos/Optimizers/OptimizerBase.py
zaporter: When I saw this, I became very sad because I spent a great deal of time with context-optimization problems that have mutiple objectives... I got terrible performance.

PR #265, opened 2021-11-01
Compute exact pareto volume

amueller:
This took me a while to figure out, but it's faster than the sampling even with a moderate amount of samples, so it might make it easier to do more elaborate computations with the Pareto volume.
This includes #264 because I'm a lazy slob.


amueller:
Hm it doesn't work for minimization right now because I need to fix a reference point that's not 0 for that, I think? Basically this works right now for maximization and assuming both objectives are positive.

PR #277, opened 2023-03-20
Add optimization target and direction to the Optimizer config; implement bulk update

motus:


mlos_bench/mlos_bench/optimizer/base_optimizer.py
```diff
@@ -115,13 +136,44 @@ def register(self, tunables: TunableGroups, status: Status, score: float = None)
...
-        score : float
-            The benchmark result, or None if the experiment was not successful.
+        score : Union[float, dict]
+            A scalar or a dict with the final benchmark results.
+            None if the experiment was not successful.
+
+        Returns
+        -------
+        value : float
+            A scalar benchmark score to MINIMIZE.
```
bpkroth, line 146:
```suggestion
            The scalar benchmark score extracted (and possibly transformed) from the dataframe that's being minimized.
```

bpkroth (approved):
Minor tweak on the return explanation for `register`, but otherwise LGTM

PR #283, opened 2023-03-24
Actual DB schema for storing the optimization configurations and results.

motus:
SQLite3 and PostgreSQL versions of the DB schema. I am not sure I want to use SQLAlchemy or other similar tools. Python-side code is using DB-API 2.0 and is agnostic about DB backends.


motus:
DuckDB tables after 5 iterations of mock environment and mock optimzier:
```
D select * from experiment_config;
┌────────────┬─────────┬───────────────────────────────┬──────────────────────────────────────────┐
│   exp_id   │  descr  │           git_repo            │                git_commit                │
│  varchar   │ varchar │            varchar            │                 varchar                  │
├────────────┼─────────┼───────────────────────────────┼──────────────────────────────────────────┤
│ RedisBench │         │ git@github.com:motus/MLOS.git │ a298cd20ec4cfd036b2498d507f4840c454d7ccc │
└────────────┴─────────┴───────────────────────────────┴──────────────────────────────────────────┘
D select * from trial_status;
┌────────────┬──────────┬─────────────────────────┬────────────────────────────┬───────────┐
│   exp_id   │ trial_id │        ts_start         │           ts_end           │  status   │
│  varchar   │  int32   │        timestamp        │         timestamp          │  varchar  │
├────────────┼──────────┼─────────────────────────┼────────────────────────────┼───────────┤
│ RedisBench │        1 │ 2023-03-24 22:10:06.868 │ 2023-03-24 22:10:06.879444 │ SUCCEEDED │
│ RedisBench │        2 │ 2023-03-24 22:10:06.888 │ 2023-03-24 22:10:06.894458 │ SUCCEEDED │
│ RedisBench │        3 │ 2023-03-24 22:10:06.908 │ 2023-03-24 22:10:06.91012  │ SUCCEEDED │
│ RedisBench │        4 │ 2023-03-24 22:10:06.928 │ 2023-03-24 22:10:06.931741 │ SUCCEEDED │
│ RedisBench │        5 │ 2023-03-24 22:10:06.945 │ 2023-03-24 22:10:06.957595 │ SUCCEEDED │
└────────────┴──────────┴─────────────────────────┴────────────────────────────┴───────────┘
D select * from trial_config;
┌────────────┬──────────┬────────────────────────────────┬───────────────┐
│   exp_id   │ trial_id │            param_id            │  param_value  │
│  varchar   │  int32   │            varchar             │    varchar    │
├────────────┼──────────┼────────────────────────────────┼───────────────┤
│ RedisBench │        1 │ vmSize                         │ Standard_B4ms │
│ RedisBench │        1 │ rootfs                         │ xfs           │
│ RedisBench │        1 │ kernel_sched_migration_cost_ns │ 13111         │
│ RedisBench │        1 │ hz                             │ 380           │
│ RedisBench │        1 │ dynamic_hz                     │ no            │
│ RedisBench │        2 │ vmSize                         │ Standard_B2s  │
│ RedisBench │        2 │ rootfs                         │ xfs           │
│ RedisBench │        2 │ kernel_sched_migration_cost_ns │ 73157         │
│ RedisBench │        2 │ hz                             │ 378           │
│ RedisBench │        2 │ dynamic_hz                     │ yes           │
│ RedisBench │        3 │ vmSize                         │ Standard_B4ms │
│ RedisBench │        3 │ rootfs                         │ ext2          │
│ RedisBench │        3 │ kernel_sched_migration_cost_ns │ 467758        │
│ RedisBench │        3 │ hz                             │ 280           │
│ RedisBench │        3 │ dynamic_hz                     │ yes           │
│ RedisBench │        4 │ vmSize                         │ Standard_B4ms │
│ RedisBench │        4 │ rootfs                         │ ext4          │
│ RedisBench │        4 │ kernel_sched_migration_cost_ns │ 16662         │
│ RedisBench │        4 │ hz                             │ 16            │
│ RedisBench │        4 │ dynamic_hz                     │ yes           │
│ RedisBench │        5 │ vmSize                         │ Standard_B2s  │
│ RedisBench │        5 │ rootfs                         │ xfs           │
│ RedisBench │        5 │ kernel_sched_migration_cost_ns │ 264950        │
│ RedisBench │        5 │ hz                             │ 309           │
│ RedisBench │        5 │ dynamic_hz                     │ yes           │
├────────────┴──────────┴────────────────────────────────┴───────────────┤
│ 25 rows                                                      4 columns │
└────────────────────────────────────────────────────────────────────────┘
D select * from trial_results;
┌────────────┬──────────┬──────────┬───────────────────┐
│   exp_id   │ trial_id │ param_id │    param_value    │
│  varchar   │  int32   │ varchar  │      varchar      │
├────────────┼──────────┼──────────┼───────────────────┤
│ RedisBench │        1 │ score    │ 70.27481964384559 │
│ RedisBench │        2 │ score    │ 69.92906039480249 │
│ RedisBench │        3 │ score    │ 70.66811325996878 │
│ RedisBench │        4 │ score    │ 80.4277082999671  │
│ RedisBench │        5 │ score    │ 70.47284419338608 │
└────────────┴──────────┴──────────┴───────────────────┘
```

mlos_bench/db/db_schema_pg.sql
```diff
@@ -0,0 +1,106 @@
...
+    git_commit VARCHAR(40) NOT NULL,
+
+    PRIMARY KEY (exp_id)
+);
+
+CREATE SEQUENCE trial_id_seq;
+
+CREATE TABLE trial_status (
+    exp_id experiment_id_t,
+    trial_id trial_id_t DEFAULT nextval('trial_id_seq'),
```
bpkroth, line 48: Was dealing with this in benchbase just now, so I'll post here as well:
https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-identity-column/
Postgres added support for a *slightly* more standard way of doing "auto-increment" with `GENERATED ALWAYS AS IDENTITY` syntax.
That's probably a good thing for us to use going forward since it's in the SQL spec there's some hope that other DBs can use it as is as well.

mlos_bench/db/db_schema_sqlite.sql
```diff
@@ -0,0 +1,82 @@
...
+    descr TEXT,
+    git_repo VARCHAR(255) NOT NULL,
+    git_commit VARCHAR(40) NOT NULL,
+
+    PRIMARY KEY (exp_id)
+);
+
+CREATE TABLE trial_status (
+    exp_id VARCHAR(128) NOT NULL,
+    trial_id INTEGER NOT NULL,
```
bpkroth, line 25: I think Sqlite also supports `GENERATED ALWAYS AS IDENTITY` syntax, but we'll have to check.
That seems slightly more preferable way to get an "auto-increment" column than just leaving it as "not-null" and expecting the driver to handle it for us.
Else, we should probably at least leave a comment in the schema that we expect that to be the case.

mlos_bench/db/db_schema_pg.sql
```diff
@@ -0,0 +1,106 @@
...
+DROP TABLE IF EXISTS trial_config CASCADE;
+DROP TABLE IF EXISTS trial_results CASCADE;
+DROP TABLE IF EXISTS trial_telemetry CASCADE;
+
+DROP TYPE IF EXISTS experiment_id_t CASCADE;
+DROP TYPE IF EXISTS trial_id_t CASCADE;
+DROP TYPE IF EXISTS param_id_t CASCADE;
+DROP TYPE IF EXISTS status_t CASCADE;
+
+CREATE DOMAIN experiment_id_t AS VARCHAR(128) NOT NULL;
```
bpkroth, line 19:
```suggestion
CREATE DOMAIN experiment_id_t AS VARCHAR(255) NOT NULL;
```

mlos_bench/db/db_schema_sqlite.sql
```diff
@@ -0,0 +1,82 @@
...
+    exp_id VARCHAR(128) NOT NULL,
+    descr TEXT,
+    git_repo VARCHAR(255) NOT NULL,
+    git_commit VARCHAR(40) NOT NULL,
+
+    PRIMARY KEY (exp_id)
+);
+
+CREATE TABLE trial_status (
+    exp_id VARCHAR(128) NOT NULL,
```
bpkroth, line 24:
```suggestion
    exp_id experiment_id_t NOT NULL,
```
Not using the type defined above.
Or at least bump it up to 255 for all of these I think?

bpkroth: Nm, that's a different file, so only the latter comment applies I guess.

bpkroth:
> DuckDB tables after 5 iterations of mock environment and mock optimzier:
> 
> ```
> D select * from experiment_config;
> ┌────────────┬─────────┬───────────────────────────────┬──────────────────────────────────────────┐
> │   exp_id   │  descr  │           git_repo            │                git_commit                │
> │  varchar   │ varchar │            varchar            │                 varchar                  │
> ├────────────┼─────────┼───────────────────────────────┼──────────────────────────────────────────┤
> │ RedisBench │         │ git@github.com:motus/MLOS.git │ a298cd20ec4cfd036b2498d507f4840c454d7ccc │
> └────────────┴─────────┴───────────────────────────────┴──────────────────────────────────────────┘
> D select * from trial_status;
> ┌────────────┬──────────┬─────────────────────────┬────────────────────────────┬───────────┐
> │   exp_id   │ trial_id │        ts_start         │           ts_end           │  status   │
> │  varchar   │  int32   │        timestamp        │         timestamp          │  varchar  │
> ├────────────┼──────────┼─────────────────────────┼────────────────────────────┼───────────┤
> │ RedisBench │        1 │ 2023-03-24 22:10:06.868 │ 2023-03-24 22:10:06.879444 │ SUCCEEDED │
> │ RedisBench │        2 │ 2023-03-24 22:10:06.888 │ 2023-03-24 22:10:06.894458 │ SUCCEEDED │
> │ RedisBench │        3 │ 2023-03-24 22:10:06.908 │ 2023-03-24 22:10:06.91012  │ SUCCEEDED │
> │ RedisBench │        4 │ 2023-03-24 22:10:06.928 │ 2023-03-24 22:10:06.931741 │ SUCCEEDED │
> │ RedisBench │        5 │ 2023-03-24 22:10:06.945 │ 2023-03-24 22:10:06.957595 │ SUCCEEDED │
> └────────────┴──────────┴─────────────────────────┴────────────────────────────┴───────────┘
> D select * from trial_config;
> ┌────────────┬──────────┬────────────────────────────────┬───────────────┐
> │   exp_id   │ trial_id │            param_id            │  param_value  │
> │  varchar   │  int32   │            varchar             │    varchar    │
> ├────────────┼──────────┼────────────────────────────────┼───────────────┤
> │ RedisBench │        1 │ vmSize                         │ Standard_B4ms │
> │ RedisBench │        1 │ rootfs                         │ xfs           │
> │ RedisBench │        1 │ kernel_sched_migration_cost_ns │ 13111         │
> │ RedisBench │        1 │ hz                             │ 380           │
> │ RedisBench │        1 │ dynamic_hz                     │ no            │
> │ RedisBench │        2 │ vmSize                         │ Standard_B2s  │
> │ RedisBench │        2 │ rootfs                         │ xfs           │
> │ RedisBench │        2 │ kernel_sched_migration_cost_ns │ 73157         │
> │ RedisBench │        2 │ hz                             │ 378           │
> │ RedisBench │        2 │ dynamic_hz                     │ yes           │
> │ RedisBench │        3 │ vmSize                         │ Standard_B4ms │
> │ RedisBench │        3 │ rootfs                         │ ext2          │
> │ RedisBench │        3 │ kernel_sched_migration_cost_ns │ 467758        │
> │ RedisBench │        3 │ hz                             │ 280           │
> │ RedisBench │        3 │ dynamic_hz                     │ yes           │
> │ RedisBench │        4 │ vmSize                         │ Standard_B4ms │
> │ RedisBench │        4 │ rootfs                         │ ext4          │
> │ RedisBench │        4 │ kernel_sched_migration_cost_ns │ 16662         │
> │ RedisBench │        4 │ hz                             │ 16            │
> │ RedisBench │        4 │ dynamic_hz                     │ yes           │
> │ RedisBench │        5 │ vmSize                         │ Standard_B2s  │
> │ RedisBench │        5 │ rootfs                         │ xfs           │
> │ RedisBench │        5 │ kernel_sched_migration_cost_ns │ 264950        │
> │ RedisBench │        5 │ hz                             │ 309           │
> │ RedisBench │        5 │ dynamic_hz                     │ yes           │
> ├────────────┴──────────┴────────────────────────────────┴───────────────┤
> │ 25 rows                                                      4 columns │
> └────────────────────────────────────────────────────────────────────────┘
> D select * from trial_results;
> ┌────────────┬──────────┬──────────┬───────────────────┐
> │   exp_id   │ trial_id │ param_id │    param_value    │
> │  varchar   │  int32   │ varchar  │      varchar      │
> ├────────────┼──────────┼──────────┼───────────────────┤
> │ RedisBench │        1 │ score    │ 70.27481964384559 │
> │ RedisBench │        2 │ score    │ 69.92906039480249 │
> │ RedisBench │        3 │ score    │ 70.66811325996878 │
> │ RedisBench │        4 │ score    │ 80.4277082999671  │
> │ RedisBench │        5 │ score    │ 70.47284419338608 │
> └────────────┴──────────┴──────────┴───────────────────┘
> ```

Wait, DuckDB schema is missing?  Is it handled by one of the others?  Do we have that documented?
Maybe a simple `README.md` in the `db/` folder with a basic description of the schema layout expectations would be a good idea too.

motus:
> Wait, DuckDB schema is missing?

oops, forgot to rename the file. this schema works for both SQLite3 *and* DuckDB

mlos_bench/db/db_schema_sqlite3_duckdb.sql
bpkroth, line 1: Maybe instead of the joint name, we leave the `db_schema_sqlite3.sql` and then add a symlink for `db_schema_duckdb.sql`?

mlos_bench/db/db_schema_sqlite3_duckdb.sql
```diff
@@ -0,0 +1,83 @@
...
+    descr TEXT,
+    git_repo VARCHAR(255) NOT NULL,
+    git_commit VARCHAR(40) NOT NULL,
+
+    PRIMARY KEY (exp_id)
+);
+
+CREATE TABLE trial_status (
+    exp_id VARCHAR(255) NOT NULL,
+    trial_id INTEGER NOT NULL,
```
bpkroth, line 25:
```suggestion
    trial_id INTEGER NOT NULL, -- NOTE: NULL used as implicit auto-increment
```

mlos_bench/db/db_schema_sqlite3_duckdb.sql
```diff
@@ -0,0 +1,83 @@
...
+    ts_start TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
+    ts_end TIMESTAMP,
+    -- Should match the text IDs of `mlos_bench.environment.Status` enum:
+    status VARCHAR(16),
```
bpkroth, line 29: I guess there's no enums in DuckDB/Sqlite?
Maybe add a constraint instead to match the behavior of the postgres schema?
One other option could be to store the `int` value of the status that's in the code and then provide a view for mapping the int to the string description.  I think this is what SqlServer does in a lot of it's system tables for instance.

mlos_bench/db/db_schema_pg.sql
```diff
@@ -0,0 +1,108 @@
...
+    'FAILED',
+    'TIMED_OUT'
+);
+
+CREATE TABLE experiment_config (
+    exp_id experiment_id_t,
+    descr TEXT,
+    git_repo VARCHAR(255) NOT NULL,
+    git_commit VARCHAR(40) NOT NULL,
+
```
bpkroth, line 42: Do we want to store the actual experiment config json content as well?

mlos_bench/db/db_schema_pg.sql
```diff
@@ -0,0 +1,108 @@
...
+    PRIMARY KEY (exp_id)
+);
+
+CREATE SEQUENCE trial_id_seq;
```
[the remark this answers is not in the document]
bpkroth: Also seems to be missing a `DROP SEQUENCE IF EXISTS`

mlos_bench/db/db_schema_pg.sql
```diff
@@ -0,0 +1,108 @@
...
+    trial_id trial_id_t,
+    param_id param_id_t,
+    param_value param_value_t,
+
+    PRIMARY KEY (exp_id, trial_id, param_id),
+    FOREIGN KEY (exp_id, trial_id) REFERENCES trial_status(exp_id, trial_id),
+    FOREIGN KEY (exp_id) REFERENCES experiment_config(exp_id)
+);
+
+CREATE TABLE trial_telemetry (
```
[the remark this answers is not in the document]
bpkroth: Though we then probably also need to think about story for schema evolution, but I might just say "exercise left to the reader" for starters.

mlos_bench/db/db_schema_pg.sql
```diff
@@ -0,0 +1,108 @@
...
+    FOREIGN KEY (exp_id) REFERENCES experiment_config(exp_id)
+);
+
+ALTER SEQUENCE trial_id_seq owned by trial_status.trial_id;
+
+CREATE TABLE experiment_merge (
+    dest_exp_id experiment_id_t,
+    dest_trial_id trial_id_t,
+    source_exp_id experiment_id_t,
+    source_trial_id trial_id_t,
```
bpkroth, line 65: Is there ever a case where we would only include *some* trials from another experiment but not all?
I'm just thinking this could lead to a lot of unnecessary data duplication.

mlos_bench/db/db_schema_pg.sql
```diff
@@ -0,0 +1,108 @@
...
+    PRIMARY KEY (exp_id, trial_id, param_id),
+    FOREIGN KEY (exp_id, trial_id) REFERENCES trial_status(exp_id, trial_id),
+    FOREIGN KEY (exp_id) REFERENCES experiment_config(exp_id)
+);
+
+CREATE TABLE trial_results (
+    exp_id experiment_id_t,
+    trial_id trial_id_t,
+    param_id param_id_t,
+    param_value param_value_t,
```
bpkroth, line 91:
```suggestion
    score_value score_value_t,
```
?

bpkroth: Also, see my comments about the computed scores for FAILED results in #284.
Maybe the right thing to do here is to just submit those scores as `NULL` and let the framework recompute them on demand.

mlos_bench/db/db_schema_pg.sql
motus: uh, that means pulling all includes and scripts... and there is always a risk that we might miss something.
I would probably go with an MVP solution right now, and later we can take a closer look at, say, MLFlow and study how they package and restore the artifacts for the experiments. It's a pretty big issue, and I am afraid we'll spend too much time and effort on it

mlos_bench/db/db_schema_pg.sql
motus: I was thinking about it, and I've decided not to use auto-increment at all, regardless of the DB engine. We actually need a trial id as a sequence starting with 1 for each experiment. Otherwise, the numbering of trials gets very confusing. Also, less variation in syntax across different DBs

mlos_bench/db/db_schema_sqlite3_duckdb.sql
motus: Turned out, by not using PostgreSQL types, we can have a single schema file for *all* SQL DBs I can think of. Let's have a simple schema first and clone it later when we actually need it.

mlos_bench/db/db_schema_sqlite3_duckdb.sql
motus: Let's not autoincrement trial_id - this leads to very confusing trial numbers within each experiment

mlos_bench/db/db_schema_pg.sql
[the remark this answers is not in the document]
motus: turned out, this sequence gets remove automatically because it is owned by a column. Anyway, I've decided to not use autoincrement for now and ditch all postgres-specific features in the first version for simplicity

mlos_bench/db/db_schema_pg.sql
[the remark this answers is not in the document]
motus: I think that storing the telemetry will come in handy pretty soon, and we already have the API and implementation for it, so why not keep it.

mlos_bench/db/db_schema_sqlite3_duckdb.sql
motus: one portable solution would be to have a const table with all possible status values and have an FK to it

motus: I'll add this to the schema now. I think we can even keep using string IDs - storing ints won't save us much space or cpu cycles, but would require an extra join for every user query

mlos_bench/db/db_schema_pg.sql
motus: yep, I've fixed the other PR to allow bulk loading the status data as well. In the DB schema, however, teh status is associated with the trial and not with the particular value. That is, if some trial fails, simply won't have any records in teh results table

mlos_bench/db/db_schema_pg.sql
motus: I would leave merging of the experiments for another PR. I have some ideas on how to implement it, but I want to discuss them with you first. Let's have a meeting next week and chat about it

mlos_bench/db/db_schema_sqlite3_duckdb.sql
motus: nah, never mind - changing foreign keys on the fly is not as easy as I had thought before.. I'll roll back the change. BTW DuckDB does support enums - maybe we can implement it later if we decide to have different dialects of the same schema again. Let's discuss on Monday

mlos_bench/db/db_schema.sql
```diff
@@ -0,0 +1,83 @@
...
+    PRIMARY KEY (exp_id, trial_id, param_id),
+    FOREIGN KEY (exp_id, trial_id) REFERENCES trial_status(exp_id, trial_id),
+    FOREIGN KEY (exp_id) REFERENCES experiment_config(exp_id)
+);
+
+CREATE TABLE trial_results (
+    exp_id VARCHAR(255) NOT NULL,
+    trial_id INTEGER NOT NULL,
+    param_id VARCHAR(255) NOT NULL,
+    param_value VARCHAR(255),
```
bpkroth, line 66:
```suggestion
    score_value VARCHAR(255),
```

For results, to keep them mentally distinct from config parameters, I suggest we rename these to `score_id` and `score_value`

mlos_bench/db/db_schema.sql
```diff
@@ -0,0 +1,83 @@
...
+    ts_start TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
+    ts_end TIMESTAMP,
+    -- Should match the text IDs of `mlos_bench.environment.Status` enum:
+    status VARCHAR(16) NOT NULL,
+
+    PRIMARY KEY (exp_id, trial_id),
+    FOREIGN KEY (exp_id) REFERENCES experiment_config(exp_id)
+);
+
+CREATE TABLE experiment_merge (
```
bpkroth, line 36: Looks like all the supported DBs so far include support for recursive CTEs (`WITH` clauses), so doing the transitive closure to get all trials (both merged and original) for an experiment should be possible with a single query.
Though I'm not sure if sqlalchemy supports that with their ORM and we might end up with some different syntax for different DBs.
Maybe later we can create a `VIEW` in the schema to provide the CTE and then "just" query that with a `SELECT` in the `mlos_bench` code.

bpkroth: https://duckdb.org/docs/sql/query_syntax/with.html
https://www.sqlite.org/lang_with.html
https://dev.mysql.com/doc/refman/8.0/en/with.html

mlos_bench/db/db_schema.sql
```diff
@@ -0,0 +1,83 @@
...
+    UNIQUE (dest_exp_id, dest_trial_id, source_exp_id, source_trial_id),
+
+    FOREIGN KEY (dest_exp_id, dest_trial_id)
+        REFERENCES trial_status(exp_id, trial_id),
+
+    FOREIGN KEY (source_exp_id, source_trial_id)
+        REFERENCES trial_status(exp_id, trial_id)
+);
+
+CREATE TABLE trial_config (
```
bpkroth, line 51: We may need one of these for tables for the experiment level.
For instance:
```
experiment_config:
  param_id: varchar
  param_type: enum
  param_default: varchar
  param_range: varchar
```

bpkroth: Without that we'd have to have the config json on hand for us to be able to tell whether two configs were mergeable.

bpkroth (changes requested):
Missing table for experiment parameters and their ranges?

mlos_bench/db/db_schema.sql
motus: good point! will change

mlos_bench/db/db_schema.sql
motus: let's do it in the next PR. I've added this table, but then figured that it requires an FK to it from `trial_config`, and that, in turn, requires the table to be pre-populated before we run any experiments, which adds extra logic to the unit test fixtures, etc., etc. 

How about we keep the DB schema to the absolute minimum in this PR and add the tunables metadata later when I implement the experiment merge?

bpkroth:
> Missing table for experiment parameters and their ranges?

We discussed this offline.  I think the decision was to add the table, but with the foreign key constraints between `trial_config` and `experiement_params` commented out for now to reduce the size and scope of the changes required in the storage API PR #286

mlos_bench/db/db_schema.sql
```diff
@@ -0,0 +1,83 @@
...
+-- The syntax works for SQLite3, DuckDB, PostgreSQL, and MySQL / MariaDB.
+
+DROP TABLE IF EXISTS trial_telemetry;
+DROP TABLE IF EXISTS trial_results;
+DROP TABLE IF EXISTS trial_config;
+DROP TABLE IF EXISTS experiment_merge;
+DROP TABLE IF EXISTS trial_status;
+DROP TABLE IF EXISTS experiment_config;
+
+CREATE TABLE experiment_config (
```
bpkroth, line 15:
```suggestion
CREATE TABLE experiment (
```

mlos_bench/db/db_schema.sql
```diff
@@ -0,0 +1,83 @@
...
+    exp_id VARCHAR(255) NOT NULL,
+    descr TEXT,
+    git_repo VARCHAR(255) NOT NULL,
+    git_commit VARCHAR(40) NOT NULL,
+
+    PRIMARY KEY (exp_id)
+);
+
+CREATE TABLE trial_status (
```
bpkroth, line 24:
```suggestion
CREATE TABLE trial (
```

mlos_bench/db/db_schema.sql
```diff
@@ -0,0 +1,83 @@
...
+
+    PRIMARY KEY (exp_id, trial_id, param_id),
+    FOREIGN KEY (exp_id, trial_id) REFERENCES trial_status(exp_id, trial_id),
+    FOREIGN KEY (exp_id) REFERENCES experiment_config(exp_id)
+);
+
+CREATE TABLE trial_results (
+    exp_id VARCHAR(255) NOT NULL,
+    trial_id INTEGER NOT NULL,
+    metric_id VARCHAR(255) NOT NULL,
```
bpkroth, line 65: Similar to the conversation of `experiment_params` is there utility in having `experiment_metrics`?
Basically, to alleviate the experiment merge problem we want all trials with an experiment to use a similar set of "scores", right?

motus: We need this to store multiple metrics per experiment - one will be used as an optimization target, but we may still want to collect the others. Note that this is different from telemetry, as we store exactly one value per metric per experiment here.

motus:
> I think the decision was to add the table, but with the foreign key constraints between `trial_config` and `experiement_params` commented out

Got it! I've just pushed a new update

mlos_bench/db/db_schema.sql
motus: done!

mlos_bench/db/db_schema.sql
motus: done!

mlos_bench/db/db_schema.sql
bpkroth: Yeah, that makes sense, but my point was if we want to be able to support reusing trials from experiments we also need to be able to know that the outputs scores from the trials are also compatible, not just the input params.
For instance, if I want to optimize for throughput in one, but I only captured time in another, then I can't reuse those trials even if they're over the same params.
Right now the only way to tell that would be to do a large distinct agg query over all the trials in an experiment.
On the other hand if the experiment logged which scores were available for all trials in it we could just check that.
Basically think of it as a materialized view or a constraint if you want.
Make sense?

motus: oh, I see. I thought it would be enough to fail when we try to load the data from multiple experiments into the optimizer and can't find a certain score.. 
Then let me think about it a bit more. It would probably make sense to save the id of the metric we're optimizing for along with the git info of the experiment configuration.

bpkroth: As discussed previously, not all trial results need have any score (i.e. they could have been failed), but if they do have a score, I think all trials within an experiment should have the same set of score metrics, no?

mlos_bench/db/db_schema.sql
bpkroth, line 1: I hate to keep nickel and diming you on feedback, but I thought of a few other items too:

We'll want to support repeating trials.
This will mean reusing a config (so maybe not storing it twice), but having separate scores/status for that trial.
I think it probably makes the most sense to consider the repeats as new trials, potentially scheduled on different execution environments (different VM instance, region, etc.) and/or different times (e.g. night vs. day, week vs. weekend).

We could consider to add a `parent_trial_id` to the `trial` table that would allow us to lookup the `trial_config` without having to store it twice.

Re: scheduling: we could consider having a `ts_start > now() AND status = PENDING` in the `trial` to mean that it should not be executed yet.

However, for future support for parallel execution, we'll also want to consider *where* it gets executed.
So, we might also need to have an `experiment_environments` table that describes currently deployed environments.
That also will give us the ability to provide a dashboard of things that can be or need to be cleaned up.

Finally, in the future servicification of this, we won't necessarily be able to rely on having all of the config files/scripts locally.
For instance, consider the case where the script dies and we need to recover or watch the results of the experiment again.
Not sure if we want to consider storing all of those in the DB or just references to them such that we can pull them back up (e.g. via the git commit in the experiment id).
There are other considerations here like auth tokens, so I'm not sure we want to or even can solve all of these now, but I thought I'd leave the thoughts for now so we can consider how the schema may have to change to accommodate all of these possible future features.

mlos_bench/db/db_schema.sql
motus: I've added an optimization target info to the experiment, and we can easily enforce that each trial (with status==SUCCEEDED) must have that field present.

Theoretically, we can also declare in advance what metrics each experiment collects and check if all of them are present when saving the data.

On the schema side, we can also enforce that we only store some known metrics for each experiment. That I understand.

What I don't understand is how we can possibly enforce - *on the schema side* - saving *all* known metrics for each trial.
Let's discuss?

mlos_bench/db/db_schema.sql
motus: We already have support for pending trials in the storage. Having status = PENDING and ts_start = NULL is enough, and we have Python logic for it, too.

re: delayed execution, trial location, etc., - I see it this way: each trial can have some additional parameters associated with it, that are not tunables. The idea is similar to const_args` vs. `tunable_params` that we have in the environment. When we execute a trial, we retrieve these const parameters, and pass them into the benchmark environment, which, in turn, can use them to adjust the way the trial is being executed. For example, we can create a pending trial in the storage, and add to it the (non-tunable) parameters like execution time, vm name, number of repeats, etc. That's simple and flexible enough imo - but again, we can add that functionality later once we have an MVP storage implementation in main

bpkroth: All of that makes sense.  I think I would actually reuse `ts_start > now()` and `status = PENDING` as a way of implementing delayed execution, but we can do that later.

Did you have thoughts about repeated trials?  Basically, there's all sorts of reasons why we might want to run `trial config c` `r > 1` times, but that'd be a little wasteful from a config storage standpoint.

Wondering if we should have another indirection table for `config_id` and have individual trials link against that vs. the `parent_trial_id` that I suggested earlier.

mlos_bench/db/db_schema.sql
```diff
@@ -0,0 +1,98 @@
...
+        REFERENCES trial(exp_id, trial_id)
+);
+
+-- Tunable parameters and their types, used for merging the data from several experiments.
+-- Records in this table must match the `mlos_bench.tunables.Tunable` class.
+CREATE TABLE tunable_params (
+    param_id VARCHAR(255) NOT NULL,
+    param_type VARCHAR(32) NOT NULL,  -- One of {int, float, categorical}
+    param_default VARCHAR(255),
+    param_meta VARCHAR(1023),  -- JSON-encoded metadata (range or categories).
```
bpkroth, line 59: Hmm, for `ConfigSpace` mapping we separate `meta` attributes from things like `range` and `values`.
Json encoding everything is probably more flexible from a schema definition standpoint, but for well known columns it seems like it makes it more difficult to query.

mlos_bench/db/db_schema.sql
bpkroth: Enforcing that all possible metrics an experiment declares as exportable by the trial runner scripts I don't think is possible in the schema (though if we declared them ahead of time, we could certainly add it as a good sanity check during the `register` operation to help catch bugs in the user scripts, so sometime to add to the list for later).

Enforcing that the ones the trial does export *are* part of that set is all I was after.

Basically, it's another way to check that the experiments are merge compatible.

For instance, suppose we run two experiments `A` and `B` where we have the same set of tunable params, so that they on the surface look like they're compatible, but for experiments `B` we've changed the scripts to include metrics for both `x` and `y`, whereas `A`'s scripts only collected `x`.  If in this case we want to optimize for `x` (ignoring multi-objective tuning for a moment), then we can simply check that `x` is a declared output of all trials for experiments in `A` and `B` and proceed.  On the other hand, if `B` wants to tune for `y`, then `A`'s trials are no longer useful.
Similar for when experiment `C` comes along.

mlos_bench/db/db_schema.sql
bpkroth: Also, I think for `execution_environment` I would prefer that we have an explicit entity table in the schema since those are real resources with real cost that we will need to be able manage cleaning up over time.

mlos_bench/db/db_schema.sql
motus: ok I've pushed a new version of the schema that implements your `experiment -> trial -> configuration` suggestion, i.e., adds another level of indirection between the trial and its tunable parameters. I'll post an ER diagram below

mlos_bench/db/db_schema.sql
motus: This table is not used now anyway. I think we can work out the details later when we actually populate it and enable all FKs to it.

mlos_bench/db/db_schema.sql
motus: I've pushed a new version of the schema. Here's how it looks like:
osat - public
Note that the `param_type` is now empty and all FK into it are commented out for now. We can work out the details of this table (`meta` and all) later when we actually use it to enforce the types of our parameters and metrics.

bpkroth:
FYI, I added some comments/ideas in a PR against yours here:
https://github.com/motus/MLOS/pull/1

motus:
This PR is superseded by #298

PR #284, opened 2023-03-24
Separate benchmark scores from configurations in the optimzier bulk update

motus:
That makes working with the data more straightforward in both optimizers and the storage.


mlos_bench/mlos_bench/optimizer/base_optimizer.py
```diff
@@ -98,17 +98,36 @@ def __init__(self, tunables: TunableGroups, config: dict):
...
         Pre-load the optimizer with the bulk data from previous experiments.
 
         Parameters
         ----------
-        data : List[dict]
-            Records of tunable values and benchmark scores from other experiments.
-            The data is expected to be in `pandas.DataFrame.to_dict('records')` format.
+        configs : List[dict]
+            Records of tunable values from other experiments.
+        scores : List[float]
```
bpkroth, line 121: In the previous `data` did we also support a `status` column in addition to scores?
For instance, I am wondering how we handle trials that failed and don't have a score, especially since the typical strategy for that is to compute a "bad" score based on all *real* data we've seen so far.
Not storing an explicit score leaves us some flexibility to manufacture the score in the future which is useful for merge experiments where we are trying to reuse data.

motus: You're right about the scores - none of our optimizers currently supports this, but we should retrieve the failed trials data anyway. I'll fix that

mlos_bench/mlos_bench/optimizer/base_optimizer.py
```diff
@@ -98,17 +98,36 @@ def __init__(self, tunables: TunableGroups, config: dict):
...
+            Benchmark results from experiments that correspond to `configs`.
+
+        Returns
+        -------
+        is_successful : bool
+            True if registration was successful, False otherwise.
         """
+        _LOG.debug("Warm-up: %d configs, %d scores", len(configs or []), len(scores or []))
+        return bool(configs and scores)
```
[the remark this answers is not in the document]
motus: Totally agree, we have to check for the size here and throw an exception. I'll update the log message, too

motus (commented):
@bpkroth I've fixed the issues and pushed the update. Please take a look!

mlos_bench/mlos_bench/optimizer/base_optimizer.py
bpkroth: Mostly makes sense.  Not sure if Status should be Optional in this case though.
On the `mlos_core` side it won't know anything about it which is fine.
It's on this shim side where we can compute fabricated "bad" scores based on real data for the "FAILED" status ones.

mlos_bench/mlos_bench/optimizer/mlos_core_optimizer.py
```diff
@@ -62,15 +72,19 @@ def suggest(self) -> TunableGroups:
     def register(self, tunables: TunableGroups, status: Status,
                  score: Union[float, dict] = None) -> float:
         score = super().register(tunables, status, score)
-        # By default, hyperparameters in ConfigurationSpace are sorted by name:
-        df_config = pd.DataFrame(dict(sorted(tunables.get_param_values().items())), index=[0])
-        _LOG.debug("Dataframe:\n%s", df_config)
-        self._opt.register(df_config, pd.Series([score]))
+        # TODO: mlos_core currently does not support registration of failed trials:
```
bpkroth, line 75: As mentioned above - `mlos_core` shouldn't need to know about that.  It just handles scores and data.

`mlos_bench` however can compute some fabricated scores for failed trials for `mlos_core` based on the *real* data it's seen so far.

bpkroth: That can be a future PR though.

bpkroth (approved):

PR #285, opened 2023-03-27
Fix the type checking and conversion for Tunable objects; add more unit tests

motus:
Summary of changes:
* Update type coercion rules for the Tunable objects; allow assignment from strings to numbers (required when restoring data from certain DBs) and from floats to ints (required for some optimizers)
* Fix the comparison for `TunableGroups` and `CovariantTunableGroup` - `dict` already does everything for us
* Move test fixtures for the individual tunables one level down to the tunables/ floder as they are not used anywhere else
* Add more unit tests for type conversions and error handling


bpkroth:
Regarding the float -> int conversion, is that guaranteed to not lose precision?
i.e. do we check that the float being converted doesn't have a fractional component?
Just thinking that we might not always want to do that by default.

mlos_bench/mlos_bench/tunables/tunable.py
```diff
@@ -109,11 +109,13 @@ def value(self, value):
         """
         Set the current value of the tunable.
         """
-        if not self.is_valid(value):
+        # We need this coercion for the values produced by some optimizers
+        # (e.g., scikit-optimize) and for data restored from certain storage
+        # systems (where values can be strings).
+        coerced_value = self._TYPE[self._type](value)
```
bpkroth, line 115: I guess this may raise a `ValueError` if the coersion can't be done?
Should we handle that and try and provide a more useful error message when that happens?
Maybe this also addresses my question about checking for `int("1.0")` being allowed but `int("1.1")` not being allowed.

motus: Yes, that throws `ValueError` when we can't convert, including the case of `int("1.1")`. I'll add a user-friendly detailed log message if that happens, but I think we have to rethrow and crash anyway - bad value here means that either our scripts are wrong, or we had saved some invalid data in the storage

motus:
> Regarding the float -> int conversion, is that guaranteed to not lose precision?

Great point, I'll add that check. This conversion happens when, e.g., scikit-optimize produces a new configuration with all floats, asd we convert some of these values to ints and categoricals. That is, all values in float -> int coercion must convert without loss of precision - and we must check for this.

motus (commented):
I've pushed the fixes re: core review. thank you @bpkroth !

PR #286, opened 2023-04-04
Storage interface + minimal implementation of SQL DB storage

motus:
Summary of changes:
* Define base API for storing and retrieving of the experimental data;
* Implement SQL DB-based storage system. (Tested for DuckDB, SQLite3, MySQL/MariaDB, and PostgreSQL);
* Fix the config loader to simplify the instantiation of `Storage` and `Optimizer` objects from JSON config;
* Fix the `Launcher` class to support Storage instantiation;
* Change the optimization loop in `run_opt.py` to save and restore the benchmark data;
* Make global `Service` functions available in both `Storage` and `Optimizer`;
* Add unit tests for the new `Storage` functionality


conda-envs/mlos_core-3.10.yml
```diff
@@ -28,5 +28,6 @@ dependencies:
     - bump2version
     - licenseheaders
     - pytest-local-badge
+    - sqlalchemy
```
motus, line 31: conda has SQLAlchemy 1.4, and we need v 2.0, hence pip

mlos_bench/config/storage/sqlite3.jsonc
```diff
@@ -0,0 +1,18 @@
+// Local SQLite3 database.
+{
+    "class": "mlos_bench.storage.SqlStorage",
+
+    // Apply --config-path resolution to these config parameters:
+    "resolve_path": ["database"],
```
motus, line 6: A new feature in config_persistence clas loader: look up for file specified in `database` parameter in `--config-path`. This works for all JSON configs including `Environment` and `Optimizer`. (The functionality is a bit intertwined with some other Storage-specific fixes to be extracted as a separate PR, though)

conda-envs/mlos_core-3.10.yml
bpkroth: Honestly not convinced conda is any better than just raw containers, but I guess we go where the devs are at.

conda-envs/mlos_core-3.11.yml
```diff
@@ -28,5 +28,6 @@ dependencies:
     - bump2version
     - licenseheaders
     - pytest-local-badge
+    - sqlalchemy
```
bpkroth, line 31: Suggestion to put this into the `mlos_bench/setup.py` as an `extra_requires` and it should get handled automatically in all of the conda envs then.
(`full` specifier below gets mapped to all different `extra_requires` combos for convenience).

motus: done!

motus:
This PR is superseded by #298

PR #289, opened 2023-04-04
Add a dependabot config for automated security update related PRs

bpkroth:


motus (approved):

motus (approved):
Nice!

bpkroth:
> Nice!

Thanks.  Borrowing a little from benchbase, though apparently the ruby yaml parser they use doesn't handle dates so I had to track that down and fix it.

PR #296, opened 2023-04-07
Devcontainer port conflict tweaks

bpkroth:
Windows (or other platforms) may listen on port 80 and cause conflicts when WSL2 tries to forward ports for both sides so that `localhost` resolution behaves as expected despite the VM technically being a separate host.
When this happens the devcontainer can fail to start due to in ability for nginx docs container to listen on port 80.
To workaround this we simply tell it to listen on port 81 instead.  The devcontainer forwards the port to a different one on the host anyways.

Additionally, we document some commands for cleaning up stale container images in the ACR service to avoid excessive storage usage.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -9,7 +9,7 @@
     "workspaceFolder": "/workspaces/${localWorkspaceFolderBasename}",
     // Use 'forwardPorts' to make a list of ports inside the container available locally.
     "forwardPorts": [
-        "docs-www:80"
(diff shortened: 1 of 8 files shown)


motus (approved):
Nice! I ran into that issue, too on my PC (took me a while to find out what service was using port 80 on Windows :) )

bpkroth:
> Nice! I ran into that issue, too on my PC (took me a while to find out what service was using port 80 on Windows :) )

Same.  It was annoying ... thanks Windows :P

PR #298, opened 2023-04-11
Minimal SQL DB storage implementation

motus:
This is an SQLAlchemy DDL definition of a minimal DB schema and a storage implementation that uses it. This PR supersedes #286 and #283 (where the schema was defined in plain SQL) and adds support for a `config` table. In subsequent PRs, I will add support of the metadata for trial parameters and output metrics.

Summary of changes:
* Define base API for storing and retrieving of the experimental data;
* Define a DB schema in SQLAlchemy;
* Implement SQL DB-based storage system. (Tested for DuckDB, SQLite3, MySQL/MariaDB, and PostgreSQL);
* Fix the config loader to simplify the instantiation of `Storage` and `Optimizer` objects from JSON config;
* Fix the `Launcher` class to support Storage instantiation;
* Change the optimization loop in `run_opt.py` to save and restore the benchmark data;
* Make global `Service` functions available in both `Storage` and `Optimizer`;
* Add unit tests for the new `Storage` functionality


mlos_bench/mlos_bench/storage/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+"""
+Interfaces to the storage backends for OS Autotune.
+"""
+
+from mlos_bench.storage.base_storage import Storage
+from mlos_bench.storage.sql_storage import SqlStorage
+
+__all__ = [
+    'Storage',
+    'SqlStorage',
```
bpkroth, line 14: Might make sense to skip loading the SQL one by default in case they aren't using that.  Since it's the only option for now, call it future work perhaps?

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -0,0 +1,231 @@
...
+        Parameters
+        ----------
+        exp_id : str
+            Unique identifier of the experiment.
+        trial_id : int
+            Starting number of the trial.
+        description : str
+            Human-readable description of the experiment.
+        opt_target : str
+            Name of metric we're optimizing for.
```
bpkroth, line 60: Also need the direction (min or max).

mlos_bench/mlos_bench/storage/sql_experiment.py
```diff
@@ -0,0 +1,178 @@
...
+        super().__enter__()
+        with self._engine.begin() as conn:
+            # Get git info and the last trial ID for the experiment.
+            exp_info = conn.execute(
+                text("""
+                SELECT e.git_repo, e.git_commit, MAX(t.trial_id) AS trial_id
+                FROM experiment AS e
+                LEFT OUTER JOIN trial AS t ON (e.exp_id = t.exp_id)
+                WHERE e.exp_id = :exp_id
+                GROUP BY e.git_repo, e.git_commit
```
bpkroth, line 49: I guess this is fine for now, but since we went through the effort of getting sql alchemy in use for schema definitions, does it also make sense to "just" use their ORM APIs as well now?

mlos_bench/mlos_bench/storage/sql_experiment.py
```diff
@@ -0,0 +1,178 @@
...
+        _LOG.debug("Create trial: %s:%d", self._experiment_id, self._trial_id)
+        with self._engine.begin() as conn:
+            try:
+                config_id = self._get_config_id(conn, tunables)
+                conn.execute(self._schema.trial.insert().values(
+                    exp_id=self._experiment_id,
+                    trial_id=self._trial_id,
+                    config_id=config_id,
+                    ts_start=datetime.now(),
+                    status='PENDING'
```
bpkroth, line 166:
```suggestion
                    status='PENDING',
```

mlos_bench/mlos_bench/storage/sql_schema.py
```diff
@@ -0,0 +1,143 @@
...
+            Column("metric_value", String(255)),
+
+            UniqueConstraint("exp_id", "trial_id", "ts", "metric_id"),
+            ForeignKeyConstraint(["exp_id", "trial_id"],
+                                 [self.trial.c.exp_id, self.trial.c.trial_id]),
+        )
+
+        _LOG.debug("Schema: %s", self._meta)
+
+    def create(self):
```
bpkroth, line 137: Request to add a `dump()` function so that we could do something like this to get the raw schema sql text for debugging:
```sh
python3
>> ...
```
```python
import mlos_bench.mlos_bench.storage.sql_schema as sql_schema
print(sql_schema.dump())
```

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -0,0 +1,231 @@
...
+            """
+            Load (tunable values, benchmark scores) to warm-up the optimizer.
+            This call returns data from ALL merged-in experiments and attempts
+            to impute the missing tunable values.
+            """
+
+        @abstractmethod
+        def pending(self):
+            """
+            Return an iterator over the pending experiment runs.
```
bpkroth, line 127:
```suggestion
            Return an iterator over the pending trial runs for this experiment.
```

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -0,0 +1,231 @@
...
+                             exc_info=(exc_type, exc_val, exc_tb))
+            return False  # Do not suppress exceptions
+
+        def __repr__(self) -> str:
+            return self._experiment_id
+
+        @abstractmethod
+        def merge(self, experiment_ids: List[str]):
+            """
+            Merge in the results of other experiments.
```
bpkroth, line 108:
```suggestion
            Merge in the results of other (compatible) experiments trials.
            Used to help warm up the optimizer for this experiment.
```

mlos_bench/mlos_bench/launcher.py
```diff
@@ -142,3 +147,11 @@ def load_env(self) -> Environment:
         return self._config_loader.load_environment(
             self._env_config_file, self._global_config,
             service=LocalExecService(parent=self._config_loader))
+
+    def load_generic(self, env: Environment, cls, json_file_name: str):
```
bpkroth, line 151: Prefer the old method of `new()` and `load()` in part due to type hint check ability I think.

bpkroth: Though, if `cls` is actually a type (it's not clear from this), but below it looks to be.

Then, I think we can use `TypeVar` tricks to make this look like it returns a bound-class/generic so that the passed in var is also the type it needs to return.

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -0,0 +1,231 @@
...
+            self._tunables = tunables  # No need to copy, it's immutable
+            self._experiment_id = experiment_id
+            self._trial_id = trial_id
+            self._description = description
+            self._opt_target = opt_target
+            (self._git_repo, self._git_commit) = get_git_info()
+
+        def __enter__(self):
+            """
+            Enter the context of the experiment.
```
bpkroth, line 86: I don't feel like these are the most self explanatory without really knowing a ton of python guts.
Just for readability, could we have other functions that subclasses must implement:

```python
@abstractmethod
def _setup(self):
    ...

@abstractmethod
def _cleanup(self):
    ...
```

And then have `__enter__` and `__exit__` call those?

bpkroth: Alternatively, add something to the docstrings that makes it very clear that the `__enter__` `__exit__` functions serve those same purposes and that `Experiments` are only ever expected to be used within a `with ... as:` context.

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -0,0 +1,231 @@
...
+        @abstractmethod
+        def load(self, opt_target: Optional[str] = None) -> Tuple[List[dict], List[float]]:
+            """
+            Load (tunable values, benchmark scores) to warm-up the optimizer.
+            This call returns data from ALL merged-in experiments and attempts
+            to impute the missing tunable values.
+            """
+
+        @abstractmethod
+        def pending(self):
```
bpkroth, line 125:
```suggestion
        def pending_trials(self):
```

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -0,0 +1,231 @@
...
+            """
+            Return an iterator over the pending experiment runs.
+            """
+
+        @abstractmethod
+        def trial(self, tunables: TunableGroups, config: Optional[Dict[str, Any]] = None):
```
bpkroth, line 131:
```suggestion
        def new_trial(self, tunables: TunableGroups, config: Optional[Dict[str, Any]] = None):
```

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -0,0 +1,231 @@
...
+            else:
+                _LOG.warning("Finishing experiment: %s", self,
+                             exc_info=(exc_type, exc_val, exc_tb))
+            return False  # Do not suppress exceptions
+
+        def __repr__(self) -> str:
+            return self._experiment_id
+
+        @abstractmethod
+        def merge(self, experiment_ids: List[str]):
```
bpkroth, line 106: We'll have to add methods to discover the list of available ids as well.

bpkroth: (future work I think)

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -0,0 +1,231 @@
...
+            tunables : TunableGroups
+                Tunable parameters of the experiment.
+            config : dict
+                Key/value pairs of additional non-tunable parameters of the trial.
+
+            Returns
+            -------
+            trial : Storage.Trial
+                An object that allows to update the storage with
+                the results of the experiment run.
```
bpkroth, line 146:
```suggestion
                the results of the experiment trial run.
```

mlos_bench/mlos_bench/storage/sql_experiment.py
bpkroth: Looks like we're already doing that elsewhere?

mlos_bench/mlos_bench/storage/sql_experiment.py
```diff
@@ -0,0 +1,178 @@
...
+                ))
+                if config is not None:
+                    self._save_params(
+                        conn, self._schema.trial_param, config,
+                        exp_id=self._experiment_id, trial_id=self._trial_id)
+                trial = Trial(self._engine, tunables, self._experiment_id,
+                              self._trial_id, self._opt_target, config)
+                self._trial_id += 1
+                return trial
+            except Exception:
```
bpkroth, line 176:
```suggestion
            except SQLException:
```
?

mlos_bench/mlos_bench/storage/sql_experiment.py
```diff
@@ -0,0 +1,178 @@
...
+                yield Trial(self._engine,
+                            self._tunables.copy().assign(tunables).reset(),
+                            self._experiment_id, trial.trial_id, self._opt_target, config)
+
+    def _get_config_id(self, conn, tunables: TunableGroups) -> int:
+        """
+        Get the config ID for the given tunables. If the config does not exist,
+        create a new record for it.
+        """
+        config_hash = hashlib.sha256(str(tunables).encode('utf-8')).hexdigest()
```
bpkroth, line 141: `... str(tunable) ...`

Do we ensure that this is always coming in a canonical form?

e.g. `json.dumps(indent=None, sort_keys=True)` or something like that.

mlos_bench/mlos_bench/tests/storage/exp_load_test.py
```diff
@@ -0,0 +1,105 @@
...
+    assert trial_ids == pending_ids
+
+
+def test_exp_trial_pending_fail(exp_storage_memory_sql: Storage.Experiment,
+                                tunable_groups: TunableGroups):
+    """
+    Start a trial, fail it, and and check that it is NOT pending.
+    """
+    trial = exp_storage_memory_sql.trial(tunable_groups)
+    trial.update(Status.FAILED)
```
bpkroth, line 67: New test to make sure we can't update a trial after it's already been updated to non-PENDING?

bpkroth: Maybe also add that to the clauses in the `update()` SQL?

mlos_bench/mlos_bench/tests/storage/trial_config_test.py
```diff
@@ -0,0 +1,28 @@
...
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+#
+"""
+Unit tests for saving and retrieving additional parameters of pending trials.
+"""
+
+from mlos_bench.storage import Storage
+from mlos_bench.tunables import TunableGroups
+
```
bpkroth, line 11: Tests for config hash uniqueness, repeatability, reusability?

mlos_bench/mlos_bench/util.py
```diff
@@ -100,3 +71,27 @@ def check_required_params(config: Dict[str, Any], required_params: Iterable[str]
...
+    Returns
+    -------
+    (git_repo, git_commit) : Tuple[str, str]
+        Git repository URL and last commit hash.
+    """
+    git_repo = subprocess.check_output(
+        ["cd", path, "&&", "git", "remote", "get-url", "origin"],
+        shell=True, text=True).strip()
+    git_commit = subprocess.check_output(
+        ["cd", path, "&&", "git", "rev-parse", "HEAD"],
```
bpkroth, line 94: This may not work for Windows (though not sure if we care anymore).

Tests for it?

mlos_bench/mlos_bench/tests/storage/exp_load_test.py
```diff
@@ -0,0 +1,105 @@
...
+    config1 = tunable_groups.copy().assign({'rootfs': 'ext4'})
+    config2 = tunable_groups.copy().assign({'rootfs': 'ext2'})
+    trial_ids = {
+        exp_storage_memory_sql.trial(config1).trial_id,
+        exp_storage_memory_sql.trial(config2).trial_id,
+        exp_storage_memory_sql.trial(config2).trial_id,  # Submit same config twice
+    }
+    pending_ids = {pending.trial_id for pending in exp_storage_memory_sql.pending()}
+    assert len(pending_ids) == 3
+    assert trial_ids == pending_ids
```
bpkroth, line 58: Could add a check here that there's still only two configs as well in the DB.

mlos_bench/config/storage/mysql.jsonc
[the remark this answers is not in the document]
motus: Totally agree! I've created #308 for it so we can remember

mlos_bench/mlos_bench/storage/__init__.py
motus: you're right, we don't use it anywhere directly. Will remove the export.

mlos_bench/mlos_bench/storage/sql_experiment.py
motus: fixed!

mlos_bench/mlos_bench/storage/base_storage.py
motus: we don't, actually. All we need here is a metric name to retrieve from storage; we don't care about the direction as long as we took the right measurements in the earlier trials.

Later we will pass the right metadata for multi-target optimization along with different directions for each metric, etc., but for now this is the absolute minimum we need to save and restore the trial data. We'll add more stuff in subsequent PRs

mlos_bench/mlos_bench/storage/sql_experiment.py
motus: I agree, let's stick to one standard. I'll push an update soon

mlos_bench/mlos_bench/storage/sql_experiment.py
motus: yes, `TunableGroups.__repr__()` produces a canonical form, although it drops some information (e.g., cost). Later we can implement both `.__str__()` and `.__repr__()` if we want both canonical and more user-friendly string representations.

mlos_bench/mlos_bench/storage/base_storage.py
motus: agree

mlos_bench/mlos_bench/storage/sql_experiment.py
motus: nope. I think we should rollback and re-throw on any exception.

mlos_bench/mlos_bench/storage/base_storage.py
motus: Good idea. I've added `_setup`/`_teardown` methods to the storage. (I don't decorate them as `@abstractmethod` because some implementations might not need to override them)

mlos_bench/mlos_bench/storage/base_storage.py
motus: done

mlos_bench/mlos_bench/storage/base_storage.py
motus: done

mlos_bench/mlos_bench/storage/sql_schema.py
motus: yep. will add it now

mlos_bench/mlos_bench/storage/__init__.py
motus: Actually, no. I've just rolled back this change. We need this export to instantiate `SqlStorage` from JSON config - the same way we do it for `Environment` and `Optimizer` classes.

mlos_bench/mlos_bench/util.py
motus: I agree, we need unit tests for it. (It seems to work for both Windows and Linux, btw)

mlos_bench/mlos_bench/tests/storage/trial_config_test.py
motus: Storage API currently does not have any methods to retrieve the individual configs, and I am not sure I want to expose that functionality on the mlos_bench side... For unit tests, however, I can tap into SQLite directly and check if the data is there. Let's discuss

mlos_bench/mlos_bench/tests/storage/exp_load_test.py
motus: same as the above:
> Storage API currently does not have any methods to retrieve the individual configs, and I am not sure I want to expose that functionality on the mlos_bench side... For unit tests, however, I can tap into SQLite directly and check if the data is there. Let's discuss

mlos_bench/mlos_bench/launcher.py
motus: I agree, I am not quite happy with that `.load_generic()` method myself. OTOH, defining `.load()` and `.new()` in every loadable class is a lot of code duplication, and it does not solve the root problem with the types, only pushes it deeper down into the `util.instantiate_from_config()` function.

I really like the `TypeVar` idea, though - let me play with it and I'll report to you later

mlos_bench/mlos_bench/storage/base_storage.py
bpkroth: Leaving multi-objective aside for now, since I agree that's future work, we do however already support minimization vs. maximizing the given objective target.
The framework needs to know that when restarting an experiment or merging them, correct?
Or is the assumption that we'll be able to pull that from the local config files still?
In that case, do we need to store the opt_target in the DB at all?

mlos_bench/mlos_bench/storage/__init__.py
bpkroth: I don't follow.  Can't we "just" specify the full module path in the json files?

bpkroth: Some reasons I wonder:
- There's generally been a principle that when this package is deployed, it should be loadable without requiring _all_ of the various supported dependencies.  For instance, in the `mlos_core` pacakge, we try to not require you to have `emukit` installed if you aren't using it.  Similarly, here if we aren't using a SQL backend, it'd be nice not to require the sqlalchemy package to be installed.  This is something we could probably fix later once we have multiple backends, but I suspect we'd forget.
- In the course of doing a bunch of the mypy work I ran across a lot of circular dependency issues with loading from a module higher in the stack as opposed to something lower down and more specific.  Maybe that's something we can handle just internally moreso, but in the spirit of least privileges/dependencies, it seemed worth considering.

Anyways, not a strong preference, just something to consider.

mlos_bench/mlos_bench/tests/storage/trial_config_test.py
bpkroth: I think retrieving configs is going to be important when we get to the point of visualizations and analysis, but we can probably add it later.

mlos_bench/mlos_bench/util.py
bpkroth: The `&&` syntax was the main part I wasn't sure about for `pwsh` vs `powershell` vs `cmd`.
However, you could do something like `["git", "-C", path, "rev-parse", "HEAD"]` to use git's builtin ability to move to a target directory, which should be cross platform and avoid extra fork/exec issues.

mlos_bench/mlos_bench/storage/__init__.py
motus: ok I'm convinced 😄 I've re-removed the export of `SqlStorage` and changed our JSONC configs so that they import this class directly as `mlos_bench.storage.sql_storage.SqlStorage`

mlos_bench/mlos_bench/storage/base_storage.py
motus: I think we've resolved that during our conversation today. We don't store the `opt_target` in the DB now, and this parameter is just for sanity check to make sure we retrieve the right metric when we load the data from previous trials. I will add a comment explaining it in the code.

mlos_bench/mlos_bench/util.py
motus: great point! I'll fix the code and add unit tests

mlos_bench/mlos_bench/launcher.py
motus: I think I've been overengineering it. Let's just have two functions, `.load_optimizer()` and `.load_storage()` - I don't think we'll have any other major subsystems in addition to our existing `Environment` / `Optimizer` / `Storage` trio anyway.

mlos_bench/mlos_bench/launcher.py
```diff
@@ -9,9 +9,11 @@
 import logging
 import argparse
 
-from typing import List, Dict, Any
+from typing import Optional, List, Dict, Any
```
bpkroth, line 12: I think this will get resolved with the merge conflicts, but I tried to make this sorted

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -0,0 +1,274 @@
...
+            config = self._config.copy()
+            config.update(global_config or {})
+            config["experimentId"] = self._experiment_id
+            config["trialId"] = self._trial_id
+            return config
+
+        @abstractmethod
+        def update(self, status: Status,
+                   value: Optional[Union[Dict[str, Any], Any]] = None
+                   ) -> Optional[Dict[str, Any]]:
```
bpkroth, line 237: Complete nit but we (myself included) are rather inconsistent with how we format function signatures with multiple lines.
In some the closing paren and return is on its own line like here, others (next function), they're grouped together with other params.
Some has multiple args per line, others one arg per line.
I don't care to fix this here necessarily, but maybe for the future we just pick one and stick to it.
I think my preference (probably borrowed from SQL) is either: 
- all on one line (if they fit), or
- one per line otherwise

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -0,0 +1,274 @@
...
+            _LOG.info("Store trial: %s :: %s %s", self, status, value)
+            if isinstance(value, dict) and self._opt_target not in value:
+                _LOG.warning("Trial %s :: opt. target missing: %s", self, self._opt_target)
+                # raise ValueError(
+                #     f"Optimization target '{self._opt_target}' is missing from {value}")
+            return {self._opt_target: value} if isinstance(value, (float, int)) else value
+
+        @abstractmethod
+        def update_telemetry(self, status: Status,
+                             value: Optional[Dict[str, Any]] = None) -> None:
```
bpkroth, line 263: same here: `Any` type hint, but docstring says `float`

mlos_bench/mlos_bench/storage/base_storage.py
bpkroth: Same for function calls.

mlos_bench/mlos_bench/storage/sql_experiment.py
```diff
@@ -0,0 +1,209 @@
...
+        cur_config = conn.execute(self._schema.config.select().where(
+            column("config_hash") == config_hash
+        )).fetchone()
+        if cur_config is not None:
+            return int(cur_config.config_id)  # mypy doesn't know it's always int
```
bpkroth, line 168: Hmm, that's too bad.  I would have hoped that that came from the schema details that we setup in sqlalchemy.

mlos_bench/mlos_bench/storage/sql_schema.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def create(self) -> 'DbSchema':
+        """
+        Create the DB schema.
+        """
+        _LOG.info("Create the DB schema")
+        self._meta.create_all(self._engine)
+        return self
+
+    def __repr__(self) -> str:
```
bpkroth, line 166: Maybe a small comment here about how it can be used to dump the schema to text?

mlos_bench/mlos_bench/storage/sql_storage.py
```diff
@@ -0,0 +1,52 @@
...
+    An implementation of the Storage interface using SQLAlchemy backend.
+    """
+
+    def __init__(self, tunables: TunableGroups, service: Optional[Service], config: dict):
+        super().__init__(tunables, service, config)
+        url = URL.create(**self._config)
+        self._repr = f"{url.get_backend_name()}:{url.database}"
+        _LOG.info("Connect to the database: %s", self)
+        # Log SQL statements if the logging level is DEBUG:
+        self._engine = create_engine(url)  # , echo=_LOG.isEnabledFor(logging.DEBUG))
```
bpkroth, line 34:
```suggestion
        self._engine = create_engine(url)
```

?

mlos_bench/mlos_bench/storage/sql_experiment.py
bpkroth: Sounds good.  Maybe we can still add a test for it:

e.g.
```python
# Python uses ordered dicts.
config_a = json.loads('{ "a": 1, "b": 2 }')
config_b = json.loads('{ "b": 2, "a": 1 }')
assert hash(config_a) == hash(config_b)
```

mlos_bench/mlos_bench/tests/storage/trial_config_test.py
```diff
@@ -0,0 +1,62 @@
...
+    config = {"location": "westus2", "num_repeats": 100}
+    trial = exp_storage_memory_sql.new_trial(tunable_groups, config)
+    (pending,) = list(exp_storage_memory_sql.pending_trials())
+    assert pending.trial_id == trial.trial_id
+    assert pending.tunables == tunable_groups
+    assert pending.config() == {
+        "location": "westus2",
+        "num_repeats": "100",
+        "experimentId": "Test-001",
+        "trialId": 1,
```
bpkroth, line 27: `trialId` and `experimentId` aren't part of the `.config()`, right?
They're part of the trial metadata only?

mlos_bench/mlos_bench/tests/util_git_test.py
```diff
@@ -0,0 +1,20 @@
...
+
+from mlos_bench.util import get_git_info
+
+
+def test_get_git_info() -> None:
+    """
+    Check that we can retrieve git info about the current repository correctly.
+    """
+    (git_repo, git_commit) = get_git_info(os.path.dirname(__file__))
+    assert "mlos" in git_repo.lower()
```
bpkroth, line 19: This will fail if you have multiple checkouts.  For instance, I have locally:
```
src/MLOS/MLOS
src/MLOS/MLOS.2
src/MLOS/MLOS.3
...
```

bpkroth: Maybe `startswith` is good enough for now?

bpkroth: Nm, you're parsing the remote ... maybe that's ok for now.

mlos_bench/mlos_bench/util.py
```diff
@@ -100,3 +71,25 @@ def check_required_params(config: Dict[str, Any], required_params: Iterable[str]
         raise ValueError(
             "The following parameters must be provided in the configuration"
             + f" or as command line arguments: {missing_params}")
+
+
+def get_git_info(path: str = ".") -> Tuple[str, str]:
```
bpkroth, line 76:
```suggestion
def get_git_info(path: str = __file__) -> Tuple[str, str]:
```

I think otherwise this will fail if it's run with `$PWD` != somewhere in the repo.

mlos_bench/mlos_bench/service/config_persistence.py
```diff
@@ -108,6 +109,75 @@ def load_config(self, json_file_name: str) -> Union[dict, List[dict]]:
         with open(json_file_name, mode='r', encoding='utf-8') as fh_json:
             return json5.load(fh_json)  # type: ignore[no-any-return]
 
+    def prepare_class_load(self, config: dict,
+                           global_config: dict = None) -> Tuple[str, Dict[str, Any]]:
```
bpkroth, line 113:
```suggestion
                           global_config: Optional[dict] = None) -> Tuple[str, Dict[str, Any]]:
```

mlos_bench/mlos_bench/service/config_persistence.py
```diff
@@ -108,6 +109,75 @@ def load_config(self, json_file_name: str) -> Union[dict, List[dict]]:
...
+            class_config[key] = self.resolve_path(class_config[key])
+
+        if _LOG.isEnabledFor(logging.DEBUG):
+            _LOG.debug("Instantiating: %s with config:\n%s",
+                       class_name, json.dumps(class_config, indent=2))
+
+        return (class_name, class_config)
+
+    def build_generic(self, cls, tunables: TunableGroups, service: Service,
+                      config: dict, global_config: dict = None):
```
bpkroth, line 148:
```suggestion
                      config: dict, global_config: Optional[dict] = None):
```

mlos_bench/mlos_bench/service/config_persistence.py
```diff
@@ -108,6 +109,75 @@ def load_config(self, json_file_name: str) -> Union[dict, List[dict]]:
...
+        # pylint: disable=too-many-arguments
```
bpkroth, line 149: Pull the same `*` trick here too?

mlos_bench/mlos_bench/service/config_persistence.py
```diff
@@ -108,6 +109,75 @@ def load_config(self, json_file_name: str) -> Union[dict, List[dict]]:
...
+        for key in class_params.intersection(config.get("resolve_path", [])):
+            class_config[key] = self.resolve_path(class_config[key])
+
+        if _LOG.isEnabledFor(logging.DEBUG):
+            _LOG.debug("Instantiating: %s with config:\n%s",
+                       class_name, json.dumps(class_config, indent=2))
+
+        return (class_name, class_config)
+
+    def build_generic(self, cls, tunables: TunableGroups, service: Service,
```
bpkroth, line 147: Is this still used?  I'd made a comment about return type hints problems on this in a previous version ...

mlos_bench/mlos_bench/service/config_persistence.py
```diff
@@ -108,6 +109,75 @@ def load_config(self, json_file_name: str) -> Union[dict, List[dict]]:
...
+            An optional service object (e.g., providing methods to load config files, etc.)
+        config : dict
+            Configuration of the class to instantiate, as loaded from JSON.
+        global_config : dict
+            Global configuration parameters (optional).
+
+        Returns
+        -------
+        inst : Storage
+            A new instance of the Storage class.
```
bpkroth, line 174: Inaccurate docstring?

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -0,0 +1,274 @@
...
+            """
+            config = self._config.copy()
+            config.update(global_config or {})
+            config["experimentId"] = self._experiment_id
+            config["trialId"] = self._trial_id
+            return config
+
+        @abstractmethod
+        def update(self, status: Status,
+                   value: Optional[Union[Dict[str, Any], Any]] = None
```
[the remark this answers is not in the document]
motus: agree. I'll change it to `float`

mlos_bench/mlos_bench/storage/base_storage.py
motus: yep. `float` it is

mlos_bench/mlos_bench/launcher.py
motus: yep. I've made another pass over the code after resolving the conflicts to make sure that mypy/pylint/flake8/pydocstyle are all happy

mlos_bench/mlos_bench/storage/sql_storage.py
motus: yeah, I have to do it properly :)

mlos_bench/mlos_bench/tests/storage/trial_config_test.py
motus: correct. here we test that we can save and retrieve some additional non-tunable parameters of the trial, such as scheduled execution time, location, etc.

mlos_bench/mlos_bench/service/config_persistence.py
motus: fixed

mlos_bench/mlos_bench/service/config_persistence.py
motus: oh that's right. great idea. I think that later I'll scan through all the code and see where else we have that pylint exception and fix it

mlos_bench/mlos_bench/service/config_persistence.py
motus: yes, we use it internally in `Launcher.load_{optimizer,storage}`. I think it's fine to use it like that when we enforce the input and output types

mlos_bench/mlos_bench/service/config_persistence.py
motus: oops, good catch

mlos_bench/mlos_bench/service/config_persistence.py
motus: fixed

mlos_bench/mlos_bench/storage/sql_experiment.py
motus: yeah, I believe that SQLAlchemy just cannot guarantee that it can return the right type for any backend it supports. Anyway, it should be a safe cast here

mlos_bench/mlos_bench/util.py
motus: great point. I'll fix it and add more checks in teh unit test

mlos_bench/mlos_bench/storage/base_storage.py
motus: I agree. I'll do it in another PR (I like to separate cosmetic changes from functionality updates)

mlos_bench/mlos_bench/storage/sql_schema.py
motus: done!

mlos_bench/mlos_bench/storage/sql_experiment.py
motus: ok I've added some unit tests for this, too. I had to update the data in other unit tests, though, because I have touched the TunableGroups fixture and now all optimizers produce new results :)

mlos_bench/mlos_bench/optimizer/base_optimizer.py
```diff
@@ -23,56 +23,7 @@ class Optimizer(metaclass=ABCMeta):
...
-
-        Returns
-        -------
-        opt : Optimizer
-            An instance of the `Optimizer` class initialized with `config`.
-        """
-        return instantiate_from_config(cls, class_name, tunables, config)
-
-    def __init__(self, tunables: TunableGroups, config: dict):
+    def __init__(self, tunables: TunableGroups, service: Optional[Service], config: dict):
```
bpkroth, line 26: Total nit, but it'd be nice if we were consistent between the different class structures on order and/or required naming of the params.
I think for Environment, it's "config, global_config, tunables, service", for instance.
Here, there's an optional param that appears before the config.  It's just a little haphazard.
Maybe cleanup slighly in a future PR?

mlos_bench/mlos_bench/service/config_persistence.py
```diff
@@ -12,15 +12,15 @@
 import json    # For logging only
 import logging
 
-from typing import List, Iterable, Optional, Union
+from typing import Optional, Union, List, Tuple, Dict, Iterable, Any
```
bpkroth, line 15:
```suggestion
from typing import Any, Dict, Iterable, List, Optional, Union, Tuple
```
nit

mlos_bench/mlos_bench/service/config_persistence.py
```diff
@@ -108,8 +110,83 @@ def load_config(self, json_file_name: str) -> Union[dict, List[dict]]:
...
+                       class_name, json.dumps(class_config, indent=2))
+
+        return (class_name, class_config)
+
+    def build_generic(self, *,
+                      cls: type,
+                      tunables: TunableGroups,
+                      service: Service,
+                      config: Dict[str, Any],
+                      global_config: Optional[Dict[str, Any]] = None) -> Any:
```
bpkroth, line 156:
```suggestion
                      global_config: Optional[Dict[str, Any]] = None) -> Union["Storage", "Optimizer", "Service"]:
```

Kinda prefer exact types if we can make it work since we went through all the effort to get mypy checking working.
Returning "Any" provides us almost no help in either development environment, or static or runtime sanity checks.

mlos_bench/mlos_bench/storage/sql_schema.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        Produce a string with all SQL statements required to create the schema
+        from scratch in current SQL dialect.
+
+        That is, return a collection of CREATE TABLE statements and such.
+        NOTE: this method is quite heavy! We use it only once at startup
+        to log the schema, and if the logging level is set to DEBUG.
```
bpkroth, line 174:
```suggestion
        to log the schema, and if the logging level is set to DEBUG.
        
        This can also be done manually like so:
        python -e "from mlos_bench.storage.sql.schema import DbSchema; print(DbSchema);"
```

In fact, if we wanted, we could do something like the following to have the done automatically:

```python
if __name__ == "__main__":
    print(DbSchema)
```

And then all one needs to do is:

```sh
python mlos_bench/storage/sql/schema.py
```

mlos_bench/mlos_bench/storage/__init__.py
bpkroth, line 1: Minor nit about the storage organization.
Assuming we do add other formats later on, separating the file structure out like so might be helpful in keeping things clean:

```
storage/
  __init__.py
  base_storage.py
  sql/
    __init__.py  # this can do the full import of all the other sql related modules at this point
    storage.py
    schema.py
    experiment.py
    (sql_)trial.py  # could take or leave the leading `sql_` prefix for all of these
  mlflow/
    __init__.py
    storage.py
    trial.py
    experiment.py
    ...
```

Again, a thing we can change later if we want.

mlos_bench/mlos_bench/tests/tunables/tunables_str_test.py
```diff
@@ -0,0 +1,58 @@
...
+from mlos_bench.tunables.tunable_groups import TunableGroups
+
+
+def test_tunable_groups_str(tunable_groups: TunableGroups) -> None:
+    """
+    Check that we produce the same string representation of TunableGroups,
+    regardless of the order in which we declare the covariant groups and
+    tunables within each covariant group.
+    """
+    # Same as `tunable_groups`, but in different order:
```
bpkroth, line 19:
```suggestion
    # Same as `tunable_groups` (defined in the `conftest.py` file, but in different order:
```

Nit: that other one is defined in the `conftest.py` file, which is possibly not obvious and harder to track down changes for.

mlos_bench/mlos_bench/tunables/tunable.py
```diff
@@ -117,6 +117,33 @@ def __eq__(self, other: object) -> bool:
             self._current_value == other._current_value
         )
 
+    def __lt__(self, other: object) -> bool:
```
bpkroth, line 120: :thumbsup:

FWIW, I did some playing around with making the `Tunables` class more type aware with Generics:
https://github.com/microsoft/MLOS/compare/main...bpkroth:MLOS:mypy-tunable-values-tweak

I think for now it's an unnecessary and overly complicated dead end, so I wasn't gonna pursue it further, but I may keep it around in case we want it in the future.

mlos_bench/mlos_bench/service/config_persistence.py
```diff
@@ -108,8 +110,83 @@ def load_config(self, json_file_name: str) -> Union[dict, List[dict]]:
...
+        for key in class_params.intersection(config.get("resolve_path", [])):
+            class_config[key] = self.resolve_path(class_config[key])
+
+        if _LOG.isEnabledFor(logging.DEBUG):
+            _LOG.debug("Instantiating: %s with config:\n%s",
+                       class_name, json.dumps(class_config, indent=2))
+
+        return (class_name, class_config)
+
+    def build_generic(self, *,
```
bpkroth, line 151: Actually, I have a way to make this typed with generics.  I'll submit a PR against yours in a sec.

bpkroth: https://github.com/motus/MLOS/pull/2

mlos_bench/mlos_bench/service/config_persistence.py
bpkroth: https://github.com/motus/MLOS/pull/2

bpkroth (approved):
Some suggested tweaks, but otherwise LGTM.  Thanks!

mlos_bench/mlos_bench/optimizer/base_optimizer.py
motus: agree. I just want to be consistent in which order we define these parameters across different classes. We can enforce that in abstract methods, but not in teh constructor, unforttunately

mlos_bench/mlos_bench/service/config_persistence.py
motus: I think that's fixed in your update

mlos_bench/mlos_bench/service/config_persistence.py
motus: That, too - I see there's a cleaner fix for this in your update

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -0,0 +1,282 @@
...
+                assert exc_type and exc_val
+                _LOG.warning("Finishing experiment: %s", self,
+                             exc_info=(exc_type, exc_val, exc_tb))
+            self._teardown(is_ok)
+            return False  # Do not suppress exceptions
+
+        def __repr__(self) -> str:
+            return self._experiment_id
+
+        def _setup(self) -> None:
```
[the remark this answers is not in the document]
motus: I removed it because not every implementation might require additional setup logic, and I wanted to provide some working default

mlos_bench/mlos_bench/storage/sql_schema.py
motus: oh wait, no - we need to instantiate an SQLAlchemy DB engine, and that, in turn, requires us to load the JSONC stirage config etc. etc. 
There might be a way, but let's do it in some other PR

mlos_bench/mlos_bench/service/config_persistence.py
motus: yep, I think it is in my branch already

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -0,0 +1,282 @@
...
+        """
+
+    class Experiment(metaclass=ABCMeta):
+        """
+        Base interface for storing the results of the experiment.
+        This class is instantiated in the `Storage.experiment()` method.
+        """
+
+        def __init__(self, tunables: TunableGroups, experiment_id: str, root_env_config: str):
+            self._tunables = tunables  # No need to copy, it's immutable
```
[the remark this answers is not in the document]
motus: I meant, the Experiment never mutates this particular object. But you're right, it's safer to make a copy here.

mlos_bench/mlos_bench/storage/__init__.py
motus: done

mlos_bench/mlos_bench/storage/sql_schema.py
bpkroth: Sounds good.  I guess I didn't look too closely, but I thought you were using some dummy engine for the schema printing, so it would have been fine.  Meh, not super important.  I can just imagine the need to debug that and easy access to the SQL text would be nice.

bpkroth:
🥳

PR #300, opened 2023-04-11
Minor tweaks from other dev work

bpkroth:
- add a common spelling word
- let "Ctrl-W" stop removing words at slashes in the devcontainer terminal
- make sure that the method vscode uses to discover tests is working


motus (approved):
looks good!

PR #303, opened 2023-04-12
Add Optional[] annotations to some mlos_core files

motus:
Few small changes to make mypy happy again (I use mypy 1.2.0)


bpkroth:
Perfect.  I actually have a larger version of this to enable more strictness as well.  I'll include this one first and then merge it into mine next.

bpkroth (approved):

PR #304, opened 2023-04-13
Minor updates

bpkroth:
- Spelling fixups
- Only tell nginx to listen on IPv6
  This actually handles both network stacks, but conflicts left with weird WSL2 networking oddities so we can run multiple devcontainers at once.


mlos_core/mlos_core/tests/spaces/adapters/llamatune_test.py
```diff
@@ -238,25 +238,25 @@ def test_special_parameter_values_biasing():    # pylint: disable=too-complex
     adapter = LlamaTuneAdapter(input_space, num_low_dims=1, special_param_values=spv_dict,
                                max_unique_values_per_param=None)
 
-    special_values_occurences = {
+    special_values_instances = {
```
bpkroth, line 241: In this case we rename the variable to also avoid a mypy warning about redefinition with different types.

bpkroth:
- Spelling fixups
- Only tell nginx to listen on IPv6
  This actually handles both network stacks, but conflicts left with weird WSL2 networking oddities so we can run multiple devcontainers at once.

motus (approved):

PR #309, opened 2023-04-24
No-op storage implementation

motus:
We'll need it for testing and one-off benchmarks. It also helps us to streamline the operation of `run_bench.py` and `run_op.py` and probably get the two converged into a single app (run_opt with null optimizer and null storage is effectively run_bench)


motus:
oh wait this is so stupid. we can simply use SQLite3 `:memory:` to the same effect

bpkroth:
> oh wait this is so stupid. we can simply use SQLite3 `:memory:` to the same effect

True.  I guess just make a special precanned config for it that we can include instead?

PR #310, opened 2023-04-24
Add a configuration for NULL storage (in-memory SQLite3).

motus:
We need it for testing and one-off benchmark runs.


mlos_bench/config/storage/null.jsonc
bpkroth, line 1: Maybe call it `in-memory` or `ephemeral` or `tmp` or some such?
`null` to me is reminiscent of `/dev/null` which isn't quite this :)

motus: I agree :) will do in a sec

bpkroth (approved):

PR #311, opened 2023-04-24
make resolve_path parameter handle lists of paths

motus:
also, add some extra type checks

### DIFF
--- a/mlos_bench/mlos_bench/service/config_persistence.py
+++ b/mlos_bench/mlos_bench/service/config_persistence.py
@@ -140,7 +140,12 @@ class ConfigPersistenceService(Service, SupportsConfigLoading):
             class_config[key] = global_config[key]
 
         for key in class_params.intersection(config.get("resolve_path", [])):
-            class_config[key] = self.resolve_path(class_config[key])
+            if isinstance(class_config[key], str):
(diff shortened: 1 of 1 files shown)


mlos_bench/mlos_bench/service/config_persistence.py
```diff
@@ -140,7 +140,12 @@ def prepare_class_load(self, config: Dict[str, Any],
             class_config[key] = global_config[key]
 
         for key in class_params.intersection(config.get("resolve_path", [])):
-            class_config[key] = self.resolve_path(class_config[key])
+            if isinstance(class_config[key], str):
+                class_config[key] = self.resolve_path(class_config[key])
+            elif isinstance(class_config[key], (list, tuple)):
```
bpkroth, line 145:
```suggestion
            elif isinstance(class_config[key], Iterable):
```

I think this should work and is more generic.

bpkroth (approved):

mlos_bench/mlos_bench/service/config_persistence.py
bpkroth: But fine as is too.

motus: nah, that will accept a `dict` but work incorrectly. I'd rather be more strict

PR #312, opened 2023-04-24
No-op optimizer that runs a single benchmark with a given config

motus:
We need it for testing and one-off benchmarks. I also want to merge `run_bench.py` and `run_opt.py` into one script (`run_bench.py` is basically a `run_opt.py` with null storage and null optimizer). Next step, I'll also move all command-line parameters into one uber-config file.


mlos_bench/mlos_bench/optimizer/null_optimizer.py
```diff
@@ -0,0 +1,39 @@
...
+
+from mlos_bench.service.base_service import Service
+from mlos_bench.service.types.config_loader_type import SupportsConfigLoading
+from mlos_bench.tunables.tunable_groups import TunableGroups
+from mlos_bench.optimizer.mock_optimizer import MockOptimizer
+
+_LOG = logging.getLogger(__name__)
+
+
+class NullOptimizer(MockOptimizer):
```
bpkroth, line 20:
```suggestion
class OneShotOptimizer(MockOptimizer):
```

mlos_bench/mlos_bench/optimizer/null_optimizer.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    Null optimizer that proposes a single configuration and converges.
+    """
+
+    def __init__(self, tunables: TunableGroups,
+                 service: Optional[Service], config: Dict[str, Any]):
+        super().__init__(tunables, service, config)
+        if self._service is not None and isinstance(self._service, SupportsConfigLoading):
```
bpkroth, line 28: I think inside `BaseService` there should be a `self._config_loader_service` instance that you can use for this.

bpkroth: It was so common that I think I added that check and attribute alias to the base class so that we didn't have to have that `isinstance` check everywhere.

bpkroth (approved):
Just the `self._config_loader_service` bit, but otherwise LGTM.  Thanks!

mlos_bench/mlos_bench/optimizer/null_optimizer.py
motus: I like it! `OneShot` it is

mlos_bench/mlos_bench/optimizer/null_optimizer.py
motus: oh, that's right. will fix it now

PR #314, opened 2023-04-25
bugfix: correct class name for OneShotOptimizer

motus:
also, make sure config loader is available even when it's the only service without a parent


bpkroth (approved):
Tests?

PR #313, opened 2023-04-24
Refactor config examples directory structure

bpkroth:
The existing configs and examples directory structure is a little bit of a mess and not currently being distributed with the `whl`.

This PR starts reorganizing it to
- match the module layout (e.g. `service` vs. `environment` vs. `optimizer` vs. `storage` base classes).
- distributes it as part of the `whl`
  - [x] make use of `importlib` to access files at runtime for the default search path:
   https://setuptools.pypa.io/en/latest/userguide/datafiles.html#accessing-data-files-at-runtime
- [x] test the renamed files for internal path fixups
  - [x] finds all "environment" jsonc files in the `config` dir and check that we can at least load them into an Environment class.
  - [x] This makes environment files self-contained so `TunableGroups` loading logic also needing fixing and tests - #327 

This should make it easier to interact with and test the pre-canned examples.

Future work:
- [-] use `jsonschema` for config validation and IDE autocomplete help

PR #317, opened 2023-04-26
Unify run_opt and run_bench before implementing an uber-config

motus:
Create a single script that can run a single trial or a full-scale optimization loop. In the next PR, I will update the `Launcher` class so it can take a single JSON config instead of a multitude of command-line options. (We'll still have the ability to use the command line to override the JSON config parameters).


Makefile
```diff
@@ -282,12 +282,9 @@ doc/source/api/mlos_bench/modules.rst: $(MLOS_BENCH_PYTHON_FILES) $(COMMON_DOC_F
 	rm -rf doc/source/api/mlos_bench
 	cd doc/ && conda run -n ${CONDA_ENV_NAME} sphinx-apidoc -f -e -M -o source/api/mlos_bench/ ../mlos_bench/ ../mlos_*/setup.py
 	# Save the help output of the mlos_bench scripts to include in the documentation.
-	conda run -n ${CONDA_ENV_NAME} mlos_bench/mlos_bench/run_opt.py --help > doc/source/api/mlos_bench/mlos_bench.run_opt.usage.txt
-	echo ".. literalinclude:: mlos_bench.run_opt.usage.txt" >> doc/source/api/mlos_bench/mlos_bench.run_opt.rst
-	echo "   :language: none" >> doc/source/api/mlos_bench/mlos_bench.run_opt.rst
-	conda run -n ${CONDA_ENV_NAME} mlos_bench/mlos_bench/run_bench.py --help > doc/source/api/mlos_bench/mlos_bench.run_bench.usage.txt
-	echo ".. literalinclude:: mlos_bench.run_bench.usage.txt" >> doc/source/api/mlos_bench/mlos_bench.run_bench.rst
-	echo "   :language: none" >> doc/source/api/mlos_bench/mlos_bench.run_bench.rst
+	conda run -n ${CONDA_ENV_NAME} mlos_bench/mlos_bench/run.py --help > doc/source/api/mlos_bench/mlos_bench.run.usage.txt
```
bpkroth, line 285: https://github.com/microsoft/MLOS/pull/313/files#diff-fe3fc947c54d13b7a130147b2331b17caad55820ffead76b85fd38dff3397208R69-R73
and
https://github.com/microsoft/MLOS/pull/313/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R291

has a changes that will distribute this script as "mlos_bench-script_name".
Do you want to slurp that change in here, or leave that for another PR?

Basically we could make it so that `mlos_bench-run` was a command on the user's `PATH` could potentially be convenient and ergonomic.

doc/source/overview.rst
```diff
@@ -194,3 +190,21 @@ Optimizer Adapters
    Optimizer
    MockOptimizer
    MlosCoreOptimizer
+
+Storage
```
bpkroth, line 194: Cool.  I guess we forgot about linking those in during #298

bpkroth (approved):
Seems fine for now.  I suspect we're going to continue to mess with the default options, but I like combining the scripts.

Have a look at the small suggestions I made.  Happy to tackle them later as well.

Makefile
motus: if you don't mind, I would do it in the next PR. This way I can continue working on my stuff without much interference

PR #318, opened 2023-04-27
Refactor for consistent use of plural submodule names

bpkroth:
Pulled out from #313


motus (approved):
Looks great! That's exactly what we should do as the first step - no more, no less

PR #319, opened 2023-04-27
Rename the conda env

bpkroth:
More cosmetic fixups: rename the conda env, which serves mlos_bench as much as mlos_core now, to just "mlos".


.devcontainer/Dockerfile
```diff
@@ -12,6 +12,7 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
         curl jq \
         vim-nox neovim python3-pynvim \
         make \
+        rename \
```
bpkroth, line 15: One other small change from #313 - adds the `rename` utility to the devcontainer.

motus (approved):

bpkroth:
Currently the container registry name is still `mloscore.azurecr.io`.  I may setup a new one and point it at that instead as well.

PR #321, opened 2023-04-28
Support defaults as first suggestion

bpkroth:
This adds support for returning the ConfigSpace defaults as the first config to try.

Splitting some work out from #313.

### DIFF
--- a/mlos_bench/config/optimizers/one_shot_opt.jsonc
+++ b/mlos_bench/config/optimizers/one_shot_opt.jsonc
@@ -6,10 +6,10 @@
         "minimize": "score",
         // Run a single benchmark with this configuration.
         // Can be overwritten from the command line.
-        "include_tunables": [
+        "include_tunable_values": [
             "tunable-values-example.json"
         ],
         // Additional values of the tunable parameters.
-        "tunables": {}
+        "tunable_values": {}
     }
 }
--- a/mlos_bench/mlos_bench/optimizers/base_optimizer.py
+++ b/mlos_bench/mlos_bench/optimizers/base_optimizer.py
@@ -10,7 +10,7 @@ and mlos_core optimizers.
 import logging
 from typing import Dict, Optional, Sequence, Tuple, Union
 from abc import ABCMeta, abstractmethod
-from distutils.util import strtobool
+from distutils.util import strtobool    # pylint: disable=deprecated-module
 
 from mlos_bench.services.base_service import Service
 from mlos_bench.environments.status import Status
@@ -19,7 +19,7 @@ from mlos_bench.tunables.tunable_groups import TunableGroups
 _LOG = logging.getLogger(__name__)
 
 
-class Optimizer(metaclass=ABCMeta):
+class Optimizer(metaclass=ABCMeta):     # pylint: disable=too-many-instance-attributes
     """
     An abstract interface between the benchmarking framework and mlos_core optimizers.
     """
--- a/mlos_bench/mlos_bench/optimizers/base_optimizer.py
+++ b/mlos_bench/mlos_bench/optimizers/base_optimizer.py
@@ -10,6 +10,7 @@ and mlos_core optimizers.
 import logging
 from typing import Dict, Optional, Sequence, Tuple, Union
 from abc import ABCMeta, abstractmethod
+from distutils.util import strtobool
 
 from mlos_bench.services.base_service import Service
 from mlos_bench.environments.status import Status
@@ -40,7 +41,8 @@ class Optimizer(metaclass=ABCMeta):
         self._tunables = tunables
         self._service = service
         self._iter = 1
-        self._max_iter = int(self._config.pop('max_iterations', 10))
+        self._use_defaults: bool = bool(strtobool(str.lower(self._config.pop('use_defaults', "True"))))
+        self._max_iter = int(self._config.pop('max_iterations', 100))
         self._opt_target: str
         _opt_target = self._config.pop('maximize', None)
         if _opt_target is None:
--- a/mlos_bench/mlos_bench/optimizers/base_optimizer.py
+++ b/mlos_bench/mlos_bench/optimizers/base_optimizer.py
@@ -42,7 +42,7 @@ class Optimizer(metaclass=ABCMeta):     # pylint: disable=too-many-instance-attr
         self._service = service
         self._iter = 1
         self._use_defaults: bool = bool(strtobool(str.lower(self._config.pop('use_defaults', "True"))))
-        self._max_iter = int(self._config.pop('max_iterations', 100))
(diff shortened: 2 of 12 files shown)


mlos_bench/config/optimizers/one_shot_opt.jsonc
```diff
@@ -6,10 +6,10 @@
         "minimize": "score",
         // Run a single benchmark with this configuration.
         // Can be overwritten from the command line.
-        "include_tunables": [
+        "include_tunable_values": [
```
bpkroth, line 9: In this case, these aren't tunables, but *values* to be assigned to tunables, so we rename the key slightly since the included json schemas are different and this makes it more clear that we're providing a mechanism for testing a particular (partial) config as well.

mlos_bench/mlos_bench/optimizers/base_optimizer.py
```diff
@@ -10,6 +10,7 @@
 import logging
 from typing import Dict, Optional, Sequence, Tuple, Union
 from abc import ABCMeta, abstractmethod
+from distutils.util import strtobool    # pylint: disable=deprecated-module
```
bpkroth, line 13: I see no mention that this particular bit is deprecated in the docs upstream and this is a handy method to use, so I've just silenced the message for now.  We should catch it in our CI pipelines if that does actually get deprecated.

https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool

mlos_bench/mlos_bench/optimizers/base_optimizer.py
```diff
@@ -40,7 +41,8 @@ def __init__(self, tunables: TunableGroups, service: Optional[Service], config:
         self._tunables = tunables
         self._service = service
         self._iter = 1
-        self._max_iter = int(self._config.pop('max_iterations', 10))
+        self._use_defaults: bool = bool(strtobool(str.lower(self._config.pop('use_defaults', "True"))))
+        self._max_iter = int(self._config.pop('max_iterations', 100))
```
bpkroth, line 45: 10 iterations is difficult to get much optimization out of, so I bumped up the default here, but we can back that out and leave it for another time if we want.

mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
```diff
@@ -35,15 +35,15 @@ def __init__(self, tunables: TunableGroups, service: Optional[Service], config:
         space = tunable_groups_to_configspace(tunables)
         _LOG.debug("ConfigSpace: %s", space)
 
-        opt_type = getattr(OptimizerType, self._config.pop('optimizer_type', 'SKOPT'))
+        opt_type = getattr(OptimizerType, self._config.pop('optimizer_type', DEFAULT_OPTIMIZER_TYPE.name))
```
bpkroth, line 38: Let `mlos_core` define what the default optimizer to use is.  Reduces our maintenance costs of having to update it in multiple places.

mlos_bench/mlos_bench/optimizers/one_shot_optimizer.py
```diff
@@ -18,20 +18,27 @@
...
+
     def __init__(self, tunables: TunableGroups,
                  service: Optional[Service], config: Dict[str, Any]):
         super().__init__(tunables, service, config)
         if self._service is None:
             _LOG.warning("Config loading service not available")
         else:
-            for data_file in config.get("include_tunables", []):
+            # Generate initial random or default suggestion based on optimizer config.
+            self._tunables = super().suggest()
```
bpkroth, line 34: Since the parent is MockOptimizer, this will call the appropriate default or random logic for us.
Then below we can overwrite those values provided from the config.

mlos_bench/mlos_bench/optimizers/base_optimizer.py
```diff
@@ -40,7 +41,8 @@ def __init__(self, tunables: TunableGroups, service: Optional[Service], config:
         self._tunables = tunables
         self._service = service
         self._iter = 1
-        self._max_iter = int(self._config.pop('max_iterations', 10))
+        self._use_defaults: bool = bool(strtobool(str.lower(self._config.pop('use_defaults', "True"))))
```
motus, line 44: It's a bit tricky. We use JSON5 parser, so the config can have `"use_defaults": true` and the value will show up as `bool` in `self._config`. OTOH, the user can override that value in the command line by writing `--use_defaults True`, and then it will be a string. We have to check for both cases here.

mlos_bench/mlos_bench/optimizers/base_optimizer.py
motus: yeah, I would probably keep some low number here, so that the user doesn't have to wonder why the program takes so long to finish. After all, I expect that `max_iterations` will be set explicitly in pretty much every config

mlos_bench/mlos_bench/optimizers/one_shot_optimizer.py
motus: makes sense!

motus (approved):
Looks good! My only nitpick is how we handle that bool value.

mlos_bench/mlos_bench/optimizers/base_optimizer.py
bpkroth: Exactly.  So in this case we convert both the string and the bool to a string, and then parse it back to a bool.  Not awesome, but more robust.

mlos_core/mlos_core/spaces/adapters/llamatune.py
```diff
@@ -100,7 +100,10 @@ def inverse_transform(self, configurations: pd.DataFrame) -> pd.DataFrame:
             # respective high-dim point; this way we can retrieve the low-dim point, from its high-dim counterpart.
             if target_config is None:
                 # Inherently it is not supported to register points, which were not suggested by the optimizer.
-                if not self._use_approximate_reverse_mapping:
+                if configuration == self.orig_parameter_space.get_default_configuration():
+                    # Default configuration should always be registerable.
```
bpkroth, line 104: This addresses the issue of us proposing the default first without really consulting the optimizer.
One could potentially argue that we should set `use_approximate_reverse_mapping` to `True` by default, but I think this was a smaller scoped change and sane to do for now.

bpkroth: @kkanellis in case he wants to comment further

mlos_bench/mlos_bench/optimizers/base_optimizer.py
bpkroth: Agreed.  I set it to 25 for now.  10 is barely into the random iterations.

PR #322, opened 2023-04-28
Circular import cleanup

bpkroth:
Some circular import cleanup in preparation for config reorgs and importlib based loading from #313

### DIFF
--- a/mlos_bench/mlos_bench/environments/base_environment.py
+++ b/mlos_bench/mlos_bench/environments/base_environment.py
@@ -9,14 +9,16 @@ A hierarchy of benchmark environments.
 import abc
 import json
 import logging
-from typing import Dict, Optional, Tuple
+from typing import Dict, Optional, Tuple, TYPE_CHECKING
 
 from mlos_bench.environments.status import Status
 from mlos_bench.services.base_service import Service
-from mlos_bench.services.types.config_loader_type import SupportsConfigLoading
 from mlos_bench.tunables.tunable import TunableValue
 from mlos_bench.tunables.tunable_groups import TunableGroups
 from mlos_bench.util import instantiate_from_config
+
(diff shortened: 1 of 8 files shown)


mlos_bench/mlos_bench/environments/base_environment.py
```diff
@@ -140,6 +138,11 @@ def __init__(self,
             _LOG.debug("Config for: %s\n%s",
                        name, json.dumps(self.config, indent=2))
 
+    @property
+    def _config_loader_service(self) -> "SupportsConfigLoading":
+        assert self._service is not None
+        return self._service.config_loader_service
```
bpkroth, line 144: Slight tweak: avoids the need to store an extra reference.

mlos_bench/mlos_bench/environments/composite_env.py
```diff
@@ -70,7 +70,7 @@ def _add_child(self, env: Environment) -> None:
         This method is called from the constructor only.
         """
         self._children.append(env)
-        self._tunable_params.update(env.tunable_params())
+        self._tunable_params.update(env.tunable_params)
```
bpkroth, line 73: Cosmetic fixup

bpkroth:
Seems to be a flaky test (`test_optimizer_with_llamatune`), but nothing I can see that changed that should affect that and is running fine in a loop on my local machine.

motus:
> Seems to be a flaky test (`test_optimizer_with_llamatune`), but nothing I can see that changed that should affect that and is running fine in a loop on my local machine.

yes, I've seen it once or twice before - an optimizer recommends a value that falls out of the expected range. Is that the one? Anyway, I think I'm fine with restarting a CI job for now; we'll debug the issue later

motus (approved):
Looks good!

PR #323, opened 2023-04-28
Support top-level config file for mlos_bench.run

motus:


azure-redis-1shot.jsonc
```diff
@@ -0,0 +1,26 @@
+{
+    "config_path": [
+        "./mlos_bench/config",
```
bpkroth, line 3: Been having to hack at these embedded paths quite a bit and there's a bunch of problems with them.

```suggestion
        "mlos_bench/config",
```

I think we should omit the leading `./` when we want the search path resolution logic to take effect and use `./` to be interpreted as either relative to the cwd where the script was executed from or relative to this file, though the latter requires additional complicated logic that I'm not sure will be great to embed in the ConfigPersistence module directly since it may need to be specialized per key.  For instance, how we refer to scripts in "environment" configs: right now they're specified as a string including "a/path/to/script plus args".  That works for shell environments, but doesn't let us identify anything about "a/path/to/script" for early validation.  Supporting alternative syntaxes like `devcontainer.json` does for array based methods like `["a/path/to/script", "plus", "args"]` could also be helpful.

Anyways, much of that discussion is not relevant here, but I figure worth bringing up for future things.


What I'd really like to do is add support for a json schemas on these files so that we can have more clearly defined semantics for expected fields in different places and easy documentation and validation, but that's also future work.

mlos_bench/README.md
```diff
@@ -100,43 +100,80 @@ Create and activate the environment with:
     conda activate mlos
     ```
 
-7. Run our configuration through `mlos_bench`.
+7. Connect the Environment, Storage, and Optimizer configurations in a top-level JSON5 configuration - say, "config_bench.jsonc":
```
bpkroth, line 103:
```suggestion
7. Connect the Environment, Service, Storage, and Optimizer configurations in a top-level [JSON5](https://spec.json5.org/) configuration - say, `config_bench.jsonc`:
  > See [example-file-name.jsonc](./path/to/example-file-name.jsonc) for a documented example.
```

bpkroth (changes requested):
Can we add a simple test that also runs the cli version so that we're checking (at least a little) what we're documenting?

azure-redis-1shot.jsonc
```diff
@@ -0,0 +1,26 @@
...
+    "tunable_values": [
+        "tunable-values-example.json"
+    ],
+
+    "globals": [
+        "global_config.json"
+    ],
+
+    "experimentId": "RedisBench",
+    "trialId": 1,
```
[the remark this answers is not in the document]
motus: absolutely, but in the next PR - this change will trigger a huge cascade of updates to all scripts that you're currently refactoring (we have `./$experimentId/$trialId/...` everywhere)

azure-redis-1shot.jsonc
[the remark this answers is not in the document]
motus: agree

azure-redis-1shot.jsonc
motus: agree. I'll remove the leading `./` now, and later we can implement all other stuff

mlos_bench/README.md
motus: very good point! I'll add it

mlos_bench/README.md
```diff
@@ -100,43 +100,80 @@ Create and activate the environment with:
...
+
+    "experimentId": "RedisBench",                    // Experiment ID (can be in global_config.json)
+    "trialId": 1                                     // Trial ID (can come from the persistent storage service)
+}
 ```
 
-The only difference between the two invocations is that the latter has the `--optimizer` parameter instead of using a fixed set of tunables via `--tunables` option.
-It also stores the results of each trial in the DuckDB database (configured via `--storage`).
+The only difference between the two configurations is that the latter has the `optimizer` parameter instead of using a fixed set of tunables via `tunable_values` option.
+It also stores the results of each trial in the DuckDB database (configured via `storage`).
```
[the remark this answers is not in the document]
motus: good question. DuckDB has stricter type checks and I think better in general; OTOH, SQLite is still more popular and I think it comes with miniconda by default (or gets pulled in by one of our existing dependencies). Also, we already use SQLite for our unit tests. Ok, I've convinced myself. Let's use SQLite by default 😄

motus (commented):
I hope that triggers a commit with all code suggestions

motus:
> Can we add a simple test that also runs the cli version so that we're checking (at least a little) what we're documenting?

Done!

mlos_bench/mlos_bench/tests/optimizers/toy_optimization_loop_test.py
```diff
@@ -58,7 +58,7 @@ def test_mock_optimization_loop(mock_env_no_noise: MockEnv,
 
 
 def test_mock_optimization_loop_no_defaults(mock_env_no_noise: MockEnv,
-                                mock_opt: MockOptimizer) -> None:
+                                            mock_opt: MockOptimizer) -> None:
```
motus, line 61: flake8 barfs

bpkroth (approved):

mlos_bench/README.md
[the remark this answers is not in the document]
bpkroth: sqlite actually ships by default with python, so just from a reduced dependency standpoint I think this makes sense.

PR #324, opened 2023-05-02
Rework a non-boot param example to a boot param

bpkroth:
Splitting out some smaller work from #313 hacking


motus (approved):
awesome, thank you!

PR #327, opened 2023-05-02
Fixes for TunableGroups merging logic

bpkroth:
Splitting more work off of #313.

In that PR we'd like to be able to make each Environment independently loadable.
To do this we need to be able to allow them to reference their Tunables.
However, other related Environments (e.g. local vs. remote), may also need to reference those same Tunables.
Moreover, we want both to obtain the same value.
To do this we would like the parent TunableGroups for the CompositeEnvironment for an Experiment to handout *references* rather than *copies* of Tunables to the child Environments.

This work adds that logic and some basic tests for it.

In essence, we make passing TunableGroups to an Environment expected, even if the first copy may be an empty collection.  We consider this one the "root" or "parent" copy.
After that, any additional Tunables loaded by subsequent Environments get `merge`d (not `update`d, so no overwriting) into the "parent" and the child gets a "view" of a `subgroup` of those.
In this way, when the Optimizer assigns new values to a Tunable, it is reflected at all levels.

Caveats:
- Currently, the `merge` logic throws an error if any overlapping CovariantTunableGroups are non-equal in any way (e.g. extra param in one vs. another).

Leaving these for future work:
- [ ] Add tests for nested CompositeEnvironments to ensure that we pass the "parent" to multiple levels deep, not just the first one.


mlos_bench/mlos_bench/environments/base_environment.py
```diff
@@ -127,12 +127,11 @@ def __init__(self,
                 raise ValueError("Missing required parameter: " + key)
 
         if tunables is None:
+            _LOG.warning("No tunables provided for %s. Tunable inheritance across composite environments may be broken.", name)
```
bpkroth, line 130: Went with this approach to keep the changes somewhat smaller for now.

mlos_bench/mlos_bench/services/config_persistence.py
```diff
@@ -52,11 +54,9 @@ def __init__(self, config: Optional[Dict[str, Any]] = None,
             self.resolve_path,
             self.load_config,
             self.prepare_class_load,
-            self.build_tunables,
```
bpkroth, line 55: Was unreferenced, so I removed it.

mlos_bench/mlos_bench/services/config_persistence.py
```diff
@@ -52,11 +54,9 @@ def __init__(self, config: Optional[Dict[str, Any]] = None,
...
             self.build_service,
             self.build_environment,
             self.load_services,
-            self.load_tunables,
```
bpkroth, line 59: Only used here, so doesn't need to be exported.

mlos_bench/mlos_bench/services/config_persistence.py
```diff
@@ -192,8 +192,8 @@ def build_generic(self, *,
         return inst
 
     def build_environment(self, config: Dict[str, Any],
+                          tunables: TunableGroups,
```
bpkroth, line 195: Make tunables a non-optional param.

bpkroth: Could alternative use the same `*` trick we did earlier to make them all required to be named, but this was fewer changes for now.

mlos_bench/mlos_bench/services/config_persistence.py
```diff
@@ -454,29 +425,31 @@ def load_services(self, json_file_names: Iterable[str],
...
-        groups = TunableGroups()
-        if parent is not None:
-            groups.update(parent)
         for fname in json_file_names:
             config = self.load_config(fname)
             assert isinstance(config, dict)
-            groups.update(TunableGroups(config))
-        return groups
+            parent.merge(TunableGroups(config))
+        return parent
```
bpkroth, line 455: This is part of the key change: we put the child tunables into the parent and return the parent.
The child tunables are then ephemeral and only used to create the object when loading the config from disk.

mlos_bench/mlos_bench/tunables/tunable_groups.py
```diff
@@ -90,8 +103,15 @@ def update(self, tunables: "TunableGroups") -> "TunableGroups":
...
         # pylint: disable=protected-access
-        self._index.update(tunables._index)
-        self._tunable_groups.update(tunables._tunable_groups)
+        # Check that covariant groups are unique, else throw an error.
+        for group in tunables._tunable_groups.values():
+            if group.name not in self._tunable_groups:
+                self._add_group(group)
+            else:
+                # Check that there's no overlap in the tunables.
+                assert self._tunable_groups[group.name] == group, \
```
bpkroth, line 112: - [ ] Change to a `ValueError`?

mlos_bench/mlos_bench/tunables/tunable_groups.py
```diff
@@ -68,17 +68,30 @@ def _add_group(self, group: CovariantTunableGroup) -> None:
...
+
         Parameters
         ----------
             group : CovariantTunableGroup
         """
+        assert group.name not in self._tunable_groups, f"Duplicate covariant tunable group name {group.name} in {self}"
         self._tunable_groups[group.name] = group
-        self._index.update(dict.fromkeys(group.get_names(), group))
+        for tunable in group.get_tunables():
+            assert tunable.name not in self._index, f"Duplicate Tunable {tunable.name} from group {group.name} in {self}"
```
bpkroth, line 80: - [ ] Add tests for these duplicate errors

mlos_bench/mlos_bench/tests/conftest.py
```diff
@@ -6,18 +6,77 @@
 Common fixtures for mock TunableGroups and Environment objects.
 """
 
+from typing import Any, Dict
+
+import json
```
motus, line 11: json5, maybe? that's what we use to load the configs

mlos_bench/mlos_bench/launcher.py
```diff
@@ -78,8 +79,9 @@ def __init__(self, description: str):
         assert env_path, "Environment configuration file is required"
         self.root_env_config = self._config_loader.resolve_path(env_path)
 
+        tunable_groups = TunableGroups()    # base tunable groups that all others get build on
```
motus, line 82: ❤️

mlos_bench/mlos_bench/tests/conftest.py
bpkroth: Done.

motus (approved):

PR #338, opened 2023-05-04
various cosmetic fixes to please pylint and flake8; shorter version of mock service implementations

motus:


mlos_bench/mlos_bench/services/config_persistence.py
```diff
@@ -9,7 +9,6 @@
 """
 
 import os
-import re
```
bpkroth, line 12: Thank you.  Missed that during PR cleanup in #313 and forgot to get back to it :)

mlos_bench/mlos_bench/services/local/temp_dir_context.py
```diff
@@ -0,0 +1,57 @@
...
+from contextlib import nullcontext
+from tempfile import TemporaryDirectory
+from typing import Optional, Union
+
+from mlos_bench.services.base_service import Service
+
+_LOG = logging.getLogger(__name__)
+
+
+class TempDirContextService(Service):
```
bpkroth, line 19: Might make sense to split out the `temp_dir_context` method from the `SupportsLocalExec` into `SupportsLocalTempDir` so that this can declare that it implements that and we can use it independently.  Otherwise, it's not entirely clear that this needs to be combined with other classes through inheritance in order to the full setup.

mlos_bench/mlos_bench/tunables/covariant_group.py
```diff
@@ -107,10 +107,13 @@ def equals_defaults(self, other: "CovariantTunableGroup") -> bool:
...
         ----------
         other : CovariantTunableGroup
+            A covariant tunable group object to compare to.
 
         Returns
         -------
-        bool
+        are_equal : bool
+            True if the two CovariantTunableGroup objects' *metadata* are the same,
+            False optherwise.
```
bpkroth, line 116:
```suggestion
            False otherwise.
```

bpkroth (approved):
A few comments and spelling inline, but otherwise LGTM

mlos_bench/mlos_bench/services/local/temp_dir_context.py
motus: Makes sense! This change was mostly to make pylint shut up about code duplication, but you're right, we should add it to the protocol

motus: mm, wait. local exec *always* depends on temp dir, and we never use temp dir outside of the local exec context, so we probably don't have to expose it to the user as an individual service. I'll make it an abstract class to make sure we never use it outside of local exec context, and write about it in the docstring. Let's talk and see if there is a more elegant solution to it, and do it in the subsequent PR.

PR #339, opened 2023-05-05
fix typos in config examples

motus:


bpkroth (approved):
LGTM

What would it take to get the cli example running in the CI pipeline?

motus:
> What would it take to get the cli example running in the CI pipeline?

Great question! Depends on what CLI example we want to run. If it does not require provisioning an extra VM - say, we want to run a mock environment for a few iterations of the optimizer, it should be pretty easy. Provisioning a VM requires security credentials to 1) deploy and shutdown a VM, and 2) to access the shared file store. Currently we have to generate that data with a script every hour or so and store it in the `global_config.json` file. We have to figure out how to use the Azure key vault or something like it to make the security credentials available in the CI environment. I have no experience with that stuff yet.

Issue #294, opened 2023-04-04
Add SMAC support to mlos_core


bpkroth:

bpkroth:
@kkanellis

PR #341, opened 2023-05-08
mlos_core: add SMAC optimizer

kkanellis:
This PR addresses #294 and adds supports for [SMAC](https://automl.github.io/SMAC3/main/) optimizer in MLOS.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -90,11 +90,6 @@ jobs:
         if (! (conda env list | Select-String -Pattern "^$env:CONDA_ENV_NAME ") ) { conda env create -v -n $env:CONDA_ENV_NAME -f conda-envs/$env:CONDA_ENV_YML }
         conda env update -v -n $env:CONDA_ENV_NAME -f conda-envs/$env:CONDA_ENV_YML --prune
 
-    # For debugging version mismatch issues:
-    - name: List the current conda packages
-      run: |
-        conda list -n $env:CONDA_ENV_NAME --show-channel-urls
-
     # This is moreso about code cleanliness, which is a dev thing, not a
     # functionality thing, and the rules for that change between python versions,
     # so only do this for the default in the devcontainer.


mlos_core/mlos_core/optimizers/bayesian_optimizers.py
```diff
@@ -39,7 +40,7 @@ def surrogate_predict(self, configurations: pd.DataFrame,
 
     @abstractmethod
     def acquisition_function(self, configurations: pd.DataFrame,
-                             context: Optional[pd.DataFrame] = None) -> Callable:
+                             context: Optional[pd.DataFrame] = None) -> npt.NDArray:
```
bpkroth, line 43: Hmm, maybe we should rename this to `aquire` instead of `aquisition_function` since it invokes it and doesn't just return the function.

mlos_core/mlos_core/optimizers/bayesian_optimizers.py
```diff
@@ -276,6 +277,177 @@ def surrogate_predict(self, configurations: pd.DataFrame, context: Optional[pd.D
         return ret
 
     def acquisition_function(self, configurations: pd.DataFrame,
-                             context: Optional[pd.DataFrame] = None) -> Callable:
+                             context: Optional[pd.DataFrame] = None) -> npt.NDArray:
         # This seems actually non-trivial to get out of skopt, so maybe we actually shouldn't implement this.
         raise NotImplementedError()
+
+
+class SmacOptimizer(BaseBayesianOptimizer):
```
bpkroth, line 285: Cosmetic, but for maintenance, I'm thinking we should split this file out to have each class in its own file first.  We should tackle that in a separate PR though.

bpkroth: Maybe do that first ...

mlos_core/mlos_core/tests/optimizers/bayesian_optimizers_test.py
```diff
@@ -41,6 +41,7 @@ def test_context_not_implemented_error(configuration_space: CS.ConfigurationSpac
         with pytest.raises(NotImplementedError):
             optimizer.surrogate_predict(suggestion, context=pd.DataFrame([["something"]]))
 
-        # acquisition function not implemented
-        with pytest.raises(NotImplementedError):
-            optimizer.acquisition_function(suggestion)
+        if optimizer_class != OptimizerType.SMAC.value: # Currently, only SMAC implements this
```
bpkroth, line 44: To prevent us from automatically excluding future optimizers that get added (e.g. flaml), can we instead exclude the two other ones that currently lack this feature?

mlos_core/setup.py
```diff
@@ -28,6 +28,7 @@
 extra_requires: Dict[str, List[str]] = {
     'emukit': ['emukit'],
     'skopt': ['scikit-optimize<=0.9.0'],  # FIXME: temporarily work around some version mismatch issues (PR 850)
+    'smac': ['smac>=2.0.0'],  # NOTE: Major refactoring on SMAC starting from v2.0.0
```
bpkroth, line 31: Need to add that to `mlos_bench/setup.py` as well.

mlos_core/mlos_core/tests/optimizers/bayesian_optimizers_test.py
bpkroth, line 1: Interesting that this is the only test.  Are there others needed?  I haven't read through all of the implementation code yet ...

We'll certainly want tests on the `mlos_bench` side.

mlos_core/mlos_core/optimizers/bayesian_optimizers.py
bpkroth: See: #342

mlos_core/mlos_core/optimizers/bayesian_optimizers.py
```diff
@@ -276,6 +277,177 @@ def surrogate_predict(self, configurations: pd.DataFrame, context: Optional[pd.D
...
+        raise RuntimeError('This function should never be called.')
+
+    def _register(self, configurations: pd.DataFrame, scores: pd.Series, context: Optional[pd.DataFrame] = None) -> None:
+        """Registers the given configurations and scores.
+
+        Parameters
+        ----------
+        configurations : pd.DataFrame
+            Dataframe of configurations / parameters. The columns are parameter names and the rows are the configurations.
+
```
motus, line 367: remove blank lines between the parameters in docstrings. they make flake8 very unhappy.
(also, maybe, run `flake8` in addition to `mypy` and `pylint --disable="fixme,consider-alternative-union-syntax"` for the entire project?)

mlos_core/mlos_core/optimizers/bayesian_optimizers.py
motus: @bpkroth I agree, but OTOH `acquisition_function` is such a common term in Bayesian optimization that it may make sense to keep the original name. At any rate, I would not do it in this PR

mlos_core/mlos_core/optimizers/bayesian_optimizers.py
```diff
@@ -276,6 +277,177 @@ def surrogate_predict(self, configurations: pd.DataFrame, context: Optional[pd.D
...
+        seed: Optional[int] = None,
+        run_name: str = 'smac',
+        output_directory: str = 'smac3_output',
+        n_random_init: Optional[int] = 10,
+        n_random_probability: Optional[float] = 0.1,
+        n_workers: int = 1,
+    ):
+        super().__init__(parameter_space, space_adapter)
+
+        from smac import HyperparameterOptimizationFacade as Optimizer_Smac # pylint: disable=import-outside-toplevel
```
motus, line 306: Just write a pylint comment once on a preceding line. Also, make sure you have at least TWO spaces before each in-line comment (flake8 checks that)

mlos_core/mlos_core/optimizers/bayesian_optimizers.py
```diff
@@ -276,6 +277,177 @@ def surrogate_predict(self, configurations: pd.DataFrame, context: Optional[pd.D
...
+        from smac import Scenario # pylint: disable=import-outside-toplevel
+        from smac.intensifier.abstract_intensifier import AbstractIntensifier # pylint: disable=import-outside-toplevel
+        from smac.initial_design import LatinHypercubeInitialDesign # pylint: disable=import-outside-toplevel
+        from smac.main.config_selector import ConfigSelector # pylint: disable=import-outside-toplevel
+        from smac.random_design.probability_design import ProbabilityRandomDesign # pylint: disable=import-outside-toplevel
+        from smac.runhistory import TrialInfo # pylint: disable=import-outside-toplevel
+
+        self.seed: int = seed or 0
```
motus, line 314: why not have `seed: int = 0` in `__init__` parameters?

mlos_core/mlos_core/optimizers/bayesian_optimizers.py
```diff
@@ -276,6 +277,177 @@ def surrogate_predict(self, configurations: pd.DataFrame, context: Optional[pd.D
...
+class SmacOptimizer(BaseBayesianOptimizer):
+    """ Wrapper class for SMAC based Bayesian optimization.
+
+    Parameters
+    ----------
+    parameter_space : ConfigSpace.ConfigurationSpace
+        The parameter space to optimize.
+    """
+    def __init__(
+        self,
```
motus, line 294: maybe consider `def __init__(self, *, parameter_space, ...)` to enforce named parameters? (not sure how it will work with the Factory, though)

mlos_core/mlos_core/tests/optimizers/bayesian_optimizers_test.py
motus: +1 for more unit tests. Ideally, I would love to have unit tests and configs in `mlos_bench`, too. (See how we do it for EMUKIT and SKOPT there)

motus (commented):
Left a few nit-picking comments. My main requests are:
1. Run `flake8` in addition to `mypy` and `pylint` and fix the issues, if there are any.
2. Add a configuration and unit tests to `mlos_bench` to make sure SMAC works with our framework (please ping @motus or @bpkroth if you have questions).

mlos_core/mlos_core/optimizers/bayesian_optimizers.py
motus: actually, no - keep `seed: Optional[int] = None` - but maybe do NOT make the seed 0 by default - this effectively makes our optimizer deterministic even if we do not specify the seed (if I understand it correctly).

kkanellis: Actually, SMAC does not accept a `None` value for its `seed` argument -- it is set to `0` by default and can only take int values.

I think there is value to having a `None` value: e.g., we can set the `seed` to -1, which results in SMAC generating a random seed internally.

bpkroth:
Should also make sure it's listed in `doc/overview.rst`

mlos_core/mlos_core/optimizers/bayesian_optimizers.py
kkanellis: Maybe open a separate PR for this later?

mlos_core/mlos_core/optimizers/bayesian_optimizers.py
bpkroth: #342 should go in shortly.  Please rebase off of that.  Thanks!

mlos_core/mlos_core/optimizers/bayesian_optimizers.py
kkanellis: Any comments on the above? @motus @bpkroth

mlos_core/mlos_core/tests/optimizers/bayesian_optimizers_test.py
kkanellis: Let's open another issue for this, so that we discuss what kind of tests we need. Happy to contribute to this.

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -0,0 +1,212 @@
...
+        # pylint: disable=import-outside-toplevel
+        from smac import HyperparameterOptimizationFacade as Optimizer_Smac
+        from smac import Scenario
+        from smac.intensifier.abstract_intensifier import AbstractIntensifier
+        from smac.initial_design import LatinHypercubeInitialDesign
+        from smac.main.config_selector import ConfigSelector
+        from smac.random_design.probability_design import ProbabilityRandomDesign
+        from smac.runhistory import TrialInfo
+
+        self.trial_info_map: Mapping[ConfigSpace.Configuration, TrialInfo] = {}  # Stores TrialInfo instances returned by .ask()
```
bpkroth, line 73:
```suggestion
        # Store for TrialInfo instances returned by .ask()
        self.trial_info_map: Mapping[ConfigSpace.Configuration, TrialInfo] = {}
```

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -0,0 +1,212 @@
...
+
+        # Instantiate Scenario
+        output_directory = Path(output_directory)
```
bpkroth, line 76: Is an output directory needed?
If not, we may want to default to not-storing.

bpkroth: Alt, when this gets instantiated via `mlos_bench` we could consider to use the `tempdir_context` to store the intermediaries in a tempdir that gets automatically cleaned up.

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -0,0 +1,212 @@
...
+        scenario: Scenario = Scenario(
+            self.optimizer_parameter_space,
+            name=run_name,
+            output_directory=output_directory,
+            deterministic=True,
+            n_trials=1e4,
```
bpkroth, line 82: This shouldn't be hard coded, correct?

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -0,0 +1,212 @@
...
+        The parameter space to optimize.
+
+    space_adapter : BaseSpaceAdapter
+        The space adapter class to employ for parameter space transformations.
+
+    seed : Optional[int]
+        Seed is used to make results reproducible. A value of `None` will use a random seed.
+
+    run_name : str
+        Name of this run. This is used to easily distinguish across different runs.
```
bpkroth, line 35: Should pass that from the `mlos_bench`'s `experiment_id`, no?

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -0,0 +1,212 @@
...
+
+    n_random_init : Optional[int]
+        Number of points evaluated at start to bootstrap the optimizer.
+
+    n_random_probability: Optional[float]
+        Probability of choosing to evaluate a random configuration during optimization.
+        Setting this to a higher value favors exploration over exploitation.
+
+    n_workers: int
+        Number of parallel workers to use.
```
bpkroth, line 48: Is this for support of simultaneous experiment trial runs, or for parallel threads during model operations?

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -0,0 +1,212 @@
...
+    """
+
+    def __init__(  # pylint: disable=too-many-arguments,too-many-locals
+        self,
+        parameter_space: ConfigSpace.ConfigurationSpace,
+        space_adapter: Optional[BaseSpaceAdapter] = None,
+        seed: Optional[int] = None,
+        run_name: str = 'smac',
```
bpkroth, line 56:
```suggestion
        run_name: str = 'mlos_core-smac',
```

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -0,0 +1,212 @@
...
+        output_directory: str = 'smac3_output',
```
bpkroth, line 57: Out of curiosity, where does SMAC3 come from?  Isn't this version 2?

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -0,0 +1,212 @@
+#
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+#
+"""
+Contains the wrapper class for SMAC Bayesian optimizers.
```
bpkroth, line 6:
```suggestion
Contains the wrapper class for SMAC Bayesian optimizers.
See Also: https://automl.github.io/SMAC3/main/index.html
```

bpkroth: Should hopefully provide a quick link to it when the docs are generated from the docstrings.

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -0,0 +1,212 @@
...
+        )
+
+    @staticmethod
+    def _dummy_target_func(seed):
+        """Dummy target function for SMAC optimizer.
+
+        Since we only use the ask-and-tell interface, this is never called.
+        """
+        # NOTE: Providing a target function when using the ask-and-tell interface is an imperfection of the API
+        # -- this planned to be fixed in some future release: https://github.com/automl/SMAC3/issues/946
```
bpkroth, line 115: 👍

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -0,0 +1,212 @@
...
+            raise NotImplementedError()
+
+        if len(self._observations) < self.base_optimizer._initial_design._n_configs:  # pylint: disable=protected-access
+            raise RuntimeError('Surrogate model can make predictions *only* after all initial points have been evaluated')
+
+        configs: npt.NDArray = convert_configurations_to_array(self._to_configspace_configs(configurations))
+        mean_predictions, _ = self.base_optimizer._config_selector._model.predict(configs)  # pylint: disable=protected-access
+        return mean_predictions.reshape(-1,)
+
+    def acquisition_function(self, configurations: pd.DataFrame, context: Optional[pd.DataFrame] = None) -> npt.NDArray:
```
bpkroth, line 187: Reposting comment from prior rendition that seems to have gotten lost:

```suggestion
    def invoke_acquisition_function(self, configurations: pd.DataFrame, context: Optional[pd.DataFrame] = None) -> npt.NDArray:
```

or

```suggestion
    def aquire(self, configurations: pd.DataFrame, context: Optional[pd.DataFrame] = None) -> npt.NDArray:
```

Thoughts?  @motus

mlos_core/mlos_core/optimizers/bayesian_optimizers.py
bpkroth: I don't think that will work out well with the factory method.  I think it's ok to leave it as is for now.  We could tackle changing that behavior in a separate PR.

mlos_core/mlos_core/optimizers/bayesian_optimizers.py
bpkroth: I agree with @motus, I think a random seed is probably a better default.

I'm not sure I see a point in accepting `None` to the argument though.

I think

```python
    seed: int = -1,
```

is probably fine as the argument to `__init__` and let SMAC throw an error if `None` is passed for some reason.

bpkroth:
> Left a few nit-picking comments. My main requests are:
> 
> 1. Run `flake8` in addition to `mypy` and `pylint` and fix the issues, if there are any.
> 2. Add a configuration and unit tests to `mlos_bench` to make sure SMAC works with our framework (please ping @motus or @bpkroth if you have questions).

Several of my comments are also about `mlos_bench` integration.

In the interest of getting something in, I think I'd be okay to split this so that we have `mlos_core` support get added in one PR and `mlos_bench` integration and testing added in another PR.

mlos_bench/setup.py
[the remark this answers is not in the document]
kkanellis: Hmm, I haven't looked at the `mlos_bench` codebase. Do you some pointers on this?

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
bpkroth: We could also split this out to a separate PR.

mlos_core/mlos_core/optimizers/bayesian_optimizers.py
kkanellis: So the default for SMAC is `0` to make experiments reproducible. `None` will throw an error.

FYI https://automl.github.io/SMAC3/main/api/smac.scenario.html#smac.scenario.Scenario

mlos_bench/setup.py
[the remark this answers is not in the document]
bpkroth: I'll leave some comments in #348 on where to look for that.

mlos_core/mlos_core/optimizers/bayesian_optimizers.py
bpkroth: Eh, I'm fine with either I guess.

Let's do this:

```python
   ...
   seed: Optional[int] = 0,
   ...

   self.seed = -1 if seed is None else seed
```

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
kkanellis: Unfortuantely, SMAC requires this to be valid. I'll use Python's `TemporaryDirectory` instance for now. Will switch to `tempdir_context` when integrating to `mlos_bench`.

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
kkanellis: This is from SMAC documentation:

> The maximum number of trials (combination of configuration, seed, budget, and instance, depending on the task)

Initially I was thinking about giving this a large number that we'll never reach. A better solution would probably be to expose this as a `max_trials` parameter in constructor.

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
kkanellis: I assumed it was the latter, but turns out it's the former (i.e., from https://automl.github.io/SMAC3/main/advanced_usage/9_parallelism.html)

So, I guess we do not want to expose this as argument to the user and set it `1`, correct?

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -0,0 +1,212 @@
...
+        Seed is used to make results reproducible. A value of `None` will use a random seed.
+
+    run_name : str
+        Name of this run. This is used to easily distinguish across different runs.
+
+    output_directory : str
+        The directory where SMAC output will saved.
+
+    n_random_init : Optional[int]
+        Number of points evaluated at start to bootstrap the optimizer.
```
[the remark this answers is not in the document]
kkanellis: If this is set to `None`, then SMAC automatically sets this based on `n_trials` value. I guess that works for us.

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
kkanellis: SMAC3 is the name of the implementation (I guess this is because it is based upon Python3, but could be wrong). Yet, the Python package itself is named `smac`. 

Here we use `smac` version 2.0, as this is the first release that implements the ask-and-tell interface.

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
motus: I like renaming, but this definitely should be a separate PR. Also, `s/aquire/acquire/`

motus:
> In the interest of getting something in, I think I'd be okay to split this so that we have mlos_core support get added
> in one PR and mlos_bench integration and testing added in another PR.

sounds good. by the way, is there something like `pip install types-smac` for mypy? if yes, we should add it to our package requirements

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -0,0 +1,253 @@
...
+    n_random_init : Optional[int]
+        Number of points evaluated at start to bootstrap the optimizer. Defaults to 10.
+
+    n_random_probability: Optional[float]
+        Probability of choosing to evaluate a random configuration during optimization.
+        Defaults to `0.1`. Setting this to a higher value favors exploration over exploitation.
+    """
+
+    def __init__(  # pylint: disable=too-many-arguments,too-many-locals
+        self,
```
motus, line 57: use `self, *, ...` to enforce named parameters and get rid of `too-many-arguments` issue

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -0,0 +1,253 @@
...
+        self.temp_output_directory: Optional[TemporaryDirectory] = None
+        if output_directory is None:
+            # pylint: disable=consider-using-with
+            try:
+                self.temp_output_directory = TemporaryDirectory(ignore_cleanup_errors=True)  # Argument added in Python 3.10
+            except TypeError:
+                self.temp_output_directory = TemporaryDirectory()
+            output_directory = self.temp_output_directory.name
+
+        # Instantiate Scenario
```
motus, line 96: a perfectly useless comment 😄

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -0,0 +1,253 @@
...
+
+        # Customize SMAC's randomized behavior
+        initial_design: Optional[LatinHypercubeInitialDesign] = None
+        if n_random_init is not None:
+            initial_design = LatinHypercubeInitialDesign(scenario=scenario, n_configs=n_random_init)
+        random_design: Optional[ProbabilityRandomDesign] = None
+        if n_random_probability is not None:
+            random_design = ProbabilityRandomDesign(probability=n_random_probability)
+
+        # Create SMAC optimizer
```
motus, line 117: I would rather remove this and other similar comments that just state the obvious

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -0,0 +1,253 @@
...
+            scenario,
+            SmacOptimizer._dummy_target_func,
+            initial_design=initial_design,
+            intensifier=intensifier,
+            random_design=random_design,
+            config_selector=config_selector,
+            overwrite=True,
+        )
+
+    def __del__(self) -> None:
```
motus, line 128: I am not sure it's a good idea to rely on `__del__()` to clean up the resources (Python does not guarantee it will be invoked at all). A much better solution would be to use `__enter__()` / `__exit__()` methods to create a `with` context (e.g., please see our own `mlos_bench.storage.base_storage.Storage.Experiment`) 

I understand that this might require changes to the overall optimzier API, and I am ok if we do this in a separate PR - but for now let's create an explicit `.close()` method and call it and also open an issue on github so we don't forget to implement a `with` context later

motus (commented):
Left a few comments. My main grief is that SMAC does not seem to work on Windows (and I was not able to figure out how to install it on my PC). If we can't resolve that issue, at least we have to make this dependency conditional and do the same with the unit tests for SMAC

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
kkanellis: May I ask why do we need to enforce the named parameters?

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
kkanellis: Hmm I wasn't aware that it was possible for `__del__()` to not be called -- nice catch!

I agree that adding a `.close()` method would be the proper way to do it, since relying on a context manager for the optimizer object can be restrictive for the user. cc'ing @bpkroth

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
bpkroth: We discussed this offline.  Let's make a separate PR for that.

bpkroth:
@kkanellis maybe try this for fixing the Windows build issues:
https://github.com/kkanellis/MLOS/pull/2

bpkroth:
Also, I just installed SMAC and it pulls in a bunch of other requirements that we probably don't need.  Maybe see if there's a keyword method of installing it with fewer dependencies pulled in?

bpkroth:
Another one to help with the schema stuff:
https://github.com/kkanellis/MLOS/pull/3

kkanellis:
> Also, I just installed SMAC and it pulls in a bunch of other requirements that we probably don't need. Maybe see if there's a keyword method of installing it with fewer dependencies pulled in?

Unfortunately, I don't see anything that can help with that.

https://github.com/automl/SMAC3/blob/main/setup.py

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
kkanellis: Just added a `.cleanup()` method to `Optimizer` class, as well as an implementation for `SmacOptimizer` concrete class.

motus: Ideally, we need both `.cleanup()` and `__enter__/__exit__`, but having just `.cleanup()` is a good first step. I was proposing `.close()` as a name just because every other class I know has it (so it's like `Closeable` interface in Java). but I'm fine with `.cleanup()`, too

bpkroth:
@kkanellis I made another PR to try and address the Windows CI issues:
https://github.com/kkanellis/MLOS/pull/4

bpkroth:
> @kkanellis I made another PR to try and address the Windows CI issues: [kkanellis#4](https://github.com/kkanellis/MLOS/pull/4)

@motus can you please also try this locally for yourself?

mlos_core/mlos_core/optimizers/__init__.py
```diff
@@ -37,12 +39,15 @@ class OptimizerType(Enum):
 
     EMUKIT = EmukitOptimizer
     """An instance of EmukitOptimizer class will be used"""
+
```
motus, line 42: flake8 complains about trailing whitespace characters here

motus (changes requested):
After merging in [kkanellis#4](https://github.com/kkanellis/MLOS/pull/4) it actually worked on my Windows PC! Let's merge in [kkanellis#4](https://github.com/kkanellis/MLOS/pull/4) (and maybe main with #365) into this PR and I'm ready to approve it!

mlos_core/mlos_core/optimizers/bayesian_optimizers.py
bpkroth: Agreed, it's just a semantics thing about what the method returns.  If it's an `@property` that returns a `Callable`, then `acquisition_function` that can be called makes sense to me.  Else, `invoke_acquisition_function` makes more sense to me.

Anyways, forked off to another issue here: #341

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
```diff
@@ -192,7 +189,13 @@ def objective(point: pd.DataFrame) -> pd.Series:   # pylint: disable=invalid-nam
         space_adapter_type=SpaceAdapterType.LLAMATUNE,
         space_adapter_kwargs=space_adapter_kwargs
     )
+
+    # Initialize an optimizer that uses the original space
+    optimizer: BaseOptimizer = OptimizerFactory.create(input_space, optimizer_type, optimizer_kwargs=kwargs)
```
motus, line 194: We'll need named parameters here. Please merge the latest master, resolve the conflicts, and fix that part
```suggestion
    optimizer: BaseOptimizer = OptimizerFactory.create(
        parameter_space=input_space,
        optimizer_type=optimizer_type,
        optimizer_kwargs=kwargs,
    )
```

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
```diff
@@ -228,7 +232,7 @@ def objective(point: pd.DataFrame) -> pd.Series:   # pylint: disable=invalid-nam
     # .surrogate_predict method not currently implemented if space adapter is employed
     if isinstance(llamatune_optimizer, BaseBayesianOptimizer):
         with pytest.raises(NotImplementedError):
-            _ = llamatune_optimizer.surrogate_predict(llamatune_best_observation[['x']])
+            _ = llamatune_optimizer.surrogate_predict(llamatune_best_observation[['x', 'y']])
```
motus, line 235:
```suggestion
            llamatune_optimizer.surrogate_predict(llamatune_best_observation[['x', 'y']])
```

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
```diff
@@ -203,7 +206,8 @@ def objective(point: pd.DataFrame) -> pd.Series:   # pylint: disable=invalid-nam
 
         # loop for llamatune-optimizer
         suggestion = llamatune_optimizer.suggest()
-        assert suggestion['x'].iloc[0] == suggestion['y'].iloc[0]   # optimizer explores 1-dimensional space
+        _x, _y = suggestion['x'].iloc[0], suggestion['y'].iloc[0]
+        assert _x == _y or isclose(_x + _y, 3.)  # optimizer explores 1-dimensional space
```
motus, line 210: maybe use `pytest.approx()` instead?
```suggestion
        assert _x == pytest.approx(_y, 0.001)
```

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
motus: Let's make it here because all other optimizers and adapters already have it and the factory expects the parameters to be named. Besides, it allows us to get rid of that `too-many-arguments` pylint exception above

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -0,0 +1,255 @@
...
+        parameter_space: ConfigSpace.ConfigurationSpace,
+        space_adapter: Optional[BaseSpaceAdapter] = None,
+        seed: Optional[int] = 0,
+        run_name: Optional[str] = None,
+        output_directory: Optional[str] = None,
+        max_trials: int = 100,
+        n_random_init: Optional[int] = 10,
+        n_random_probability: Optional[float] = 0.1,
+    ):
+        super().__init__(parameter_space, space_adapter)
```
motus, line 67: we ues named parameters in the base constructor, too, so you'll need that:
```suggestion
        super().__init__(
            parameter_space=parameter_space,
            space_adapter=space_adapter,
        )
```

motus (commented):
Added a few more comments. Otherwise, I am ready to approve it after merging in the latest master

motus (commented):
Added a few more comments. Otherwise, I am ready to approve it after merging in the latest master

motus:
@kkanellis oops, I am afraid I've screwed up the last merge. can you please fix it so I don't have to send a PR? Sorry about the mess!

bpkroth:
@kkanellis let us know if you have some time to finish this.  If not I may fork it and do the rest next week so we have the feature (will keep you in the history obviously).

bpkroth:
@kkanellis I made https://github.com/kkanellis/MLOS/pull/6 to fixup the remaining build issues.

bpkroth:
> LGTM

Actually, there was another tweak to the Windows CI pipelines required:
https://github.com/kkanellis/MLOS/pull/7

bpkroth:
@kkanellis I left (yet another) tweak for the CI stuff here:
https://github.com/kkanellis/MLOS/pull/8

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -0,0 +1,257 @@
...
+
+from pathlib import Path
+from typing import Mapping, Optional
+from tempfile import TemporaryDirectory
+
+import ConfigSpace
+import numpy.typing as npt
+import pandas as pd
+
+from mlos_core.optimizers.bayesian_optimizers import BaseBayesianOptimizer
```
motus, line 18:
```suggestion
from mlos_core.optimizers.bayesian_optimizers.bayesian_optimizer import BaseBayesianOptimizer
```
cyclic import

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
```diff
@@ -203,7 +195,17 @@ def objective(point: pd.DataFrame) -> pd.Series:   # pylint: disable=invalid-nam
         space_adapter_type=SpaceAdapterType.LLAMATUNE,
         space_adapter_kwargs=space_adapter_kwargs,
     )
+
```
motus, line 198:
```suggestion
    # pylint: disable=too-many-locals
```
(should be at the beginning of the function, actually)

motus (approved):
two pylint issues prevent us from passing the CI checks in devcontainer, otherwise LGTM

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
```diff
@@ -203,7 +195,17 @@ def objective(point: pd.DataFrame) -> pd.Series:   # pylint: disable=invalid-nam
         space_adapter_type=SpaceAdapterType.LLAMATUNE,
         space_adapter_kwargs=space_adapter_kwargs,
     )
+    # pylint: disable=too-many-locals
```
motus, line 198:
```suggestion

```

PR #351, opened 2023-05-10
mlos_core add FLAML optimizer

kkanellis:
This PR adds support for [FLAML](https://microsoft.github.io/FLAML/) optimizer.


mlos_core/mlos_core/optimizers/flaml_optimizer.py
```diff
@@ -0,0 +1,170 @@
...
+        self.low_cost_partial_config = low_cost_partial_config
+
+        self.evaluated_configs: Dict[int, dict] = {}
+        self._suggested_config: Optional[dict]
+
+    def _register(self, configurations: pd.DataFrame, scores: pd.Series,
+                  context: Optional[pd.DataFrame] = None) -> None:
+        """Registers the given configurations and scores.
+
+        Doesn't do anything on the RandomOptimizer except storing configurations for logging.
```
bpkroth, line 54:
```suggestion
        Doesn't do anything on the RandomOptimizer except storing configurations for logging.
```

Bad comment?

mlos_core/setup.py
```diff
@@ -28,6 +28,7 @@
 extra_requires: Dict[str, List[str]] = {
     'emukit': ['emukit'],
     'skopt': ['scikit-optimize<=0.9.0'],  # FIXME: temporarily work around some version mismatch issues (PR 850)
+    'flaml': ['flaml[blendsearch]'],
```
bpkroth, line 31: Don't forget to add to `mlos_bench`

bpkroth, line 31: Are there other suboptions that might make sense here?

mlos_core/mlos_core/spaces/__init__.py
```diff
@@ -85,3 +85,39 @@ def _one_parameter_convert(parameter: ConfigSpace.hyperparameters.Hyperparameter
         raise ValueError(f"Type of parameter {parameter} ({type(parameter)}) not supported.")
 
     return emukit.core.ParameterSpace([_one_parameter_convert(param) for param in config_space.get_hyperparameters()])
+
+
+def configspace_to_flaml_space(config_space: ConfigSpace.ConfigurationSpace) -> dict:
```
bpkroth, line 90: Here's another one we might want to start splitting out to multiple files.
One reason to do this would be to adjust the way we handle imports so they can be top level instead of nested.
Separate PR  for that too ...

mlos_core/setup.py
kkanellis: I don't think so. Here's the full list of suboptions: https://microsoft.github.io/FLAML/docs/Installation

Most of these suboptions are used for AutoML purposes. Blendsearch is the only that makes sense for our cases imho.

mlos_core/mlos_core/spaces/__init__.py
kkanellis: Yep, happy to do this, once both PRs are merged! :)

motus (commented):
Looks good, and all tests pass on my PC (on both Windows and Linux devcontainer). Also, kudos for having zero pylint/flake8/mypy issues!

I've left a few nitpicks - please take a look and merge in the latest updates, and I'll be ready to approve and merge it in

motus:
Hmm, looks like I've screwed up something during the automatic merge 😦 @kkanellis can you please take a look?

kkanellis:
> Hmm, looks like I've screwed up something during the automatic merge frowning @kkanellis can you please take a look?

I think the problem is on the `mlos_bench` side -- not `mlos_core`: 


> FAILED
> mlos_bench/mlos_bench/tests/config/schemas/optimizers/test_optimizer_schemas.py::test_case_coverage_mlos_core_optimizer_type[OptimizerType.FLAML-bad] - NotImplementedError: Missing test case for type bad for MlosCore Optimizer type OptimizerType.FLAML

motus:
> I think the problem is on the `mlos_bench` side

oh, thanks for checking. I'll take a look and send a PR against your branch soon!

motus:
@kkanellis I've sent a PR against your branch that fixes the CS conversion issue. I am not sure what is the right way of fixing the unit tests that fail in mlos_bench. @bpkroth can you please take a look? Apparently, now our test coverage tests in mlos_bench fail every time a new optimizer appear in mlos_core. ☹️ Should we relax our runes a bit for such case? WDYT?

motus (approved):
Looks good!

PR #349, opened 2023-05-10
mlos_bench config json schema validation - storage

bpkroth:
Extends #340 and #346 to storage configs.
See Also: #331

PR #346, opened 2023-05-10
mlos_bench config json schema validation - tunable params

bpkroth:
Extends #340 to tunable params.
See Also: #331

PR #340, opened 2023-05-08
mlos_bench config json schema validation: optimizers and tunable_values

bpkroth:
This PR introduces initial json schemas for mlos_bench optimizer configs and tunable_values configs, their validation at load time, and tests for both.

Future PRs will handle other config types.
See Also: #331

Issue #331, opened 2023-05-03
mlos_bench: add json schemas for configs


bpkroth:
This would improve both validation logic as well as editing and documentation of supported parameters.

Forked off from #313

- [x] optimizers - #340 
- [x] tunable params - #346
- [x] tunable values - #340 
- [x] globals - #352 
- [x] environments - #359 
- [ ] services - #368 
- [x] storage - #349 
- [x] CLI args config - #352

PR #352, opened 2023-05-10
mlos_bench config json schema validation - cli

bpkroth:
Extends #340, #346, #349 to cli configs.
See Also: #331 

Also updates `--log_level` to support named logging levels (e.g. `INFO`)

- [x] add tests
  - [x] also backfilling tests of loading cli config examples (similar to what #313 did)
- [x] global config handling


mlos_bench/mlos_bench/launcher.py
```diff
@@ -59,6 +61,7 @@ def __init__(self, description: str):
             config = {}
 
         log_level = args.log_level or config.get("log_level", _LOG_LEVEL)
+        log_level = logging.getLevelName(log_level)
```
motus, line 64: We need to double check if this works. First of all, `_LOG_LEVEL` declared above is still an `int`. Second, `logging.getLevelName()` seem to have a bit weird behavior:
```python
In [7]: logging.getLevelName(logging.INFO)
Out[7]: 'INFO'

In [8]: logging.getLevelName("INFO")
Out[8]: 20
```
😮 

I just hope that `.setLevel()` is equally weird (ok, "flexible") and it all works out at the end 😄

motus: Besides, it might be a good idea to allow *both* `int` *and* enum log levels - it is ok to have a log level of, e.g., 15 for messages which importance is somewhere between debug and info

mlos_bench/mlos_bench/services/config_persistence.py
```diff
@@ -120,7 +127,24 @@ def load_config(self, json_file_name: str) -> Union[Dict[str, Any], List[Dict[st
...
-            return json5.load(fh_json)  # type: ignore[no-any-return]
+            config = json5.load(fh_json)
+        if schema_type is not None:
+            try:
+                schema_type.validate(config)
+            except (ValidationError, SchemaError) as ex:
+                _LOG.error("Failed to validate config %s against schema type %s at %s",
+                           json_file_name, schema_type.name, schema_type.value)
+                raise ValueError(f"Failed to validate config {json_file_name} against " +
+                                 "schema type {schema_type.name} at {schema_type.value}") from ex
```
motus, line 138:
```suggestion
                                 f"schema type {schema_type.name} at {schema_type.value}") from ex
```

motus (commented):
Left a few comments, but let's merge in other PRs first - I am a bit confused about which files I've already reviewed before

mlos_bench/mlos_bench/config/cli/TODO
bpkroth, line 1: - [ ] Did that.  Remove this file.

mlos_bench/mlos_bench/launcher.py
bpkroth: I did some limited tests and it appears to work.  And will properly throw an error if you give it a garbage value like `--log-level foo`
I'm not sure why we'd want to support in-between log levels though - they're pretty non-standard and not very intuitive.

mlos_bench/mlos_bench/tests/config/schemas/tunable-values/test_tunable_values_schemas.py
```diff
@@ -27,6 +27,6 @@
 @pytest.mark.parametrize("test_case_name", sorted(TEST_CASES.by_path))
 def test_tunable_values_configs_against_schema(test_case_name: str) -> None:
     """
-    Checks that the optimizer config validates against the schema.
+    Checks that the tunable values config validates against the schema.
```
bpkroth, line 30: fixup a bad copy/paste comment

mlos_bench/mlos_bench/services/config_persistence.py
```diff
@@ -56,7 +56,7 @@ def __init__(self, config: Optional[Dict[str, Any]] = None,
             An optional parent service that can provide mixin functions.
         """
         super().__init__(config, parent)
-        self._config_path = self.config.get("config_path", [])
+        self._config_path: List[str] = self.config.get("config_path", [])
```
bpkroth, line 59: type fixups

mlos_bench/mlos_bench/launcher.py
bpkroth: I have another small fix for this pending but github won't let me push atm (https://www.githubstatus.com/ is showing issues).

motus (approved):
Looks good!

PR #355, opened 2023-05-13
bugfix: make mlos_core optimizer shim work when the optimization parameter name is not "score"

motus:


mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
```diff
@@ -93,5 +93,6 @@ def get_best_observation(self) -> Union[Tuple[float, TunableGroups], Tuple[None,
         if len(df_config) == 0:
             return (None, None)
         params = df_config.iloc[0].to_dict()
-        score = params.pop(self._opt_target) * self._opt_sign
+        _LOG.debug("Best observation: %s", params)
+        score = params.pop("score") * self._opt_sign  # mlos_core always uses the `score` column
```
bpkroth, line 97: That's true for the format of the mlos_core dataframe for now (will change once multi-objective gets reintegrated), but we still need to convert whatever the declared target from the results of the application's format into that field for them.

I thought the design goal was to provide a handful of well known parsers to automatically extract relevant scores out of the provided application logs (e.g. csv column extract, json field extract, etc.)

motus: yes, we do extract multiple metrics from the logs; the limitation is currently on the mlos_core side, where we hardcode a single metric named "score" as a column in a pandas dataframe. This is just a minimal fix to make mlos_bench work with multiple metrics without many changes to mlos_core. Later we will implement multi-objective optimization in mlos_core and revert this fix

bpkroth: kk

bpkroth (approved):

Issue #363, opened 2023-05-18
mlos_core: require named params for all instantiations


bpkroth:
Discussed in #341, we may want to make all optimizer __init__ params required to be named.  e.g.

```python
def __init__(self, *, all, other, params)
   ...
```

PR #365, opened 2023-05-18
Use keyword arguments in all mlos_core optimizers and space adapters

motus:
As per #363 
Also fix some collateral pylint issues and remove unnecessary pylint exceptions.

### DIFF
--- a/mlos_core/mlos_core/tests/spaces/adapters/llamatune_test.py
+++ b/mlos_core/mlos_core/tests/spaces/adapters/llamatune_test.py
@@ -66,11 +66,18 @@ def test_num_low_dims(num_target_space_dims: int, param_space_kwargs: dict) -> N
 
     # Number of target parameter space dimensions should be fewer than those of the original space
     with pytest.raises(ValueError):
-        _ = LlamaTuneAdapter(input_space, num_low_dims=len(input_space.get_hyperparameter_names()))
+        LlamaTuneAdapter(
+            orig_config_space=input_space,
+            num_low_dims=len(input_space.get_hyperparameter_names())
+        )
 
     # Enable only low-dimensional space projections
-    adapter = LlamaTuneAdapter(input_space, num_low_dims=num_target_space_dims,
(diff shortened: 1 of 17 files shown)


motus:
@kkanellis this might require a small fix to the SMAC optimizer constructor (basically, use `__init__(self, *, ...)` and make sure all parameters from the base class have the same names in the child constructor)

mlos_core/mlos_core/tests/spaces/adapters/llamatune_test.py
```diff
@@ -115,29 +122,30 @@ def test_special_parameter_values_validation() -> None:
...
     with pytest.raises(NotImplementedError):
         special_param_values_dict_2 = {'cont': -1}
-        _ = LlamaTuneAdapter(
-            input_space,
+        LlamaTuneAdapter(
+            orig_parameter_space=input_space,
             num_low_dims=2,
             special_param_values=special_param_values_dict_2,
-            max_unique_values_per_param=None
+            max_unique_values_per_param=None,
```
bpkroth, line 138: Would be cool if we could be more consistent about trailing commas or not and maybe even enforce it somehow.
I personally like to use them everywhere it's allowed since it makes `yy P cW` style vi edits easier and leaves git diffs only the line that got added, but I understand not all languages or sites support that.

bpkroth (approved):
LGTM

mlos_core/mlos_core/tests/spaces/adapters/llamatune_test.py
kkanellis: Yep, I agree and I try to put trailing commas everywhere -- I just missed that one!

kkanellis:
> @kkanellis this might require a small fix to the SMAC optimizer constructor (basically, use `__init__(self, *, ...)` and make sure all parameters from the base class have the same names in the child constructor)

Let's merge this first, and I'll make changes to SMAC.

PR #359, opened 2023-05-16
mlos_bench config json schema validation - environments

bpkroth:
Extends #340, #346, #349, #352 to Environment configs.
See Also: #331 

- [x] main schema as a set of subschemas
- [x] connect schema validation in the mlos_bench code
- [x] test-cases

PR #368, opened 2023-05-19
mlos_bench config json schema validation - services

bpkroth:
Extends #340, #346, #349, #352, #359 to Service configs.
See Also: #331 

- [x] main schema as a set of subschemas
- [x] connect schema validation in the mlos_bench code
- [x] test-cases

To do this we removed support for configs that were flat lists of service configs: `[ {"class": "service.class"} ]` and turned those into objects that have a single key `"services"` with the list instead. `{ "services": [ ... ] }`
This makes sure that all configs are dicts/objects that we can optionally put `"$schema"` attributes on but also simplifies some of the config loading logic.


mlos_bench/mlos_bench/config/schemas/config_schemas.py
```diff
@@ -56,6 +56,8 @@ def _load_schemas(cls) -> None:
                     schema = json.load(schema_file)
                     cls._SCHEMA_STORE[file_path] = schema
                     # Let the schema be referenced by its id as well.
+                    assert "$id" in schema
+                    assert schema["$id"] not in cls._SCHEMA_STORE
```
bpkroth, line 60: Make sure we don't overwrite a schema by id - that's an error in schema development.

mlos_bench/mlos_bench/config/services/remote/azure/service-fileshare.jsonc
```diff
@@ -1,12 +1,12 @@
-[
```
bpkroth, line 1: Removing support for flat lists of service configs.

mlos_bench/mlos_bench/config/services/remote/azure/service-linux-vm-ops.jsonc
```diff
@@ -1,22 +1,24 @@
-[
-    {
-        "class": "mlos_bench.services.remote.azure.AzureVMService",
+{
+    //"$schema": "../../../../config/schemas/services/service-schema.json",
 
-        "config": {
+    "services": [
```
bpkroth, line 4: Here's an example of using lists of service configs as an object.

Makefile
```diff
@@ -113,6 +113,16 @@ build/pylint.%.${CONDA_ENV_NAME}.build-stamp: build/conda-env.${CONDA_ENV_NAME}.
 	conda run -n ${CONDA_ENV_NAME} pylint -j0 $(filter-out .pylintrc,$+)
 	touch $@
 
+.PHONY: flake8
+flake8: conda-env build/flake8.mlos_core.${CONDA_ENV_NAME}.build-stamp build/flake8.mlos_bench.${CONDA_ENV_NAME}.build-stamp
```
bpkroth, line 117: Rules to (optionally for now) run flake8 on all files.

motus (approved):
LGTM! will test in master

PR #371, opened 2023-05-22
Duplicated Redis, changing every ocurance of redis w\ mysql. Also add…

yyshadyy:
…ed some tunables


yyshadyy:
@microsoft-github-policy-service agree company="Microsoft"

PR #374, opened 2023-05-24
move duplicated functionality for merging config parameters into a util function

motus:


mlos_bench/mlos_bench/util.py
```diff
@@ -69,13 +107,7 @@ def prepare_class_load(config: dict,
         Name of the class to instantiate and its configuration.
     """
     class_name = config["class"]
-    class_config = config.setdefault("config", {})
```
bpkroth, line 72: nit: concise onliners are nice, but I don't think it's obvious when reading that `setdefault()` returns the key being set, so I kinda prefer this line that makes it more explicit that the `class_config` is its own entity and we're acting on merging that below rather than some ephemeral looking operation that's happening inside the merge_params(...) call.

mlos_bench/mlos_bench/services/config_persistence.py
```diff
@@ -24,7 +24,7 @@
 from mlos_bench.services.base_service import Service
 from mlos_bench.services.types.config_loader_type import SupportsConfigLoading
 from mlos_bench.tunables.tunable_groups import TunableGroups
-from mlos_bench.util import instantiate_from_config, path_join, BaseTypeVar
+from mlos_bench.util import merge_parameters, instantiate_from_config, path_join, BaseTypeVar
```
bpkroth, line 27:
```suggestion
from mlos_bench.util import instantiate_from_config, merge_parameters, path_join, BaseTypeVar
```

nit: keeps it more sorted and matching the style of the one above as well

bpkroth (approved):
minor nits, but otherwise LGTM

mlos_bench/mlos_bench/util.py
motus: agree. Let me split it back into two lines here

PR #376, opened 2023-05-26
Migrate Azure-specific parameters from VMEnv to AzureVMService

motus:
Note that in this PR we duplicate `vmName` in all environments that use `AzureVMService`. I will move this parameter into a global config in the `CompositeEnv` in the next PR.

### DIFF
--- a/mlos_bench/mlos_bench/services/remote/azure/azure_services.py
+++ b/mlos_bench/mlos_bench/services/remote/azure/azure_services.py
@@ -419,13 +419,14 @@ class AzureVMService(Service, SupportsVMOps, SupportsRemoteExec):
             parameters extracted from the response JSON, or {} if the status is FAILED.
             Status is one of {PENDING, SUCCEEDED, FAILED}
         """
+        config = merge_parameters(dest=self.config.copy(), source=params)
         params = merge_parameters(dest=self._deploy_params.copy(), source=params)
-        _LOG.info("Deploy: %s :: %s", self.config["deploymentName"], params)
+        _LOG.info("Deploy: %s :: %s", config["deploymentName"], params)
 
         url = self._URL_DEPLOY.format(
-            subscription=self.config["subscription"],
-            resource_group=self.config["resourceGroup"],
-            deployment_name=self.config["deploymentName"],
+            subscription=config["subscription"],
+            resource_group=config["resourceGroup"],
+            deployment_name=config["deploymentName"],
         )
 
         json_req = {
@@ -453,13 +454,14 @@ class AzureVMService(Service, SupportsVMOps, SupportsRemoteExec):
             _LOG.info("Response: %s", response)
 
         if response.status_code == 200:
-            return (Status.PENDING, params)
+            return (Status.PENDING, config)
         elif response.status_code == 201:
             output = self._extract_arm_parameters(response.json())
             if _LOG.isEnabledFor(logging.DEBUG):
                 _LOG.debug("Extracted parameters:\n%s", json.dumps(output, indent=2))
             params.update(output)
             params.setdefault("asyncResultsUrl", url)
+            params.setdefault("deploymentName", config["deploymentName"])
             return (Status.PENDING, params)
         else:
             _LOG.error("Response: %s :: %s", response, response.text)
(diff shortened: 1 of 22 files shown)


mlos_bench/mlos_bench/services/remote/azure/azure_services.py
```diff
@@ -397,10 +372,25 @@ def _check_deployment(self, _params: dict) -> Tuple[Status, dict]:
...
             Status is one of {SUCCEEDED, PENDING, FAILED}
         """
-        _LOG.info("Check deployment: %s", self.config["vmName"])
+        config = merge_parameters(
+            dest=self.config.copy(),
+            source=params,
+            required_keys=[
+                "subscription",
+                "resourceGroup",
+                "deploymentName",
```
bpkroth, line 381: nit:
Most of these required keys are repeated across `merge_parameters()` calls through the file.
Does it make sense to move those into `__init__` so that they can be caught earlier at config instantiation time, rather than at runtime and then remove them from the other sites as implicitly required and reduce some boilerplate?

mlos_bench/mlos_bench/services/types/vm_provisioner_type.py
```diff
@@ -72,32 +72,47 @@ def vm_start(self, params: dict) -> Tuple["Status", dict]:
...
+
         Returns
         -------
         result : (Status, dict={})
             A pair of Status and result. The result is always {}.
             Status is one of {PENDING, SUCCEEDED, FAILED}
         """
 
-    def vm_restart(self) -> Tuple["Status", dict]:
+    def vm_restart(self, params: dict) -> Tuple["Status", dict]:
```
bpkroth, line 91: `params` is a flat `dict` of *all* the tunables (for that env), global configs, const args, etc.
That could be a lot of things.
Some of which may involve some things that aren't excepted by the Service that implements this since it may have different semantics for different systems.  Just wondering if we're going to have to filter this in the future ...

bpkroth (changes requested):
Generally good.  Needs some more test cases for the schema changes though.

bpkroth:
> Generally good. Needs some more test cases for the schema changes though.

Made this to help with that and tweak a few other things I noticed.  Can split those out to another PR as well if you want, but they're small and somewhat related.
https://github.com/motus/MLOS/pull/4

mlos_bench/mlos_bench/services/types/vm_provisioner_type.py
motus: yes, we will definitely need to filter those, I just have to think of an elegant way to do it. Will do in the subsequent PRs

mlos_bench/mlos_bench/services/remote/azure/azure_services.py
```diff
@@ -231,9 +203,9 @@ def _azure_vm_post_helper(self, url: str) -> Tuple[Status, dict]:
         # Logical flow for async operations based on:
         # https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/async-operations
         if response.status_code == 200:
-            return (Status.SUCCEEDED, {})
+            return (Status.SUCCEEDED, params)
```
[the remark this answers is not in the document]
motus: because below we mutate the `params` which might be unexpected to the caller

mlos_bench/mlos_bench/services/remote/azure/azure_services.py
motus: it's a bit complicated. `required_keys` are not only for validation, but also to enforce copying the data from the global config if the corresponding placeholder parameter is omitted from the config section. I'll think if there is a way to do it more elegantly and send a new PR later

mlos_bench/mlos_bench/services/remote/azure/azure_services.py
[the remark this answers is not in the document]
bpkroth: But then if the caller is expecting to have a fresh copy returned and they go to mutate those in turn, they're mutating their original copy in one case and a new copy in the other case.  It's a little inconsistent is all.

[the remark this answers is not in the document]
motus: agree. I'll push the fix in a second

bpkroth:
I know I signed off on this already, but wondering if we might need some additional tests here too.

PR #377, opened 2023-05-31
Propagate CompositeEnv const_args parameters to its children

motus:


mlos_bench/mlos_bench/environments/composite_env.py
```diff
@@ -52,6 +52,13 @@ def __init__(self,
 
         self._children: List[Environment] = []
 
+        # To support trees of composite environments (e.g. for multiple VM experiments),
+        # each CompositeEnv gets a copy of the original global config and adjusts it with
+        # the `const_args` specific to it.
+        global_config = (global_config or {}).copy()
+        for (key, val) in self._const_args.items():
+            global_config.setdefault(key, val)
```
bpkroth, line 60: Tests for this change?

bpkroth: e.g. a sample CompositeEnv tree config and then make changes to each side and not the other and verify that the parent and sibling was not altered.

motus: you're right. will add now

mlos_bench/mlos_bench/tests/environments/composite_env_test.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+@pytest.fixture
+def composite_env(tunable_groups: TunableGroups) -> CompositeEnv:
+    """
+    Test fixture for CompositeEnv.
+    """
+    return CompositeEnv(
+        name="Composite Test Environment",
+        config={
```
bpkroth, line 25: Another example that included nested CompositeEnvironments would also be good, but our config schema doesn't currently support that directly (have to use `include_children`), so maybe we fix both in a future PR.

bpkroth: To do that we can use a variant on [recursive json schemas](http://json-schema.org/understanding-json-schema/structuring.html#recursion).

bpkroth (approved):
LGTM
I left some comments about additional tests, but given they require more changes I think we can punt that to a future PR.

PR #382, opened 2023-06-05
Nested composite env schema and dynamic const_args support

### LINKED ISSUE #381: Supported nested CompositeEnvs in json config schemas
Another example that included nested CompositeEnvironments would also be good, but our config schema doesn't currently support that directly (have to use `include_children`), so maybe we fix both in a future PR.

_Originally posted by @bpkroth in https://github.com/microsoft/MLOS/pull/377#discussion_r1213330785_
            
To do that we can use a variant on [recursive json schemas](http://json-schema.org/understanding-json-schema/structuring.html#recursion).

bpkroth:
- Fixes #381 to add nested inline CompositeEnv config support
- Enforces uniqueness on some more arrays.
- Cleanup some old comments.
- Add support for dynamic `$param_name` substitution in `const_args`

- [x] Add additional tests


mlos_bench/mlos_bench/config/schemas/environments/composite-env-subschema.json
```diff
@@ -33,16 +33,25 @@
...
+                                    "uniqueItems": true,
                                     "minItems": 1
                                 },
                                 "children": {
                                     "description": "List of inline child environment configs to include in the composite environment.",
                                     "type": "array",
                                     "items": {
-                                        "$comment": "Each config included can be a leaf environment config.  We do not currently support inline nested composite environments.",
-                                        "$ref": "./leaf-environment-subschemas.json",
-                                        "unevaluatedProperties": false
```
bpkroth, line 44: Add this back in?

mlos_bench/mlos_bench/tests/environments/composite_env_test.py
```diff
@@ -70,11 +112,38 @@ def test_composite_env_params(composite_env: CompositeEnv) -> None:
...
         "vmName": "Mock VM",        # const_args from the parent
         "EnvId": 2,                 # const_args from the child
         "idle": "halt",             # tunable_params from the parent
+        # "someConst": "root"        # not required, so not passed from the parent
+    }
+    assert isinstance(composite_env.children[2], CompositeEnv)
+    # CompositeEnv child should receive everything from the parent since it didn't specify a tunable_params subgroup to filter on.
+    # It should also override the const_args.
+    new_params = composite_env.parameters.copy()
+    new_params["vmName"] = "Nested Mock VM"
```
bpkroth, line 127: @motus a topic to discuss: right now the way that CompositeEnv and merge_parameters() works, it appears that the parent value overrides the value in the children.
I guess that makes sense for the "placeholder" values.
But for child CompositeEnvs, where we might want to have one side be a client and another side be the server, we need to override it in the child CompositeEnv, and then let that propogate down.
Thoughts?

mlos_bench/mlos_bench/tests/environments/composite_env_test.py
```diff
@@ -54,6 +55,47 @@ def composite_env(tunable_groups: TunableGroups) -> CompositeEnv:
...
+                                "class": "mlos_bench.environments.mock_env.MockEnv",
+                                "config": {
+                                    "tunable_params": ["boot"],
+                                    "const_args": {
+                                        "vmName": "Placeholder VM",
+                                        "someConst": "leaf"
+                                    },
+                                    "required_args": ["vmName", "someConst"],
+                                    "range": [60, 120],
+                                    "metrics": ["score"],
```
bpkroth, line 94: @motus 
Another issue is some of the repetition in these configs.  I wonder if we could have a sort "inheritable" args vs. not config.
Thoughts?

motus (commented):
overall, looks good. I'll test it on my PC now and check if the data propagates correctly down the environment hierarchy

mlos_bench/mlos_bench/tests/environments/composite_env_test.py
motus: @bpkroth you're right, parameters propagate top to bottom, and in this example *all* `vmName` values in *all* environments regardless of their position in the tree will be set to the same value of `"Mock VM"` as it is specified at the top level. We do it this way so we can pass the global parameters (e.g., `accessToken` or `storageFileShareName`) downstream to leaf environments and services. In our MySQL setup we don't have any conflicts between the top and bottom levels - at the top we provision two VMs with the names parameterized as `vmName_server` and `vmName_client`, and at the next level we use `vmName` (this parameter does not occur anywhere above) to which we assign a copy of either `vmName_server` or `vmName_client` value. Yes, there is duplication, and later we might want to introduce some copy rules, so we can assign, e.g., `"vmName": "$vmName_server"` but for the time being it allows us to provision two VMs and run our experiments. Let's discuss it next week when we have time

mlos_bench/mlos_bench/tests/environments/composite_env_test.py
motus: yeah, I was thinking about it, but I've decided to avoid any big changes to the framework for now and focus on the experiments; later we can think of an elegant way of solving this issue - but maybe we don't even have to fix it in JSON but instead introduce some high-level DSL that would compile into JSONs and scripts that we use now. Again, let's discuss when we have time

mlos_bench/mlos_bench/config/schemas/environments/composite-env-subschema.json
```diff
@@ -33,16 +33,26 @@
                                         "type": "string",
                                         "pattern": "[.]json[c]?$"
```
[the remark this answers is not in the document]
bpkroth: `.json` files are also valid and more common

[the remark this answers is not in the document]
bpkroth: The backslashes required ugly escapes and I find the attachment rules of `?` in regex easier to read this way.

mlos_bench/mlos_bench/services/config_persistence.py
```diff
@@ -401,11 +424,13 @@ def load_environment(self, json_file_name: str,
         """
         config = self.load_config(json_file_name, ConfigSchema.ENVIRONMENT)
         assert isinstance(config, dict)
-        return self.build_environment(config, tunables, global_config, service)
+        return self.build_environment(config, tunables, global_config, parent_args, service)
 
-    def load_environment_list(self, json_file_name: str,
+    def load_environment_list(self,     # pylint: disable=too-many-arguments
```
bpkroth, line 429: Could get rid of that pylint warning by changing these all to `*` to require named params, but that change was more extensive and verbose so I skipped it for now.  Happy to change it in a future PR if we want to.

bpkroth:
@motus up to you, but I'm ok merging this even though Windows is failing atm and coming back to it later.  I think this should help avoid some of your other workaround hacks for schema use atm and let me improve #389 some more.

mlos_bench/mlos_bench/services/config_persistence.py
motus: sounds good! I'll fix it later when I have cycles

motus (approved):

PR #383, opened 2023-06-06
Explicitly specify which parameters to pass as shell env variables to the remote script

motus:
Presently we pass *all* Environment parameters (i.e., a combination of `const_args` and `tunable_params`) as shell env variables when we run a local or remote script. this approach has several problems:
* too many parameters leak into the shell environment;
* all names must be valid shell env variable identifiers (e.g., we crash on parameters like `ima.ahash_bufsize`)
* all values must be strings

In this PR:
* keep the old behavior by default;
* explicitly convert all values to strings and all keys to valid identifiers
* implement `ScriptEnv` base class for `RemoteEnv` and `LocalEnv` that contains `setup/run/teardown` script properties and the renaming logic for the shell script parameters
* add `shell_params` parameter so that the user can specify which parameters to pass to shell scripts
* add `shell_params_rename` parameter to explicitly rename the parameters, if necessary.

TODO:
* [x] add `shell_params` and `shell_params_rename` to JSON templates
* [x] fix the unit tests for JSON templates that involve `ScriptEnv` and derived classes

### DIFF
--- a/mlos_bench/mlos_bench/config/schemas/environments/common-environment-subschemas.json
+++ b/mlos_bench/mlos_bench/config/schemas/environments/common-environment-subschemas.json
@@ -40,8 +40,10 @@
             },
             "minProperties": 1
         },
+
         "command_lines": {
             "type": "array",
+            "description": "The command lines to execute.",
             "items": {
                 "$comment": "TODO: Add support for array syntax for command lines that execute without a shell.",
                 "type": [
@@ -54,12 +56,15 @@
             "type": "object",
             "properties": {
                 "setup": {
+                    "description": "The command lines to execute for the setup phase.",
                     "$ref": "#/$defs/command_lines"
                 },
                 "run": {
+                    "description": "The command lines to execute for the run phase.",
                     "$ref": "#/$defs/command_lines"
                 },
                 "teardown": {
+                    "description": "The command lines to execute for the teardown phase.",
                     "$ref": "#/$defs/command_lines"
                 }
             },
@@ -82,6 +87,48 @@
                 }
             ]
         },
+
+        "shell_param_type": {
+            "description": "All parameters must be valid shell variable names, else use shell_params_mapping.",
+            "type": "string",
+            "pattern": "^[a-zA-Z_][a-zA-Z0-9_]+$"
+        },
+        "shell_params_config": {
+            "type": "object",
+            "properties": {
+                "shell_params_match": {
+                    "description": "A list of regexes to match against the environment parameters. All matches will be exposed to shell commands as environment variables.",
+                    "type": "array",
+                    "items": {
+                        "type": "string",
+                        "format": "regex",
+                        "examples": [
+                            "^MLOS_[a-zA-Z0-9_]+$"
+                        ]
+                    }
+                },
+                "shell_params": {
+                    "description": "An explicit list of parameters (and their values) to pass to the shell as environment variables.  These will override any parameters matched by shell_params_match.",
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/$defs/shell_param_type"
+                    },
+                    "uniqueItems": true,
+                    "minItems": 1
+                },
+                "shell_params_mapping": {
+                    "description": "An explicit mapping of environment variable names from params to be populated with the parameter values and passed to the shell commands. These will override any parameters matched by shell_params_match.",
+                    "type": "object",
+                    "propertyNames": {
+                        "$ref": "#/$defs/shell_param_type"
+                    },
+                    "additionalProperties": {
+                        "type": "string"
+                    }
+                }
+            }
+        },
+
         "file_download_config": {
             "type": "object",
             "properties": {
--- a/mlos_bench/mlos_bench/environments/local/local_env.py
+++ b/mlos_bench/mlos_bench/environments/local/local_env.py
@@ -9,7 +9,7 @@ Scheduler-side benchmark environment to run scripts locally.
 import json
 import logging
 
-from typing import Optional, Tuple
+from typing import Iterable, Optional, Tuple
 
 import pandas
 
@@ -114,15 +114,7 @@ class LocalEnv(ScriptEnv):
                     json.dump(tunables.get_param_values(self._tunable_params.get_covariant_group_names()),
                               fh_tunables)
 
-            (return_code, _stdout, stderr) = self._local_exec_service.local_exec(
-                self._script_setup, env=self._params, cwd=temp_dir)
-
-            if return_code == 0:
-                _LOG.info("Local set up complete: %s", self)
-            else:
-                _LOG.warning("ERROR: Local setup returns with code %d stderr:\n%s",
-                             return_code, stderr)
-
+            return_code = self._local_exec(self._script_setup, temp_dir)
             self._is_ready = bool(return_code == 0)
 
         return self._is_ready
@@ -145,13 +137,8 @@ class LocalEnv(ScriptEnv):
 
         with self._local_exec_service.temp_dir_context(self._temp_dir) as temp_dir:
 
-            _LOG.info("Run script locally on: %s at %s", self, temp_dir)
-            (return_code, _stdout, stderr) = self._local_exec_service.local_exec(
-                self._script_run, env=self._params, cwd=temp_dir)
-
+            return_code = self._local_exec(self._script_run, temp_dir)
             if return_code != 0:
-                _LOG.warning("ERROR: Local run returns with code %d stderr:\n%s",
-                             return_code, stderr)
                 return (Status.FAILED, None)
 
             assert self._read_results_file is not None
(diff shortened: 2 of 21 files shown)


bpkroth:
> implement explicit renaming of the parameters

This PR is a good start, but this bit above isn't going to sustainable.
There's a note in the sysctl settings for instance to convert them to paths (e.g. `/proc/sys/vm/...`), which also don't work for env vars, but are more generalizable for other reasons (that's the only way to do `/sys/...` settings for instance.

I think providing the *option* to expose some params as env vars with a well understood conversion is a nice convenience, but in many cases some sort of wrapper script will need to be provided to handle json (e.g. `cat | jq`) to extract values as necessary from a file or string we provide.

mlos_bench/mlos_bench/environments/remote/remote_env.py
```diff
@@ -151,6 +156,19 @@ def teardown(self) -> None:
...
+        Get the *shell* environment parameters to be passed to the remote script.
+
+        Returns
+        -------
+        env_params : Dict[str, str]
+            Parameters to pass as *shell* environment variables into the script.
+            This is usually a subset of `_params` with some possible conversions.
+        """
+        keys = self._params.keys() if self._script_params is None else self._script_params
+        return {re.sub(r"\W", "_", key): str(self._params[key]) for key in keys}
```
bpkroth, line 170: That might be too broad.  Maybe make it more explicit the type of characters we do or do not support?

mlos_bench/mlos_bench/tests/services/remote/azure/azure_services_test.py
```diff
@@ -111,7 +111,7 @@ def test_remote_exec_status(mock_requests: MagicMock, azure_vm_service: AzureVMS
     mock_response.status_code = http_status_code
     mock_requests.post.return_value = mock_response
 
-    status, _ = azure_vm_service.remote_exec(script, {})
+    status, _ = azure_vm_service.remote_exec(script, {}, {})
```
bpkroth, line 114: named params?

mlos_bench/mlos_bench/services/types/remote_exec_type.py
bpkroth, line 1: Not quite sure how we implement this, but it'd be cool to mock out a test script that does a printenv and verifies that it only got the known env params passed to it and not some others.

bpkroth (approved):
Left a few improvement comments, but overall looks good.  Thanks!

bpkroth:
Further improvement we just discussed offline:
1. Support a bare list of setting names that can be converted directly and exposed to the script.
    -  enforce in both the schema and the code that any setting name listed there that can't be directly converted to an environment variable name should be disallowed (can use regex constraints in the `"pattern"` property in the `"items"` property of the `"array"` to do this):
```json
{
    "settings_as_env_vars": {
        "type": "array"
        "items": {
            "type": "string",
            "pattern": "^[a-zA-Z_][a-zA-Z0-9_]+$"
        }
    }
}
```
2. Also support a "mapping" for all other "settings -> env name" that might be needed.
    Can enforce a similar sort of constraint on the property values in this case as above.

mlos_bench/mlos_bench/environments/remote/remote_env.py
motus: done

mlos_bench/mlos_bench/tests/services/remote/azure/azure_services_test.py
motus: done

mlos_bench/mlos_bench/services/types/remote_exec_type.py
motus: yes, this is a great idea and should be doable. Can you please help me with JSON schema tests? I will add this check as soon as all other tests pass

mlos_bench/mlos_bench/environments/local/local_env.py
```diff
@@ -183,6 +162,31 @@ def teardown(self) -> None:
...
+        script : Iterable[str]
+            Lines of the script to run locally.
+            Treat every line as a separate command to run.
+        cwd : Optional[str]
+            Work directory to run the script at.
+
+        Returns
+        -------
+        return_code : int
+            Return code of the script. 0 if sucdessful.
```
bpkroth, line 184:
```suggestion
            Return code of the script. 0 if successful.
```

mlos_bench/mlos_bench/environments/script_env.py
```diff
@@ -0,0 +1,91 @@
...
+            if self._script_params_rename:
+                # Only rename specified - use it.
+                rename = self._script_params_rename.copy()
+            else:
+                # Neither `script_params` nor rename are specified - use all params.
+                rename = {key: self._RE_INVALID.sub("_", key) for key in self._params}
+        else:
+            # Use `script_params` and rename if specified.
+            rename = {key: self._RE_INVALID.sub("_", key) for key in self._script_params}
+            rename.update(self._script_params_rename)
```
bpkroth, line 89: I think listing a param in both `shell_params` and `shell_params_rename` is a config mistake, no?
Maybe prevent that?

bpkroth: One other feature I could see being useful: regex based param names.
E.g. expose all params that match `^MLOS_[a-zA-Z0-9_]+$`
Could be a reasonable way to shorthand not explictly listing all args/settings names.
In that case the order of preference should probably go:
- regex
- explicit list overrides regex matches
- param map overrides regex matches
- any overlap between explicit list and param map are an error

mlos_bench/mlos_bench/services/remote/azure/azure_services.py
```diff
@@ -612,7 +620,7 @@ def remote_exec(self, script: Iterable[str], params: dict) -> Tuple[Status, dict
         """
         config = merge_parameters(
             dest=self.config.copy(),
-            source=params,
+            source=config,
```
bpkroth, line 623: Bug?  Tests?

bpkroth: Nm, just variable rename.

mlos_bench/mlos_bench/services/types/remote_exec_type.py
bpkroth: https://github.com/motus/MLOS/pull/5

bpkroth: I didn't quite match your naming I guess.  Feel free to adjust.  Thanks!

mlos_bench/mlos_bench/environments/script_env.py
bpkroth: I left some comments about this in the schema files here: https://github.com/motus/MLOS/pull/5

mlos_bench/mlos_bench/services/types/remote_exec_type.py
motus: sure, thank you!

mlos_bench/mlos_bench/environments/script_env.py
motus: I think it is ok to have the same parameter in both `shell_params` and `shell_params_rename`. this way, if the user wants to keep the registry of all parameters that we pass to the script environment, we give them this option. I am also hesitant about the sophisticated renaming rules. I don't think the user should pass too many parameters to the environment anyway, so just having a sane default rule + explicit manual override should be sufficient

bpkroth: On the later point I am fine.
In general though, I am worried about config behavior that requires reading the code to understand and prefer to give early feedback of potential issues as often as possible.
Anyways, it's fine for now.

mlos_bench/mlos_bench/config/schemas/environments/common-environment-subschemas.json
```diff
@@ -82,6 +87,48 @@
...
+
+        "shell_param_type": {
+            "description": "All parameters must be valid shell variable names, else use shell_params_mapping.",
+            "type": "string",
+            "pattern": "^[a-zA-Z_][a-zA-Z0-9_]+$"
+        },
+        "shell_params_config": {
+            "type": "object",
+            "properties": {
+                "shell_params_match": {
```
bpkroth, line 99: @motus, since this isn't supported, we should remove it.

mlos_bench/mlos_bench/environments/script_env.py
```diff
@@ -0,0 +1,99 @@
...
+        """
+        super().__init__(name=name, config=config, global_config=global_config,
+                         tunables=tunables, service=service)
+
+        self._script_setup = self.config.get("setup")
+        self._script_run = self.config.get("run")
+        self._script_teardown = self.config.get("teardown")
+
+        self._script_params: Optional[Iterable[str]] = self.config.get("script_params")
+        self._script_params_rename: Dict[str, str] = self.config.get("script_params_rename", {})
```
bpkroth, line 69: These also don't look like they match the schema names.
The end to end test script I mentioned would probably help catch this.

mlos_bench/mlos_bench/config/schemas/environments/common-environment-subschemas.json
bpkroth: Although an example of ones that might be handy to include: `AZURE_*` environment variables.  In fact these could be setup in the shell environment, slurped into the `mlos_bench` globals automatically, and then passed along to the remote/local env for scripts which would be super convenient.

PR #384, opened 2023-06-06
Json schema validation opt-out option for dev cycle

bpkroth:
This PR allows skipping json schema validation during tight dev cycle loops to avoid having to synchronize changes in multiple places while getting things working initially in the code.

To enable it one simply does the following:

```sh
# Export an environment variable.
export MLOS_BENCH_SKIP_SCHEMA_VALIDATION=true
# Run the mlos_bench script as usual.
mlos_bench --config ...
```

Alternatively, this can be added to the `.env` file and the devcontainer should pick it up on launch.

Caveats:
- It is intended for local use only and expected that once the python side is ready, the json schema side will be fixed.
- It probably doesn't work at all for tests atm, since some of the tests are negative tests (check for failures) and so would fail (there might be a way to fix that in all of the tests as well).
- Since this only operates on the mlos_bench code, it does nothing for the schema validator of the editor of the configs, and so they may show errors during the process (this is normal for schema development anyways and somewhat helpful often).
  - Note: this is more complicated when the configs are in a separate repo since then the only way to instruct the editor of what schema to use in the first place is the `"$schema": "https://host/path/url/to/schema.json"` tag, in which case it will examine the schema file at that URL.
    In that case there is an option:
    Instruct the local editor to use the locally developed schema files, though absolute paths might be required.  As this is supposed to be a temporary hack, that seems an annoying, but acceptable workaround.  See `.vscode/settings.json` in this repo for an example.

### DIFF
--- a/mlos_bench/mlos_bench/config/schemas/cli/globals-schema.json
+++ b/mlos_bench/mlos_bench/config/schemas/cli/globals-schema.json
@@ -13,13 +13,23 @@
     },
     "patternProperties": {
         "^[^$]+$": {
-            "$comment": "Global configs are either strings, numbers, booleans or nulls.",
+            "$comment": "Global configs are either strings, numbers, booleans, or nulls, or an array of the same.",
             "type": [
                 "string",
                 "number",
                 "boolean",
-                "null"
-            ]
+                "null",
+                "array"
+            ],
+            "items": {
+                "$comment": "The value of any const_arg list can be a string, number, boolean, or null.",
+                "type": [
+                    "string",
+                    "number",
+                    "boolean",
+                    "null"
+                ]
+            }
         }
     },
     "not": {
--- a/mlos_bench/mlos_bench/config/schemas/environments/common-environment-subschemas.json
+++ b/mlos_bench/mlos_bench/config/schemas/environments/common-environment-subschemas.json
@@ -28,13 +28,23 @@
                     "description": "Default argument value constants to use for the Environment when not overridden by globals.",
                     "type": "object",
                     "additionalProperties": {
-                        "$comment": "The value of any const_arg can be a string, number, boolean, or null.",
+                        "$comment": "The value of any const_arg can be a string, number, boolean, or null, or an array of the same.",
                         "type": [
                             "string",
                             "number",
                             "boolean",
-                            "null"
-                        ]
+                            "null",
+                            "array"
+                        ],
+                        "items": {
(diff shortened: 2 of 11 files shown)


mlos_bench/mlos_bench/config/schemas/config_schemas.py
```diff
@@ -98,5 +98,8 @@ def validate(self, config: dict) -> None:
         jsonschema.exceptions.ValidationError
         jsonschema.exceptions.SchemaError
         """
-        resolver: jsonschema.RefResolver = jsonschema.RefResolver.from_schema(self.schema, store=SCHEMA_STORE)
-        jsonschema.validate(instance=config, schema=self.schema, resolver=resolver)
+        # Allow skipping schema validation for tight dev cycle changes.
+        # Note: this may cause pytest to fail if it's expecting exceptions to be raised for invalid configs.
+        if environ.get('MLOS_BENCH_SKIP_SCHEMA_VALIDATION', 'false').lower() not in {'true', 'y', 'yes', '1'}:
```
motus, line 103:
```suggestion
        if environ.get('MLOS_BENCH_SKIP_SCHEMA_VALIDATION', 'false').lower() not in {'true', 'y', 'yes', 'on', '1'}:
```

motus (approved):

mlos_bench/mlos_bench/config/schemas/config_schemas.py
```diff
@@ -98,5 +98,8 @@ def validate(self, config: dict) -> None:
         jsonschema.exceptions.ValidationError
         jsonschema.exceptions.SchemaError
         """
-        resolver: jsonschema.RefResolver = jsonschema.RefResolver.from_schema(self.schema, store=SCHEMA_STORE)
-        jsonschema.validate(instance=config, schema=self.schema, resolver=resolver)
+        # Allow skipping schema validation for tight dev cycle changes.
+        # Note: this may cause pytest to fail if it's expecting exceptions to be raised for invalid configs.
+        if environ.get('MLOS_BENCH_SKIP_SCHEMA_VALIDATION', 'false').lower() not in {'true', 'y', 'yes', 'on', '1'}:
```
bpkroth, line 103: Potential future work improvement: fetching environment variables is somewhat expensive.  It might make sense to do this at script load time to cache the value and only check a boolean after that.

motus: I think we can do it later. Fetching an environment variable is still cheaper than validating a JSON

PR #387, opened 2023-06-09
Use ._get_env_params() for upload/download paths; issue a warning when json validation is off

motus:
also: do not require for at least one of setup/run/teardown sections to be present - it's ok to have none, e.g., when uploading/downloading files


mlos_bench/mlos_bench/config/schemas/config_schemas.py
```diff
@@ -100,6 +105,8 @@ def validate(self, config: dict) -> None:
         """
         # Allow skipping schema validation for tight dev cycle changes.
         # Note: this may cause pytest to fail if it's expecting exceptions to be raised for invalid configs.
```
bpkroth, line 107: Move this comment up towards the env var along with a "See: validate()" or some such?

mlos_bench/mlos_bench/environments/script_env.py
```diff
@@ -68,11 +68,6 @@ def __init__(self,
         self._script_params: Optional[Iterable[str]] = self.config.get("script_params")
         self._script_params_rename: Dict[str, str] = self.config.get("script_params_rename", {})
 
-        if self._script_setup is None and \
-           self._script_run is None and \
-           self._script_teardown is None:
-            raise ValueError("At least one of {setup, run, teardown} must be present")
```
bpkroth, line 74: I feel like the underlying envs that use this class should require either one of these or one of the read_results_file/dump_params.
Otherwise, the envs don't do anything and it's not really useful to have them and we should give earlier feedback on that.

bpkroth:
There are some relevant schema changes in #389 if you wanted to cherry-pick some of them now, else get that sorted out after this.

Could still use some eyes/help on #382 though.

mlos_bench/mlos_bench/environments/script_env.py
motus: In our MySQL setup there are two use cases:
1. `LocalFileShareEnv` with only the `upload` section -- simply upload some scripts and other files to the shared storage; no need to run any scripts locally
2. `LocalFileShareEnv` with `download` and `read_results_file` -- download file from remote storage and ingest it immediately without any local processing

In `RemoteEnv` and `LocalEnv` we still might want to enforce having one of setup/run/teardown sections to be present, but since `LocalFileShareEnv` is inherited from `LocalEnv`, it complicates things a bit. I suggest that we merge this PR into master and live with relaxed constraints for a day or two so we can run our experiments, and then issue another PR that checks for all possible combinations of these config parameters

bpkroth: Ack.  I'll take that up in #389

bpkroth (approved):
Good for now.

bpkroth:
Looks to need fixups from #391 first.

PR #388, opened 2023-06-12
SMAC Optimizer: CI testing

bpkroth:
Draft version to help work out CI issues for #341 

For some reason adding the SMAC optimizer has been causing increased devcontainer rebuild times, though it's not clear why yet.


bpkroth:
Merged into #294

Issue #386, opened 2023-06-09
FIxup bugs from #383 - Add end to end environment variable test script


bpkroth:
So we don't loose track of it - need a followup PR to fixup bugs introduced in #383:

These also don't look like they match the schema names.
- [x] The end to end test script I mentioned would probably help catch this.

_Originally posted by @bpkroth in https://github.com/microsoft/MLOS/pull/383#discussion_r1222268700_

PR #389, opened 2023-06-12
Schema and other fixups

bpkroth:
Follow on to #382 and #387
Also addresses some of the comments from #386 (still missing an end-to-end test though).

- ~[ ] Per, #387, we should enable more complex requirements checking of what options must be required for LocalEnv vs RemoteEnv vs LocalFileShareEnv~
  Skipping this for now.


scripts/generate-azure-credentials-config.sh
bpkroth, line 1: Make this script executable for direct usage.

motus (commented):
Looks good, although I feel a bit uneasy about the number of files that we have to change every time there is an update to the schema

motus (approved):

PR #406, opened 2023-06-21
small fix to make SMAC __del__/cleanup work

motus:
We need to define `self._temp_output_directory` at the top of the constructor because it is used later in `__del__()`/`cleanup()`


bpkroth (approved):
LGTM
Any tests?

PR #407, opened 2023-06-21
Impute missing values when restoring configs from earlier experiments/trials

motus:
This is the first step towards mergeable experiments. Replace missing values with the defaults when restoring trail data from storage.


bpkroth (approved):
LGTM, but tests?

PR #409, opened 2023-06-21
Do not inherit any tunables if `tunable_params` not specified.

motus:
(Old default was to use ALL tunables). @bpkroth you were right about this one

### DIFF
--- a/mlos_bench/mlos_bench/tests/conftest.py
+++ b/mlos_bench/mlos_bench/tests/conftest.py
@@ -111,7 +111,7 @@ def mock_env(tunable_groups: TunableGroups) -> MockEnv:
     return MockEnv(
         name="Test Env",
         config={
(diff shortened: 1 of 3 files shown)


mlos_bench/mlos_bench/tests/conftest.py
```diff
@@ -111,7 +111,7 @@ def mock_env(tunable_groups: TunableGroups) -> MockEnv:
     return MockEnv(
         name="Test Env",
         config={
-            "tunable_groups": ["provision", "boot", "kernel"],
+            "tunable_params": ["provision", "boot", "kernel"],
```
bpkroth, line 114: Methinks #414 would probably have caught that mistake.

bpkroth (approved):
Additional tests to check that no params are exported when non are specified?

PR #411, opened 2023-06-22
introduce the "meta" field in Tunable class

motus:
We'll need it for many things, but right now the plan is to store the units and scalers for some tunable values and use it for MySQL, e.g., to produce config values like `innodb_buffer_pool_size = 128M`

### DIFF
--- a/mlos_bench/mlos_bench/tunables/tunable.py
+++ b/mlos_bench/mlos_bench/tunables/tunable.py
@@ -9,7 +9,7 @@ import copy
 import collections
 import logging
 
-from typing import Dict, List, Optional, Sequence, Tuple, Type, TypedDict, Union
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Type, TypedDict, Union
 
 _LOG = logging.getLogger(__name__)
 
@@ -33,6 +33,7 @@ class TunableDict(TypedDict, total=False):
     values: Optional[List[Optional[str]]]
     range: Optional[Union[Sequence[int], Sequence[float]]]
     special: Optional[Union[List[int], List[str]]]
+    meta: Dict[str, Any]
 
 
 class Tunable:  # pylint: disable=too-many-instance-attributes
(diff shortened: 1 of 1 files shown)


mlos_bench/mlos_bench/tunables/tunable.py
```diff
@@ -63,6 +64,7 @@ def __init__(self, name: str, config: TunableDict):
         self._description = config.get("description")
         self._default = config["default"]
         self._values = config.get("values")
+        self._meta: Dict[str, Any] = config.get("meta", {})
```
bpkroth, line 67:
```suggestion
        self._meta: Dict[str, TunableValue] = config.get("meta", {})
```

Do you forsee a case where it should be more than this?

mlos_bench/mlos_bench/tunables/tunable.py
```diff
@@ -376,3 +378,11 @@ def categories(self) -> List[Optional[str]]:
...
         assert self._values is not None
         return self._values
+
+    @property
+    def meta(self) -> Dict[str, Any]:
+        """
+        Get the tunable's metadata. This is a free-form dictionary that can be used to
+        store any additional information about the tunable (e.g., the unit information).
+        """
+        return self._meta
```
bpkroth, line 388: What about during the conversion to ConfigSpace?

motus: good point! we should add it to the `meta` field there. Let's do it in the next PR, along with the unit tests

mlos_bench/mlos_bench/tunables/tunable.py
motus: actually, yes - e.g., something along the line of:
```json
"multiplier_suffix": {
    1024: "K",
    1048576: "M",
    1073741824: "G"
}
```
or, say,
```json
"constraints": {
    "less_than": ["innodb_io_capacity_max"]
}
```
etc.

mlos_bench/mlos_bench/tunables/tunable.py
bpkroth: Please make Issues to track the future PRs so we don't lose track of them.  Thanks!

bpkroth (approved):

mlos_bench/mlos_bench/tunables/tunable.py
motus: Created issue #419

PR #412, opened 2023-06-22
Pass Inf to the optimizer if the trial had failed

motus:


mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
```diff
@@ -59,16 +59,17 @@ def bulk_register(self, configs: Sequence[dict], scores: Sequence[Optional[float
...
+            df_status = pd.Series(status)
+            df_scores[df_status != Status.SUCCEEDED] = float("inf")
+            df_status_completed = df_status.apply(Status.is_completed)
+            df_configs = df_configs[df_status_completed]
+            df_scores = df_scores[df_status_completed]
         # External data can have incorrect types (e.g., all strings).
         for (tunable, _group) in self._tunables:
             df_configs[tunable.name] = df_configs[tunable.name].astype(tunable.dtype)
-        self._opt.register(df_configs, df_scores)
+        self._opt.register(df_configs, df_scores * self._opt_sign)
```
bpkroth, line 72: Tests to make sure that we get `-inf` when the task is `maximize`?

bpkroth (approved):
Needs tests, but OK for now.

Per existing science, in the future, we should replace this logic with something that tracks the "worst" seen "real" trial data so far, and submits the bad score as `N * worst_real_seen` instead.

Probably need an Issue for this improvement ...

mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
motus: yes, we need more unit tests to make sure that mlos_core optimizer gets the right values. I'll create an issue for it.

motus: Added more unit tests in #416

mlos_bench/mlos_bench/optimizers/base_optimizer.py
```diff
@@ -152,12 +152,15 @@ def _get_score(self, status: Status, score: Optional[Union[float, Dict[str, floa
...
-        assert score is not None
-        if isinstance(score, dict):
-            score = score[self._opt_target]
-        return float(score) * self._opt_sign
+        if status.is_succeeded:
+            assert score is not None
+            if isinstance(score, dict):
+                score = score[self._opt_target]
+            return float(score) * self._opt_sign
+        assert score is None
```
bpkroth, line 162: 👍

PR #413, opened 2023-06-22
Bulk load the data of failed and timed out trials into the optimizer

motus:
Before that, we loaded only the data from successful runs.
We probably need to merge #412 and #407 before this one

### DIFF
--- a/mlos_bench/mlos_bench/run.py
+++ b/mlos_bench/mlos_bench/run.py
@@ -84,8 +84,8 @@ def _optimize(env: Environment,
         # Load (tunable values, benchmark scores) to warm-up the optimizer.
         # `.load()` returns data from ALL merged-in experiments and attempts
         # to impute the missing tunable values.
-        (configs, scores) = exp.load()
-        opt.bulk_register(configs, scores)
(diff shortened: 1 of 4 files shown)


mlos_bench/mlos_bench/run.py
```diff
@@ -84,8 +84,8 @@ def _optimize(env: Environment,
         # Load (tunable values, benchmark scores) to warm-up the optimizer.
         # `.load()` returns data from ALL merged-in experiments and attempts
         # to impute the missing tunable values.
-        (configs, scores) = exp.load()
-        opt.bulk_register(configs, scores)
+        (configs, scores, status) = exp.load()
+        opt.bulk_register(configs, scores, status)
```
bpkroth, line 88: I think there's a TODO somewhere in `mlos_core_optimizer.py` to handle FAILED status that we should maybe mark down if appropriate.

motus: yep, thank you for reminding me. fixed that, too!

bpkroth (approved):

PR #420, opened 2023-06-23
Move parameter initialization to the top level CLI config

motus:
Summary of changes:
* Introduce parameters `--random_init` (Boolean) and `--random_seed` (int) at the top (CLI) config level
* If `random_init` is `true`, initialize the tunable parameters with random values; otherwise, use the defaults
* If `tunable_values` specified, use that data to override the values of (some) tunables
* Add a Boolean parameter `start_with_defaults` to the Optimizer configs; when `true`, on the first iteration use the values that has been already assigned to the tunables instead of the optimizer's suggestion
* use `start_with_defaults` flag only when there were no prior data, e.g., retrieved from the DB
* Fix the unit tests and JSON schemas accordingly


mlos_bench/mlos_bench/optimizers/base_optimizer.py
```diff
@@ -41,8 +41,8 @@ def __init__(self, tunables: TunableGroups, service: Optional[Service], config:
         self._tunables = tunables
         self._service = service
         self._iter = 1
-        self._use_defaults: bool = bool(strtobool(str(self._config.pop('use_defaults', True))))
-        self._max_iter = int(self._config.pop('max_iterations', 25))
+        self._use_defaults: bool = bool(strtobool(str(self._config.pop('use_defaults', False))))
```
bpkroth, line 44: I think a desirable thing is for us to check the "default" values in the ConfigSpace as the very first experiment Trial.
Leaving this as `True` helps accomplish that

bpkroth: Also, this is maybe just a confusion for me from the PR title, but leaving the defaults for unspecified initialization is then not a "random initialization" - it's pretty explicit in the configs, no?

motus: I've fixed it in the subsequent commits - will mark this conversation resolved for now

mlos_bench/mlos_bench/launcher.py
```diff
@@ -91,12 +92,15 @@ def __init__(self, description: str, long_text: str = ""):
                          " Run `mlos_bench --help` and consult `README.md` for more info.")
         self.root_env_config = self._config_loader.resolve_path(env_path)
 
-        tunable_groups = TunableGroups()    # base tunable groups that all others get build on
         self.environment: Environment = self._config_loader.load_environment(
-            self.root_env_config, tunable_groups, self.global_config, service=self._parent_service)
+            self.root_env_config, TunableGroups(), self.global_config, service=self._parent_service)
+
+        # NOTE: Init tunable values *after* the Environment, but *before* the Optimizer
```
bpkroth, line 98: Why's that?

mlos_bench/mlos_bench/launcher.py
```diff
@@ -142,6 +146,15 @@ def _parse_args(parser: argparse.ArgumentParser) -> Tuple[argparse.Namespace, Li
             help='Path to the storage configuration file.' +
                  ' If omitted, use the ephemeral in-memory SQL storage.')
 
+        parser.add_argument(
+            '--random_init', required=False, default=False,
+            dest='random_init', action='store_true',
+            help='Initialize tunables with random values. (Before applying --tunable_values).')
+
+        parser.add_argument(
+            '--random_seed', required=False, type=int,
```
bpkroth, line 155: Should add this to the cli config schema.

mlos_bench/mlos_bench/launcher.py
```diff
@@ -208,11 +221,18 @@ def _load_config(self,
...
+                             args_tunables: Optional[str]) -> TunableGroups:
         """
-        Load key/value pairs of the tunable values from given JSON files, if any.
+        Initialize the tunables and load key/value pairs of the tunable values
+        from given JSON files, if specified.
         """
         tunables = self.environment.tunable_params
+
+        if random_init:
+            opt = MockOptimizer(tunables=tunables, service=None, config={"seed": seed})
```
bpkroth, line 233: Creating an optimizer and throwing it away just to get random values?
Do you need to pass "random_init" to the MockOptimizer config?

mlos_bench/mlos_bench/optimizers/base_optimizer.py
```diff
@@ -41,8 +41,10 @@ def __init__(self, tunables: TunableGroups, service: Optional[Service], config:
         self._tunables = tunables
         self._service = service
         self._iter = 1
-        self._use_defaults: bool = bool(strtobool(str(self._config.pop('use_defaults', True))))
-        self._max_iter = int(self._config.pop('max_iterations', 25))
+        # If True, suggest the already assigned tunable values for the first iteration:
```
bpkroth, line 44:
```suggestion
        # If True, suggest the "already assigned tunable values" for the first iteration:
```

"already assigned tunable values" means something slightly different than "start_with_defaults".

We should probably document that somewhere for the user.  For instance:
> "start_with_defaults" means initialize the first iteration with the `tunable_params` defaults, and possibly overlay those with any `tunable_values` that were provided.

mlos_bench/mlos_bench/optimizers/base_optimizer.py
```diff
@@ -91,7 +93,11 @@ def bulk_register(self, configs: Sequence[dict], scores: Sequence[Optional[float
             raise ValueError("Numbers of configs and scores do not match.")
         if status is not None and len(configs or []) != len(status or []):
             raise ValueError("Numbers of configs and status values do not match.")
-        return bool(configs and scores)
+        has_data = bool(configs and scores)
+        if has_data and self._start_with_defaults:
```
bpkroth, line 97: Why not move that to `suggest()`?

bpkroth: e.g. increase `self.iter` - the optimizer has been trained with that many samples already after all.

mlos_bench/mlos_bench/tests/config/schemas/optimizers/test-cases/bad/invalid/one_shot_bad_tunable_values_type_1.jsonc
bpkroth, line 1: Do we have a version like this in the top level now instead?

mlos_bench/mlos_bench/tests/optimizers/conftest.py
```diff
@@ -24,6 +41,7 @@ def mock_opt(tunable_groups: TunableGroups) -> MockOptimizer:
         config={
             "minimize": "score",
             "max_iterations": 5,
+            "start_with_defaults": True,
```
bpkroth, line 44: Necessary?  Shouldn't that be the default?

mlos_bench/mlos_bench/tests/optimizers/mock_opt_test.py
```diff
@@ -47,6 +41,21 @@ def mock_configurations() -> list:
...
+    """
+    A list of 2-tuples of (tunable_values, score) to test the optimizers.
+    """
+    return [
+        ({
+            "vmSize": "Standard_B4ms",
+            "idle": "halt",
+            "kernel_sched_migration_cost_ns": -1,
+            "kernel_sched_latency_ns": 2000000,
+        }, 88.88),
```
bpkroth, line 55: Any way to pull the defaults out of the actual optimizer.  Worries me that that's defined elsewhere and there will be unrelated files to edit when one of those changes, but maybe no way to avoid it.

mlos_bench/mlos_bench/tests/optimizers/mock_opt_test.py
```diff
@@ -71,6 +80,15 @@ def test_mock_optimizer(mock_opt: MockOptimizer, mock_configurations: list) -> N
     assert score == pytest.approx(66.66, 0.01)
 
 
+def test_mock_optimizer_no_defaults(mock_opt_no_defaults: MockOptimizer,
+                                    mock_configurations_no_defaults: list) -> None:
+    """
+    Make sure that mock optimizer produces consistent suggestions.
+    """
+    score = _optimize(mock_opt_no_defaults, mock_configurations_no_defaults)
+    assert score == pytest.approx(66.66, 0.01)
```
bpkroth, line 89: Is it expected that these still come up with the same score?

bpkroth: I guess my concern is that right now there isn't a good test that I've seen so far that `start_with_defaults` actually returns the defaults for the first iteration (but not when bulk registered).
Maybe we just need explicit tests for those cases.

mlos_bench/mlos_bench/optimizers/base_optimizer.py
bpkroth: Is the issue with interacting with how the base classes and abstract methods interact?

I feel like we should do this regardless though.
> e.g. increase self.iter - the optimizer has been trained with that many samples already after all.

mlos_bench/mlos_bench/launcher.py
motus: done

mlos_bench/mlos_bench/launcher.py
motus: a line just above this one is `tunable_groups = TunableGroups()` and we don't use that variable anywhere else

mlos_bench/mlos_bench/optimizers/base_optimizer.py
motus: done

mlos_bench/mlos_bench/tests/optimizers/conftest.py
motus: made `True` teh default

mlos_bench/mlos_bench/launcher.py
motus: that's the easiest way to do things IMO. Technically, we can factor out that random initialization and use it in both the MockOptimizer and here, but maybe I'll do it in the next PR.

motus: also, no, we don't need to pass `random_init` to `MockOptimizer` - it is always random (except for the first iteration when `start_with_defaults` is True)

mlos_bench/mlos_bench/optimizers/base_optimizer.py
motus: right now when we specify `max_iterations=N` in the config that means "train the optimizer for N more iterations", regardless of how many iterations we had ran before. I think it is a reasonable approach. It also makes sense when we implement merging of the experiments - then we just say, e.g., "merge 100 trials from this experiment and 50 trials from that experiment and run the optimizer for additional 50 iterations"

mlos_bench/mlos_bench/tests/optimizers/mock_opt_test.py
motus: Technically, we can extract them from the `tunable_groups` fixture, but would rather keep this code as is

mlos_bench/mlos_bench/tests/optimizers/mock_opt_test.py
motus: why, that's exactly the test - we have two instances of MockOptimizer, one for each value of `start_with_defaults`, and we check what values they suggest at the first iteration

mlos_bench/mlos_bench/optimizers/base_optimizer.py
motus: Anyway, I would keep the existing logic for now - we can always change it later, and now we have lots of other more important things to do

mlos_bench/mlos_bench/tests/config/schemas/optimizers/test-cases/bad/invalid/one_shot_bad_tunable_values_type_1.jsonc
motus: at the top level we always include the files with values, so there is no need for these checks I guess

mlos_bench/mlos_bench/config/schemas/cli/cli-schema.json
```diff
@@ -62,6 +62,16 @@
...
         },
 
+        "random_init": {
+            "description": "If true, initialize the tunables with random values; otherwise, use the defaults.",
+            "type": "boolean"
+        },
+
+        "random_seed": {
+            "description": "Random seed to use in conjunction with random_init.",
+            "type": "number"
```
bpkroth, line 72: I think only ints are supported, right?

bpkroth: Can also add tests to the "full" cli example to make sure we don't accidentally drop it in the future.

mlos_bench/mlos_bench/optimizers/base_optimizer.py
bpkroth: Kk, we should make it more clear that it's "N *more* iterations" somewhere. Maybe just update the description in the schema file for now.

mlos_bench/mlos_bench/tests/optimizers/mock_opt_test.py
bpkroth: I missed the details in the `_optimize()` call (which you didn't change) already does an `assert` that the `suggest` value returned matches those in the listed mock_configs.  I thought we were only checking that the got some reasonable looking score when registering those.

But do we have one that tests the logic when bulk_registering is in effect?

mlos_bench/mlos_bench/launcher.py
bpkroth: I saw that, I was just curious if there was something about that logic that's important to work that way.  Do we manipulate the params in the Environment somehow?

mlos_bench/mlos_bench/optimizers/base_optimizer.py
motus: updated the description in the json schema

mlos_bench/mlos_bench/config/schemas/cli/cli-schema.json
motus: done

mlos_bench/mlos_bench/tests/optimizers/mock_opt_test.py
motus: good idea! just added a test for it

mlos_bench/mlos_bench/config/schemas/optimizers/optimizer-schema.json
```diff
@@ -24,43 +24,23 @@
                     "type": "string"
                 },
                 "max_iterations": {
-                    "description": "The maximum number of iterations to run.",
+                    "description": "The maximum number of additional iterations to run when we launch the app.",
```
bpkroth, line 27:
```suggestion
                    "description": "The maximum number of additional (in the case of merging experiment data or resuming experiments) iterations to run when we launch the app.",
```

bpkroth (approved):

PR #421, opened 2023-06-23
Make all `Status.is_*` functions instead of properties.

motus:
this is consistent with other functions like, e.g., `str.isalpha()`, but, more importantly, allows us to use them applicatively, e.g., `df.score[df.status.apply(Status.is_completed)]`


bpkroth (approved):
LGTM

PR #425, opened 2023-06-28
mypy issue: score can be None in Experiment.load() results

motus:


mlos_bench/mlos_bench/storage/sql/experiment.py
```diff
@@ -89,7 +89,7 @@ def merge(self, experiment_ids: List[str]) -> None:
         _LOG.info("Merge: %s <- %s", self._experiment_id, experiment_ids)
         raise NotImplementedError()
 
-    def load(self, opt_target: Optional[str] = None) -> Tuple[List[dict], List[float], List[Status]]:
+    def load(self, opt_target: Optional[str] = None) -> Tuple[List[dict], List[Optional[float]], List[Status]]:
```
bpkroth, line 92: Need any tests?

bpkroth (approved):

mlos_bench/mlos_bench/storage/sql/experiment.py
motus: we do have tests, and this is just a small issue in type annotation.. 
I wonder why it was not caught by mypy earlier, though

PR #428, opened 2023-06-29
Remove Emukit Support to avoid GPy Issues

### LINKED ISSUE #426: Remove Emukit support to avoid GPy requirement
GPy and its dependencies (`nose`) are undermaintained and getting it up to date is more than I have time atm:
https://github.com/SheffieldML/GPy/pull/1020#issuecomment-1613489576

This is causing issues with some of our other dependency mismatches as well.
For now, I'm going to look into removing Emukit support since it's the only thing that depends on GPy.

bpkroth:
Removes Emukit to avoid GPy issues.  Closes #426
Switches most things to SMAC instead.


motus (approved):
ah, the pleasure of deleting code

PR #435, opened 2023-07-05
Pass global config to Service constructor

motus:
Turns out, it is not enough to substitute the top-level Service config parameters with the globals - sometimes the placeholders are not at the top level (e.g., `deploymentTemplateParameters` of `AzureVMService`) so we need to pass the `global_config` explicitly into each Service constructor.

This update allows us to collect all configurable parameters from all services and environments in a single config


mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
```diff
@@ -24,7 +24,7 @@ class AzureAuthService(Service, SupportsAuth):
 
     _REQ_INTERVAL = 300   # = 5 min
 
-    def __init__(self, config: dict, parent: Service):
+    def __init__(self, config: dict, global_config: dict, parent: Service):
```
bpkroth, line 27: Not the end of the world probably, but we switch at some point from these being optional and declared with types to just `dict`.  Same for `parent`.

bpkroth: Would be good to be consistent or document why not.

mlos_bench/mlos_bench/services/remote/azure/azure_services.py
```diff
@@ -152,7 +154,7 @@ def __init__(self, config: dict, parent: Service):
         self._deploy_template = template
 
         self._deploy_params = merge_parameters(
-            dest=config['deploymentTemplateParameters'].copy(), source=self.config)
+            dest=config['deploymentTemplateParameters'].copy(), source=global_config)
```
bpkroth, line 157: Can you help me understand why we're merging from global configs instead of config here?  Seems to be the crux of the whole PR, but I didn't fully follow from the description.  Any chance we can have a workable example to look at?  Even if just in test form to see how the change works differently?

motus: you're right, that's point of this PR.

In our current implementation, we replace the values of service parameters with the globals *before* we instantiate the Service class. In other words, we call `merge_parameters(dest=service_config, source=global_config)` and *then* instantiate the service with that `service_config`. That works great for all services which configs are flat dicts of key/value pairs; `AzuerVMService` config, however, has an extra level of `deploymentTemplateParameters`. oops.

I've hacked around this issue by copying the parameters into it from the top level of `AzureVMService` config (hence `source=self.config`). That works, but it requires us to duplicate every parameter that can be possibly pushed from the globals into `deploymentTemplateParameters` at the top level of the service config. As a result, our config right now looks like this:
```jsonc
{
    "class": "mlos_bench.services.remote.azure.AzureVMService",

    "config": {

        "subscription": "PLACEHOLDER; AZURE SUBSCRIPTION ID",
        "resourceGroup": "PLACEHOLDER; e.g., os-autotune",
        "deploymentName": "PLACEHOLDER; e.g., mlos-2vms",

        "deploymentTemplatePath": "services/azure/arm-templates/ARM-client-server-2vms.jsonc",

        // Every parameter in deploymentTemplateParameters must be also present at the top level
        "experimentId": "PLACEHOLDER; e.g., MysqlBench2VMs",
        "storageAccountName": "PLACEHOLDER; e.g., osatsharedstorage",
        "storageFileShareName": "PLACEHOLDER; e.g., os-autotune-file-share",
        "vmName_server": "PLACEHOLDER; e.g., mlos-server-vm",
        "vmName_client": "PLACEHOLDER; e.g., mlos-client-vm",
        "location": "PLACEHOLDER; e.g., westus2",
        "ubuntuOSVersion": "PLACEHOLDER; e.g., 18.04-LTS",

        "deploymentTemplateParameters": {
            // Duplicate all placeholder parameters here at the level above
            "experimentId": "PLACEHOLDER; e.g., MysqlBench2VMs",
            "storageAccountName": "PLACEHOLDER; e.g., osatsharedstorage",
            "storageFileShareName": "PLACEHOLDER; e.g., os-autotune-file-share",
            "vmName_server": "PLACEHOLDER; e.g., mlos-server-vm",
            "vmName_client": "PLACEHOLDER; e.g., mlos-client-vm",
            "location": "PLACEHOLDER; e.g., westus2",
            "ubuntuOSVersion": "PLACEHOLDER; e.g., 18.04-LTS"
        }
    }
}
```
That is already confusing for one or two parameters, and it gets really ugly really fast when we have more of them. I thought that pushing `global_config` to every Service constructor is a smaller evil, and it is not visible to the users :smile:

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
motus: You are right. This PR is a good opportunity to make all constructors consistent - at least for the Service class hierarchy. I'll push an update soon and let you know.

mlos_bench/mlos_bench/services/remote/azure/azure_services.py
bpkroth: I think it makes sense.  Can you list what the new config would look like?

bpkroth (approved):
LGTM

mlos_bench/mlos_bench/services/remote/azure/azure_services.py
motus: sure! it becomes simply
```jsonc
{
    "class": "mlos_bench.services.remote.azure.AzureVMService",

    "config": {

        "subscription": "PLACEHOLDER; AZURE SUBSCRIPTION ID",
        "resourceGroup": "PLACEHOLDER; e.g., os-autotune",
        "deploymentName": "PLACEHOLDER; e.g., mlos-2vms",

        "deploymentTemplatePath": "services/azure/arm-templates/ARM-client-server-2vms.jsonc",

        "deploymentTemplateParameters": {

            "experimentId": "PLACEHOLDER; e.g., MysqlBench2VMs",

            "storageAccountName": "PLACEHOLDER; e.g., osatsharedstorage",
            "storageFileShareName": "PLACEHOLDER; e.g., os-autotune-file-share",

            "vmName_server": "PLACEHOLDER; e.g., mlos-server-vm",
            "vmName_client": "PLACEHOLDER; e.g., mlos-client-vm",

            "location": "PLACEHOLDER; e.g., westus2",

            "ubuntuOSVersion": "PLACEHOLDER; e.g., 18.04-LTS"
        }
    }
}
```

PR #436, opened 2023-07-05
Enable python 3.11 support

### LINKED ISSUE #293: Dependencies do not support Python 3.11
https://github.com/SheffieldML/GPy/issues/998

Currently we've worked around this by pinning the floating python version for the `mlos_core` conda env below 3.11:
https://github.com/microsoft/MLOS/blob/1cd609263f5e84008cb75e09b4c8d8355bf84185/conda-envs/mlos_core.yml#L25

Options could be to drop GPy support (which is kinda slow anyways), but Emukit currently relies on it.

When we have SMAC support (#294) we could just drop Emukit as well.

bpkroth:
Closes #293

### DIFF
--- a/.github/workflows/build-dist-test.ps1
+++ b/.github/workflows/build-dist-test.ps1
@@ -57,11 +57,6 @@ if ($LASTEXITCODE -ne 0) {
 }
 # Add a few extras we have to pull in from conda-forge.
 # See Also: mlos-windows.yml
-conda install -y -v -n mlos-dist-test conda-forge::GPy
-if ($LASTEXITCODE -ne 0) {
-    Write-Error "Failed to install GPy dependencies."
-    exit $LASTEXITCODE
-}
 # FIXME: conda on Windows doesn't appear to respect ">=0.9.0" as a version constraint despite various quoting tweaks.
 conda install -y -v -n mlos-dist-test -c conda-forge 'pyrfr=0.9'
 if ($LASTEXITCODE -ne 0) {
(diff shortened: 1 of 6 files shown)


bpkroth:
Hmm, seems pyrfr is still lacking 3.11 support: https://github.com/automl/random_forest_run/issues/75

bpkroth:
Or maybe not.  Seems to pass in the python 3.11 conda check.  Just the devcontainer fails.  Will need to debug further later.

.github/workflows/build-dist-test.ps1
bpkroth, line 1: CLRF auto conversions.

Basically just removed the GPy install portion.

conda-envs/mlos-3.11.yml
```diff
@@ -17,6 +17,7 @@ dependencies:
   - nb_conda_kernels
   - matplotlib
   - seaborn
+  - swig
```
bpkroth, line 20: swig is necessary for pyrfr in 3.11

mlos_bench/setup.py
```diff
@@ -76,7 +76,7 @@
         'mlos-core==' + _VERSION,
         'requests',
         'json5',
-        'jsonschema',
+        'jsonschema<4.18.0',    # FIXME: temporarily pinning to avoid breaking changes in 4.18.0
```
bpkroth, line 79: This was causing CI issues due to the recently updated 4.18 release of jsonschema.  For now I pinned it back and will address it in #437

bpkroth: PR for this is in #438 which should get merged first.

bpkroth:
#438 should get merged first and then this is ready

motus (approved):

PR #440, opened 2023-07-10
Bugfix: update the env tunable values in setup, do not mutate the parent when loading the tunables

motus:
Summary of changes:
* Do not mutate the parent collection of tunables when loading them from a file.
* Fix the confusing behavior of the environment: we did not assign the new values to the environment's tunables after `.setup()`. That is ok most of the time as we never use the tunables alone when setting up the environment (we merge them into `self._params` along with `const_args` and globals), but it is still a bit confusing to not update the tunables on setup.
* Add unit tests to check the new behavior.
* Add logging statements to debug the issue

### DIFF
--- a/mlos_bench/mlos_bench/environments/base_environment.py
+++ b/mlos_bench/mlos_bench/environments/base_environment.py
@@ -224,7 +224,10 @@ class Environment(metaclass=abc.ABCMeta):
         """
         _LOG.info("Setup %s :: %s", self, tunables)
         assert isinstance(tunables, TunableGroups)
-
+        # Assign new values to the environment's tunable parameters:
+        groups = list(self._tunable_params.get_covariant_group_names())
+        self._tunable_params.assign(tunables.get_param_values(groups))
+        # Combine tunables, const_args, and global config into `self._params`:
         self._params = self._combine_tunables(tunables)
         merge_parameters(dest=self._params, source=global_config)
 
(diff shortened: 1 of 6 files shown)


mlos_bench/mlos_bench/environments/base_environment.py
```diff
@@ -224,7 +224,10 @@ def setup(self, tunables: TunableGroups, global_config: Optional[dict] = None) -
         """
         _LOG.info("Setup %s :: %s", self, tunables)
         assert isinstance(tunables, TunableGroups)
-
+        # Assign new values to the environment's tunable parameters:
+        groups = list(self._tunable_params.get_covariant_group_names())
+        self._tunable_params.assign(tunables.get_param_values(groups))
+        # Combine tunables, const_args, and global config into `self._params`:
         self._params = self._combine_tunables(tunables)
```
bpkroth, line 231: Doesn't `_combine_tunables` already do this, or am I missing something?

bpkroth: Nm, I think I see the issue.

motus: no, `_combine_tunables()` does not mutate the environment, and I wanted to have that mutation logic explicit, so I put it nere

bpkroth (approved):
LGTM

PR #445, opened 2023-07-12
Dump the tunables' metdata to JSON file so that the `LocalEnv` scripts can use it

motus:
In the subsequent PRs, we will port the tunables' metadata and Linux kernel configuration scripts that use it from our private repo.
A PR with unit tests will also follow (turns out, we don't have *any* tests for `LocalEnv` atm).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -117,5 +117,9 @@
         "editor.defaultFormatter": "ms-python.autopep8",
         "editor.formatOnSave": true,
         "editor.formatOnSaveMode": "modifications"
-    }
+    },
+    "python.testing.pytestArgs": [
+        "."
+    ],
+    "python.testing.unittestEnabled": false
 }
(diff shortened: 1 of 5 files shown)


.vscode/settings.json
```diff
@@ -117,5 +117,9 @@
         "editor.defaultFormatter": "ms-python.autopep8",
         "editor.formatOnSave": true,
         "editor.formatOnSaveMode": "modifications"
-    }
+    },
+    "python.testing.pytestArgs": [
+        "."
+    ],
```
bpkroth, line 123: What's this solving?  It should already be handled in `setup.cfg`:

```dosini
[tool:pytest]
...
testpaths = mlos_core mlos_bench
```

bpkroth (approved):
LGTM

.vscode/settings.json
motus: oops, I just did not notice that stuff got committed. I will roll back this change now

PR #446, opened 2023-07-13
make globals, config_path, and tunable_values complement each other in CLI config and command line options

motus:
In our current implementation, command-line options like `--config_path` supersede the corresponding parameters in the CLI config file. For parameters like `globals`, `config_path`, and `tunable_values`, i.e., for lists of files and directories, such behavior is confusing and inconvenient. It would be much better for JSON and command line to complement each other - i.e., to *merge* the data from both sources. Note that the order of merging is important, since the command line parameters should still have priority to JSON


bpkroth (approved):
Yep, found the same thing with some recent hacking.  Thanks!

PR #447, opened 2023-07-14
Add the ability to re-run the benchmark for configurations already in DB

motus:
Introduce `--configId` command line option, e.g., 
`mlos_bench --config config/cli/cli-azure-mysql-bench.jsonc --globals experiment_mysql-kernel-hp.jsonc --configId 377`

Also, do not preload the existing results or resume pending trials if we just want to run a single benchmark.


bpkroth (approved):
Probably needs more tests, but LGTM atm.

PR #448, opened 2023-07-14
Add "meta" to the tunables' JSON schema

motus:


mlos_bench/mlos_bench/config/schemas/tunables/tunable-params-schema.json
```diff
@@ -10,6 +10,10 @@
                 "description": {
                     "type": "string"
                 },
+                "meta": {
+                    "type": "object",
+                    "description": "Free-form metadata associated with the parameter"
```
bpkroth, line 15: I have some related changes for this in #451 that might be worth looking at.
In particular, it abstracts out the "meta" definition to a separate subschema and references it in each of the other tunable_params types for easier maintenance.
It also restricts what can go in there slightly.
My version didn't have tests yet though.

bpkroth: Side note: the current example use of `meta` for path prefixes is going to have to change.  It prevents our ability to tune two separate block device parameters with the same name due to the way we flatten them out by only parameter name.  For instance, `/sys/block/sda/queue/scheduler` and `/sys/block/sdb/queue/scheduler`
That's in part why I had kept full paths for the param names originally.
But we can address this in a future PR.

motus: You're right, and I've already almost ran into that issue with 1GB vs 2MB hugepages. OTOH, `"meta"` is just a way to pass some additional information about the tunables to the user code that produces the configuration scripts. That is, we can pass any information as long as the config-generating script can use it. For instance, we can write something like
```jsonc
"sda_scheduler": {
    "description": "/dev/sda disk scheduler",
    "type": "categorical",
    "meta": {"name": "/sys/block/sda/queue/scheduler"},
    "default": "none",
    "values": ["none", "mq-deadline", "bfq", "kyber"]
}
```
and then just use `"name"` metadata in `generate_kernel_config_script.py` or whatever.

motus: @bpkroth as for the schema, I don't really care what PR it comes from. If you have a better implementation, please feel free to close this PR and publish your version instead

bpkroth (approved):
Made some small changes to consolidate the meta definition a tad, but I think it's good now.

PR #451, opened 2023-07-18
WIP: Hacking: various minor fixups

bpkroth:
- look in local directory `.azure/config` files so we can set some configs via `az config set --local foo=bar`
  - [ ] A (future) further improvement would be for us to look at those automatically in the the various `AzureServices` classes.
- Expand `read_results_file` vars as well
  - [ ] needs tests
- Start improving OS selection in VM setup


mlos_bench/mlos_bench/environments/local/local_fileshare_env.py
```diff
@@ -69,6 +69,8 @@ def __init__(self,
             "LocalEnv requires a service that supports file upload/download operations"
         self._file_share_service: SupportsFileShareOps = self._service
 
+        # TODO: Replace these with expand_string util functions?
```
motus, line 72: not sure about `expand_string()` - we can build a `string.Template` once (e.g., in the constructor) and then just do `.safe_substitute()` multiple times. (not that it will boost the performance much, but still :) )

bpkroth: For sure.  That was just quick hacking.  I may throw this away as we discussed elsewhere.
I think the rest of the changes from this hacking branch are mostly in at this point.

S-code652 (approved):

PR #453, opened 2023-07-20
Make "include_tunables" work for composite environments

motus:
There is a bug in `CompositeEnv` constructor that prevents the parent's tunables to be pushed down into the child environments. In particular, the tunables from `"include_tunables"` in the composite environments are not available to all children without mixing them into the `CompositeEnv._tunable_params`. 

That issue makes impossible some perfectly legitimate configs, e.g.,
```jsonc
{
    "name": "Setup Ubuntu VM",
    "class": "mlos_bench.environments.composite_env.CompositeEnv",
    "include_tunables": ["environments/os/linux/runtime/linux-runtime-tunables.jsonc"],
    "config": {
        "children": [
            {
                "name": "Generate Linux kernel parameters for Ubuntu",
                "class": "mlos_bench.environments.LocalFileShareEnv",
                "config": {
                    "tunable_params": ["$linux-runtime"],
                    // ... etc.
```
Note that `"include_tunables"` is in the composite env whereas `"tunable_params"` is one level below it.

We need that kind of configurations for local-remote-local composites where all three child environments share the same set of tunables. This PR fixes the logic for propagating the tunables' data up and down the environment hierarchy

### DIFF
--- a/mlos_bench/mlos_bench/environments/composite_env.py
+++ b/mlos_bench/mlos_bench/environments/composite_env.py
@@ -51,7 +51,13 @@ class CompositeEnv(Environment):
         super().__init__(name=name, config=config, global_config=global_config,
                          tunables=tunables, service=service)
 
-        _LOG.debug("Build composite environment '%s' START: %s", self, self.tunable_params)
+        # By default, the Environment includes only the tunables explicitly specified
+        # in the "tunable_params" section of the config. `CompositeEnv`, however, must
+        # retain all tunables from its children environments plus the ones that come
+        # from the "include_tunables".
+        tunables = tunables.copy() if tunables else TunableGroups()
+
+        _LOG.debug("Build composite environment '%s' START: %s", self, tunables)
         self._children: List[Environment] = []
 
         # To support trees of composite environments (e.g. for multiple VM experiments),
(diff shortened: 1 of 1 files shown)


mlos_bench/mlos_bench/environments/composite_env.py
```diff
@@ -51,7 +51,13 @@ def __init__(self,
         super().__init__(name=name, config=config, global_config=global_config,
                          tunables=tunables, service=service)
 
-        _LOG.debug("Build composite environment '%s' START: %s", self, self.tunable_params)
+        # By default, the Environment includes only the tunables explicitly specified
+        # in the "tunable_params" section of the config. `CompositeEnv`, however, must
+        # retain all tunables from its children environments plus the ones that come
```
bpkroth, line 56:
```suggestion
        # retain all tunables from its parent environment plus the ones that come
```

?

bpkroth: Don't tunables flow down, not up?

bpkroth: I guess I'm also not clear on why the CompositeEnv needs to include its own tunables at all.  What is it tuning?  Shouldn't those belong in one of the other Envs? (OSEnv, VMEnv, RemoteEnv, etc.?)

bpkroth: Nm, we talked thru this.  I was misunderstanding.

bpkroth (approved):

PR #454, opened 2023-07-20
Make required_args apply to both const_args *and* tunables

motus:
Right now `required_args` applies only to `const_args` and the globals. It is useful to have `const_args` and `tunable_params` interchangeable without updating `required_params` in low-level configs.


bpkroth (approved):
Tests?

PR #461, opened 2023-07-27
mlos_bench: shell command line parsing improvements

bpkroth:
Adds support for splitting a single shell command line into subcommands and applying the first token script name full path resolution logic to it.
Also cleans up slightly the handling of `.` directories from #459.
For example, in the following:
```sh
. env.sh && partial/path/to/app/env/script.py || another/script.sh
```
`.` should be left alone as equivalent to the shell keyword `source`, and both scripts should resolved to an expanded full path, with the python script prefixed with a python executable path.

Tests for both are included.


mlos_bench/mlos_bench/tests/services/local/local_exec_test.py
```diff
@@ -85,6 +128,16 @@ def test_run_script_read_csv(local_exec_service: LocalExecService) -> None:
...
 
+        if sys.platform == 'win32':
+            # On Windows we need to remove the trailing ' ' from the CSV file
+            # that was written due to the way the Python recomposes shell
+            # expansions by reintroducing ' 's between lists of arguments.
+            with open(path_join(temp_dir, "output.csv"), "rt", encoding="utf-8") as fh_output:
+                lines = fh_output.readlines()
+            with open(path_join(temp_dir, "output.csv"), "wt", encoding="utf-8") as fh_output:
+                for line in lines:
+                    fh_output.write(line.rstrip() + os.linesep)
```
bpkroth, line 139: Not super pleased about this special case, but I don't see a great way around it.
Inside the subprocess module, Python is taking our list of arguments and inserting a space between them which `cmd` is too dumb to strip back out when doing the redirect.

mlos_bench/mlos_bench/optimizers/base_optimizer.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, tunables: TunableGroups, service: Optional[Service], config:
         self._start_with_defaults: bool = bool(
             strtobool(str(self._config.pop('start_with_defaults', True))))
         self._max_iter = int(self._config.pop('max_iterations', 100))
-        self._opt_target = self._config.pop('optimization_target', 'score')
+        self._opt_target = str(self._config.pop('optimization_target', 'score'))
```
bpkroth, line 49: Addresses an rnrelated mypy warning flagged from the return call type of `target` property.

mlos_bench/mlos_bench/tests/services/local/local_exec_test.py
```diff
@@ -86,6 +128,12 @@ def test_run_script_read_csv(local_exec_service: LocalExecService) -> None:
         assert stderr.strip() == ""
 
         data = pandas.read_csv(path_join(temp_dir, "output.csv"))
+        if sys.platform == 'win32':
+            # Workaround for Python's subprocess module on Windows adding a
+            # space inbetween the col1,col2 arg and the redirect symbol which
+            # cmd poorly interprets as being part of the original string arg.
+            # Without this, we get "col2 " as the second column name.
+            data.rename(str.rstrip, axis='columns', inplace=True)
```
bpkroth, line 136: Wonder if it's worth us doing this inside the `read_results_file` routine and then calling that here directly.
@motus thoughts?

motus: yes, that's a good idea! we can do it in a separate PR - let me finish the `read_telemetry_file` and then we'll do it in a more generic way

motus (approved):
Looks great! We can add more improvements later

mlos_bench/mlos_bench/services/local/local_exec.py
```diff
@@ -120,18 +181,12 @@ def _local_exec_script(self, script_line: str,
...
-
-        cmd = [script_path] + cmd[1:]
-        if script_path.strip().lower().endswith(".py"):
-            cmd = [sys.executable] + cmd
+        # Split the command line into set of subcmd tokens.
+        # For each subcmd, perform path resolution fixups for any scripts being executed.
+        subcmds = split_cmdline(script_line)
+        subcmds = [self._resolve_cmdline_script_path(subcmd) for subcmd in subcmds]
+        # Finally recombine all of the fixed up subcmd tokens into the original.
+        cmd = [token for subcmd in subcmds for token in subcmd]
```
motus, line 189: Could that be the source of the problem? It looks like we try to resolve the paths of _all_ tokens of a command line. Before this change we were doing it only for the script

bpkroth: We're still only doing it for the script.

Per this PR, the `split_cmdline` splits complex shell commands like `foo path/to/file && bar || baz --args` into an array of individual commands:
```python
[
  [`foo`, `path/to/file`],
  [`&&`],
  [`bar`],
  [`||`],
  [`baz`, `--args`],
]
```
https://github.com/bpkroth/MLOS/blob/c7fa42b30b05addb3f8236c88e2a33a21b40baf7/mlos_bench/mlos_bench/services/local/local_exec.py#L28

Then `_resolve_cmdline_script_path` only looks at the first item in each subcmd list to resolve it:
https://github.com/bpkroth/MLOS/blob/c7fa42b30b05addb3f8236c88e2a33a21b40baf7/mlos_bench/mlos_bench/services/local/local_exec.py#L136

There's no sane way to do that for arguments after the initial subcommand.
They could be files, string args, etc.

PR #466, opened 2023-08-01
Save and restore the trial telemetry data

motus:
Summary of changes:
* Fix the `Trial.update_telemetry()` method in Storage
* Implement `Experiment.load_telemetry()` method
* Update the `Environment.status()` to be compatible with the Storage API and the upcoming changes in `LocalEnv`
* Always pass the timestamp to `Trial.update()` (later we'll need to make sure the timestamp always comes from the remote system)
* Add unit tests for `update_telemetry/load_telemetry` functionality


bpkroth:
> Update the Environment.status() to be compatible with the Storage API and the upcoming changes in LocalEnv

Can you give some teaser explanations on this a bit?

mlos_bench/mlos_bench/run.py
```diff
@@ -137,7 +139,8 @@ def _run(env: Environment, opt: Optimizer,
 
     if not env.setup(trial.tunables, trial.config(global_config)):
         _LOG.warning("Setup failed: %s :: %s", env, trial.tunables)
-        trial.update(Status.FAILED)
+        # FIXME: Use the actual timestamp from the environment.
+        trial.update(Status.FAILED, datetime.now())
```
bpkroth, line 143:
```suggestion
        trial.update(Status.FAILED, datetime.utcnow())
```

To make things more easily comparable, let's standardize on using UTC time everywhere.

mlos_bench/mlos_bench/storage/sql/experiment.py
```diff
@@ -93,6 +93,20 @@ def load_config(self, config_id: int) -> Dict[str, Any]:
...
+                self._schema.trial_telemetry.select().where(
+                    self._schema.trial_telemetry.c.exp_id == self._experiment_id,
+                    self._schema.trial_telemetry.c.trial_id == trial_id
+                ).order_by(
+                    self._schema.trial_telemetry.c.ts,
+                    self._schema.trial_telemetry.c.metric_id,
+                )
+            )
+            return [(row.ts, row.metric_id, row.metric_value)
+                    for row in cur_telemetry.fetchall()]
```
bpkroth, line 108: Seems like a good candidate for a DataFrame instead of a list of tuples.
We lose schema info I suppose though ...

motus: yes - but if remember correctly, we've decided not to use pandas anywhere in the Storage or Environment API

mlos_bench/mlos_bench/storage/sql/trial.py
```diff
@@ -92,14 +77,19 @@ def _update(self, table: Table, timestamp: Optional[datetime],
...
+                try:
+                    conn.execute(self._schema.trial_telemetry.insert().values(
+                        exp_id=self._experiment_id,
+                        trial_id=self._trial_id,
+                        ts=timestamp,
+                        metric_id=key,
+                        metric_value=None if val is None else str(val),
+                    ))
+                except IntegrityError as ex:
+                    _LOG.warning("Record already exists: %s :: %s", (timestamp, key, val), ex)
```
bpkroth, line 95: Maybe also log the existing value if possible?

bpkroth: Also makes me wonder if we need to consider the granularity of the timestamp to include `ms` or some such.

mlos_bench/mlos_bench/tests/storage/exp_load_test.py
```diff
@@ -66,7 +66,7 @@ def test_exp_trial_pending_fail(exp_storage_memory_sql: Storage.Experiment,
     Start a trial, fail it, and and check that it is NOT pending.
     """
     trial = exp_storage_memory_sql.new_trial(tunable_groups)
-    trial.update(Status.FAILED)
+    trial.update(Status.FAILED, datetime.now())
```
bpkroth, line 69:
```suggestion
    trial.update(Status.FAILED, datetime.utcnow())
```

mlos_bench/mlos_bench/tests/storage/exp_load_test.py
```diff
@@ -77,7 +77,7 @@ def test_exp_trial_success(exp_storage_memory_sql: Storage.Experiment,
     Start a trial, finish it successfully, and and check that it is NOT pending.
     """
     trial = exp_storage_memory_sql.new_trial(tunable_groups)
-    trial.update(Status.SUCCEEDED, 99.9)
+    trial.update(Status.SUCCEEDED, datetime.now(), 99.9)
```
bpkroth, line 80:
```suggestion
    trial.update(Status.SUCCEEDED, datetime.utcnow(), 99.9)
```

mlos_bench/mlos_bench/run.py
motus: Actually, this is a temporary hack - I think we should actually use a timestamp from the remote VM. That is, record the time when the benchmark had failed rather than when we've noticed it

mlos_bench/mlos_bench/tests/storage/exp_load_test.py
bpkroth: Same for the rest ...

mlos_bench/mlos_bench/tests/storage/trial_telemetry_test.py
```diff
@@ -0,0 +1,73 @@
...
+    Make sure update_telemetry() and load_telemetry() methods work.
+    """
+    trial = exp_storage_memory_sql.new_trial(tunable_groups)
+    assert exp_storage_memory_sql.load_telemetry(trial.trial_id) == []
+
+    trial.update_telemetry(Status.RUNNING, telemetry_data)
+    assert exp_storage_memory_sql.load_telemetry(trial.trial_id) == _telemetry_str(telemetry_data)
+
+
+def test_update_telemetry_twice(exp_storage_memory_sql: Storage.Experiment,
```
bpkroth, line 63: Add a test for `IntegrityConstraint` handling when metrics are attempted to be inserted twice?

motus:
> Can you give some teaser explanations on this a bit?

@bpkroth Sure - it is in this PR already at [base_environment.py:322](https://github.com/microsoft/MLOS/blob/1d59363c11d9d50921d6e72be0afbff53c7a9874/mlos_bench/mlos_bench/environments/base_environment.py#L322)
That is, telemetry data is a bunch or triplets (timestamp, metric_id, metric_value) where a combination of (timestamp, metric_id) is unique but the same metric_id can appear multiple times
This is different from the final results, where we get pairs of (metric_id, metric_value) and metric_id is a unique key

mlos_bench/mlos_bench/tests/storage/trial_telemetry_test.py
motus: oh, we actually swallow that exception and only print a warning. We want the `.update_telemetry()` method to be idempotent. It is totally fine to retrieve the same telemetry data (that is, unique pairs of [timestamp, metric_id]) twice - e.g., when on the remote VM a script keeps appending this data to a file and we repeatedly download the same file and ingest it

mlos_bench/mlos_bench/storage/sql/experiment.py
bpkroth: Ah, could have been and I just forgot.
Am just thinking that with the amount of data for telemetry in particular and possibly wanting to slice and dice it different ways that might be more efficient, but we can tackle that later if it shows up.

bpkroth:
> > Can you give some teaser explanations on this a bit?
> 
> @bpkroth Sure - it is in this PR already at [base_environment.py:322](https://github.com/microsoft/MLOS/blob/1d59363c11d9d50921d6e72be0afbff53c7a9874/mlos_bench/mlos_bench/environments/base_environment.py#L322) That is, telemetry data is a bunch or triplets (timestamp, metric_id, metric_value) where a combination of (timestamp, metric_id) is unique but the same metric_id can appear multiple times This is different from the final results, where we get pairs of (metric_id, metric_value) and metric_id is a unique key

We will probably have to extend the `setup, `run`, `teardown` lifecycle to also include `poll_status` or some such so that we can get status overtime.

Another nice feature that this can help support in the future: early trial aborting.

e.g., if a particular trial looks like it has no possibility of doing better than existing ones for that particular context (also unimplemented) at this point in time into the trial, then we can abort it early, report a fake "bad" score to the optimizer, or one extrapolated and scaled from the partial metrics, and have it save that time to use for another config point.

Future work ...

mlos_bench/mlos_bench/storage/sql/trial.py
motus: as far as I know, `datetime` does include microseconds but it is up to the telemetry-generating scripts to decide on the granularity. In real applications, I believe we'll have telemetry data aggregated for 1 minute intervals or so. still, I want to use `datetime` in our API because this is Python standard and both SQLAlchemy and pandas handle it well

motus:
> We will probably have to extend the `setup`, `run`, `teardown` lifecycle to also include`poll_status` or some such so that we can get status overtime.
> 
> Another nice feature that this can help support in the future: early trial aborting.

Absolutely! I am actually fixing the Environment API now so we can easily add such functionality

mlos_bench/mlos_bench/tests/storage/trial_telemetry_test.py
bpkroth: Yeah, I saw that.  We should add a test for that was my suggestion.

bpkroth: Looks like right now we do it in a loop, which is ineffecient but the only correct way to handle the scenario you describe.
So, to prevent some "smart" person (i.e., me) from trying to turn that into a more effecient bulk insert in the future, where leading row repeats would fail, and cause subsequent rows to also fail even if they're new, we should check that for this handling logic.

mlos_bench/mlos_bench/run.py
motus: on a second thought, let's use UTC time everywhere. later we can get timestamps from a remote VM, but it still going to be UTC

mlos_bench/mlos_bench/tests/storage/trial_telemetry_test.py
motus: oh, I finally understand what you mean. I did some research on SQLAlchemy - there is a function `Insert.on_conflict_do_nothing()`, but it is only available for a few dialects (and MySQL is _not_ one of them) ☹️ 
I'll play with it a bit more later - apparently, there is an upsert functionality in a higher-level Session API; I'll see if we can use it in our code

motus: I cannot figure it out now. I've added a comment in `.update_telemetry()` explaining the decision; I'll replace it with a better solution later (most likely, explicitly check if current dialect supports `on_conflict_do_nothing` and use bulk insert if it does).

mlos_bench/mlos_bench/storage/sql/trial.py
```diff
@@ -92,14 +77,22 @@ def _update(self, table: Table, timestamp: Optional[datetime],
...
         return metrics
 
-    def update_telemetry(self, status: Status,
-                         metrics: Optional[Dict[str, float]] = None) -> None:
+    def update_telemetry(self, status: Status, metrics: List[Tuple[datetime, str, Any]]) -> None:
         super().update_telemetry(status, metrics)
-        self._update(self._schema.trial_telemetry, None, status, metrics)
+        # NOTE: Not every SQLAlchemy dialect supports `Insert.on_conflict_do_nothing()`
+        # and we need to keep `.update_telemetry()` idempotent; hence a loop instead of
+        # a bulk upsert.
```
bpkroth, line 86: :heart:

PR #469, opened 2023-08-02
Introduce the Environment context for setup/run/teardown (not enforced yet)

motus:
This simplifies `LocalEnv` and `LocalFileShareEnv` logic a lot and will make our life easier for asynchronous environments. We also need it for getting the telemetry data from the environment


mlos_bench/mlos_bench/run.py
```diff
@@ -136,24 +136,25 @@ def _run(env: Environment, opt: Optimizer,
...
+    with env as env_context:
+
+        if not env_context.setup(trial.tunables, trial.config(global_config)):
+            _LOG.warning("Setup failed: %s :: %s", env_context, trial.tunables)
+            # FIXME: Use the actual timestamp from the environment.
+            trial.update(Status.FAILED, datetime.utcnow())
+            opt.register(trial.tunables, Status.FAILED)
+            return
+
+        # In async mode, poll the environment for status and telemetry
```
bpkroth, line 148:
```suggestion
        # In async mode (TODO), poll the environment for status and telemetry
```

mlos_bench/mlos_bench/run.py
```diff
@@ -136,24 +136,25 @@ def _run(env: Environment, opt: Optimizer,
...
+        # and update the storage with the intermediate results.
+        (status, telemetry) = env_context.status()
+        trial.update_telemetry(status, telemetry)
```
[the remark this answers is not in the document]
motus: exactly!

mlos_bench/mlos_bench/environments/base_environment.py
```diff
@@ -174,6 +176,30 @@ def _config_loader_service(self) -> "SupportsConfigLoading":
...
+    def __enter__(self) -> 'Environment':
+        """
+        Enter the environment's benchmarking context.
+        """
+        _LOG.debug("Environment START :: %s", self)
+        assert not self._in_context
+        self._in_context = True
+        return self
+
+    def __exit__(self, exc_type: Optional[Type[BaseException]],
```
bpkroth, line 188:
```suggestion
    def __exit__(self, exception_type: Optional[Type[BaseException]],
```

Slightly more descriptive variables perhaps?

`exc` reads almost like `exec` to me.

mlos_bench/mlos_bench/environments/base_environment.py
```diff
@@ -268,6 +294,9 @@ def setup(self, tunables: TunableGroups, global_config: Optional[dict] = None) -
         _LOG.info("Setup %s :: %s", self, tunables)
         assert isinstance(tunables, TunableGroups)
 
+        # TODO: Make sure we create a context before invoking setup/run/status/teardown
+        # assert self._in_context
```
bpkroth, line 298: When does this get enabled?

motus: today, in the upcoming PR

mlos_bench/mlos_bench/environments/base_environment.py
motus: agree. will fix in a second

mlos_bench/mlos_bench/environments/local/local_env.py
```diff
@@ -63,13 +69,31 @@ def __init__(self,
...
+        self._temp_dir: Optional[str] = None
+        self._temp_dir_context: Union[TemporaryDirectory, nullcontext, None] = None
 
         self._dump_params_file: Optional[str] = self.config.get("dump_params_file")
         self._dump_meta_file: Optional[str] = self.config.get("dump_meta_file")
         self._read_results_file: Optional[str] = self.config.get("read_results_file")
 
+    def __enter__(self) -> Environment:
+        assert self._temp_dir is None and self._temp_dir_context is None
+        self._temp_dir_context = self._local_exec_service.temp_dir_context(self.config.get("temp_dir"))
```
bpkroth, line 81: One hiccup with this could be that if `temp_dir` config is invalid for some reason (e.g., bad path passed in the config), then the error won't be propagated until runtime, rather than during initialization.  I think it's preferable to do the former whenever possible.

Perhaps we can add a quick sanity check on `config["temp_dir"]` during init that it is in fact a valid `Path` (or else `None`).

mlos_bench/mlos_bench/environments/local/local_fileshare_env.py
```diff
@@ -140,15 +137,11 @@ def run(self) -> Tuple[Status, Optional[dict]]:
...
-            self._temp_dir = prev_temp_dir
-            return result
+        assert self._temp_dir is not None
+        params = self._get_env_params()
+        params["PWD"] = self._temp_dir
+        for (path_from, path_to) in self._expand(self._download, params):
+            self._file_share_service.download(
+                path_from, self._config_loader_service.resolve_path(
+                    path_to, extra_paths=[self._temp_dir]))
+        return super().run()
```
bpkroth, line 147: Order of this seems a little funny to me.  The assert (not enabled yet) would happen after the work is being done, and it would return some results from the parent rather than the target callee?

bpkroth (commented):
High level comment on the title of the PR - this doesn't actually require that behavior yet - just moves towards adding support for it.

mlos_bench/mlos_bench/environments/base_environment.py
motus: absolutely! once I uncomment this, all tests break and that makes it a much bigger PR, so I've decided to split it in two

mlos_bench/mlos_bench/environments/local/local_fileshare_env.py
motus: yes, the parent of `LocalFileShareEnv` is `LocalEnv`, and it has logic that has to be executed _after_ the download. The assertion part is a bit wacky, but we have the assertions in `.setup()` already.. Anyway, I think it won't hurt to add one more assertion before the download

mlos_bench/mlos_bench/environments/local/local_env.py
motus: I've added some checks in the constructor

mlos_bench/mlos_bench/environments/local/local_fileshare_env.py
motus: oh wait, we do have an assert at teh beginning of `LocalFileShareEnv.run()`!

mlos_bench/mlos_bench/environments/local/local_env.py
```diff
@@ -63,13 +70,36 @@ def __init__(self,
...
             "LocalEnv requires a service that supports local execution"
         self._local_exec_service: SupportsLocalExec = self._service
+        self._temp_dir: Optional[str] = None
+        self._temp_dir_context: Union[TemporaryDirectory, nullcontext, None] = None
 
-        self._temp_dir = self.config.get("temp_dir")
+        temp_dir = self.config.get("temp_dir")
+        if temp_dir is not None and not os.path.isdir(temp_dir):
+            _LOG.error("temp_dir path is invalid: %s", temp_dir)
+            raise ValueError(f"Temp directory path is invalid: {temp_dir}")
```
bpkroth, line 79: This is probably a little too strict.  Creating the base temp_dir on demand should be ok, as well (as long as we have permissions for it).

bpkroth (approved):
Minor tweaks on the constructor check, but otherwise looks good.

mlos_bench/mlos_bench/environments/local/local_env.py
motus: I see. Let me roll back the change and I will do it properly with unit tests and stuff later

PR #470, opened 2023-08-02
Improvements to SMAC optimizer integration for mlos_bench

bpkroth:
- [ ] Add support for passing `$experimentId` as `run_name`


mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
```diff
@@ -38,6 +40,23 @@ def __init__(self, tunables: TunableGroups, service: Optional[Service], config:
...
+            else:
+                _LOG.warning("No output_directory specified. SMAC will use temporary directory.")
+
+            # Set max_trials == max_iterations.
+            if 'max_trials' not in self._config:
+                self._config['max_trials'] = self._max_iter
+            assert self._config['max_trials'] == self._max_iter, \
+                f"max_trials {self._config['max_trials']} != max_iterations {self._max_iter}"
+
+            # TODO: Set run_name from experimentId (global_params not currently accessible from here).
```
bpkroth, line 58: Would like to add support for this to make examining logs easier, but it would require some changes to the `__init__` signature that I haven't dug through yet.

mlos_bench/mlos_bench/tests/optimizers/toy_optimization_loop_test.py
```diff
@@ -93,10 +93,10 @@ def test_smac_optimization_loop(mock_env_no_noise: MockEnv,
     Toy optimization loop with mock environment and SMAC optimizer.
     """
     (score, tunables) = _optimize(mock_env_no_noise, smac_opt)
-    assert score == pytest.approx(75.0, 0.01)
+    assert score == pytest.approx(67.6, 0.01)
```
bpkroth, line 96: These changed because previously the `max_trial` was still set to 100 inside SMAC due to lack of config propagation.  This changes the way the initial random sampling proceeds.

mlos_bench/mlos_bench/config/schemas/optimizers/mlos_core-optimizer-subschema.json
```diff
@@ -60,7 +60,20 @@
...
+                        "description": "Directory to store SMAC output files.",
+                        "type": ["string", "null"],
+                        "pattern": "^[a-zA-Z0-9_./-]+$",
+                        "example": "smac_output"
+                    },
+                    "run_name": {
+                        "description": "Name to give for the smac output (e.g., $experimentId). If null is used, SMAC will calculate a hash based name.",
+                        "type": ["string", "null"],
+                        "pattern": "^[a-zA-Z0-9_.-]+$",
+                        "example": "$experimentId"
```
bpkroth, line 74: Note: this won't actually work just yet.  We don't do `$var` expansion on all places.  Will tackle that in a different PR.

mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
bpkroth: #474 is needed before this

motus (approved):
Looks good! Let's merge it in after #477 

Also, in the next PRs, we should think about creating an optimizer context, similar to the one we have in `mlos_bench.Environment`. This way, we can easily create and clean up a temp directory for the optimizer

mlos_bench/mlos_bench/tests/optimizers/toy_optimization_loop_test.py
bpkroth: But now running into an issue with them not being deterministic across python versions ...

.github/workflows/linux.yml
```diff
@@ -134,6 +134,10 @@ jobs:
     #- name: Run lint checks
     #  run: make CONDA_ENV_NAME=$CONDA_ENV_NAME check
 
+    # Debugging - #470
+    - name: Run mlos_bench optimizer tests
+      run: conda run -n $CONDA_ENV_NAME pytest mlos_bench/mlos_bench/tests/optimizers/
```
bpkroth, line 139: - [ ] TODO: Remove this

mlos_bench/mlos_bench/tests/optimizers/conftest.py
```diff
@@ -102,6 +102,11 @@ def flaml_opt_max(tunable_groups: TunableGroups) -> MlosCoreOptimizer:
         },
     )
 
+# FIXME: SMAC's RF model can be non-deterministic at low iterations, which are
+# normally calculated as a percentage of the max_iterations and number of
+# tunable dimensions, so for now we set the initial random samples equal to the
+# number of iterations and control them with a seed.
```
bpkroth, line 108: @motus FYI ^^^

bpkroth: *argh!* SMAC is override that:
https://github.com/automl/SMAC3/blob/541ee7e0383b491b86d1a23dcff669f2efad616d/smac/initial_design/abstract_initial_design.py#L92

bpkroth: Added a way to pass the override for that too: `max_ratio`

This feels kludgey, but I don't want to be tracking down random seed issues anymore and just want the tests to pass reliably for now.

I filed a bug (#482) so we can come back and address this in the future.  Am worried about reproducible science mostly, but that will come with lots of data and we need to move on to get to that.

motus (approved):

Issue #357, opened 2023-05-15
Rename trialId, experimentId, vmName, etc. for snake_case consistency


bpkroth:
Low prio

motus:
Fixed in #471 
The remaining IDs are mostly for the ARM template parameters and follow the ARM template and Azure API naming conventions

PR #471, opened 2023-08-03
experiment_id fixups and restrictions

bpkroth:
Adds schema restrictions on experiment_id since we use (or plan to use) it in path names.

Per #357, also renames for snake case consistency.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -50,10 +50,14 @@
         },
         {
             "fileMatch": [
-                "global_config.jsonc",
-                "global_config.json",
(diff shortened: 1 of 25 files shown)


motus (approved):
Looks good! Now we need to do teh same to our configs in other repositories

bpkroth:
> Looks good! Now we need to do teh same to our configs in other repositories

I can do that tomorrow.  It's basically a couple of `rg -l ... | xargs sed -is 's/.../.../g'` and related `git status --numstat | xargs git add -p` types of pipelines and pretty quick to do.

PR #472, opened 2023-08-03
Autopep8 devcontainer fixups

bpkroth:
The recommended autopep8 extension doesn't seem to be working for us:
https://github.com/microsoft/vscode-autopep8/issues/16

So, for now I'm reverting to the old way of doing things.

Also, the previous extension relied on a newer version of autopep8 than conda was installing anyways, so I switched to installing it from pip.


motus (approved):
LGTM

PR #473, opened 2023-08-03
Enable in_context checks for all environments, update unit tests

motus:
Plus a few bugfixes:
* in the `Launcher`, we were invoking teardown not in context
* in the `CompositeEnv`, we were not creating contexts for child environments

Also, updated the unit tests to check for the new conditions

To be merged after #469


mlos_bench/mlos_bench/environments/base_environment.py
```diff
@@ -329,8 +329,7 @@ def teardown(self) -> None:
         single call.
         """
         _LOG.info("Teardown %s", self)
-        # TODO: Make sure we create a context before invoking setup/run/status/teardown
```
bpkroth, line 332: Maybe leave this comment and just remove the TODO?  It helps explain the assert a little bit.

mlos_bench/mlos_bench/environments/composite_env.py
```diff
@@ -82,6 +86,18 @@ def __init__(self,
...
 
+    def __enter__(self) -> Environment:
+        self._contexts = [env.__enter__() for env in self._children]
+        return super().__enter__()
+
+    def __exit__(self, ex_type: Optional[Type[BaseException]],
+                 ex_val: Optional[BaseException],
+                 ex_tb: Optional[TracebackType]) -> Literal[False]:
+        for env in reversed(self._children):
+            env.__exit__(ex_type, ex_val, ex_tb)
```
bpkroth, line 97: What happens if one of these raises an exception, but not all of them?
Is there retry logic somewhere that (maybe in the future) could get implemented?
Would this all or nothing cleanup of the list of existing contexts cause problems?

mlos_bench/mlos_bench/environments/composite_env.py
```diff
@@ -59,6 +62,7 @@ def __init__(self,
 
         _LOG.debug("Build composite environment '%s' START: %s", self, tunables)
         self._children: List[Environment] = []
+        self._contexts: List[Environment] = []
```
bpkroth, line 65:
```suggestion
        self._child_contexts: List[Environment] = []
```

bpkroth: Slight rename for clarity.
Without that, it's not entirely clear on a read through that a member `context` in that list even has a `run()` method for instance.

bpkroth (approved):
Looks pretty good to me.
Minor variable name rename request.
And a question about partial failure handling, but that part is probably not something that gets changed here.  Moreso curiosity for future.  May come into play with parallel runners.

mlos_bench/mlos_bench/environments/composite_env.py
motus: ooooh, that's right! Really nice catch! Will fix it now

mlos_bench/mlos_bench/environments/composite_env.py
motus: yep

mlos_bench/mlos_bench/environments/base_environment.py
motus: makes perfect sense! will do it now

PR #474, opened 2023-08-03
mlos_bench: pass global_config to remaining base classes

bpkroth:
For supporting `experiment_id` as `run_name` in #470.

Future work:
- Consolidate global_config variable expansion recursively and in other configs.


motus (approved):
Looks good!

PR #476, opened 2023-08-03
Use of service principal for generating access tokens

eujing:
This PR aims to reduce the need for interactive logins during an ongoing experiment by authenticating as a service principal (instead of personal account). This should enable us to run longer experiments without having to re-login.

The requirements of this are:
- An existing SP that has the correct access to the target resource group
- A certificate to authenticate as the SP, which is stored in a key vault that the current user has access to

The flow of authentication is inspired by @bpkroth's protocol:
1) On initialization, as the current user retrieve the SP's certificate from the key vault
2) Log in as the SP using the certificate
3) During runtime when tokens are requested, we request them as the SP

Will follow up with a PR in the MySQL repo with an example script that automates the set up of the SP, key vault, certificates and resource group in general.

### DIFF
--- a/mlos_bench/mlos_bench/config/schemas/services/remote/azure/azure-auth-service-subschema.json
+++ b/mlos_bench/mlos_bench/config/schemas/services/remote/azure/azure-auth-service-subschema.json
@@ -23,6 +23,18 @@
                         "tokenRequestInterval": {
                             "description": "Time before token expiration to request a new token",
                             "type": "number"
+                        },
+                        "keyVaultName": {
+                            "description": "Name of key vault to retrieve the certificate from, for authorizing the Service Principal",
+                            "type": "string"
+                        },
+                        "certName": {
+                            "description": "Name of the Service Principal's certificate in the key vault",
+                            "type": "string"
+                        },
+                        "spClientId": {
+                            "description": "Client ID of the Service Principal to authenticate as",
+                            "type": "string"
                         }
                     }
                 }
--- a/mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
+++ b/mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
@@ -7,13 +7,16 @@ A collection Service functions for managing VMs on Azure.
 """
 
 import datetime
-import json
 import logging
-import subprocess
+from base64 import b64decode
 from typing import Any, Dict, Optional
+
+import azure.identity as azure_id
+import azure.keyvault.secrets as keyvault_secrets
 
 from mlos_bench.services.base_service import Service
 from mlos_bench.services.types.authenticator_type import SupportsAuth
+from mlos_bench.util import check_required_params
 
 _LOG = logging.getLogger(__name__)
 
@@ -24,6 +27,7 @@ class AzureAuthService(Service, SupportsAuth):
     """
 
     _REQ_INTERVAL = 300   # = 5 min
+    _TENTANT_ID = "72f988bf-86f1-41af-91ab-2d7cd011db47"
 
     def __init__(self,
                  config: Optional[Dict[str, Any]] = None,
@@ -47,11 +51,41 @@ class AzureAuthService(Service, SupportsAuth):
         # Register methods that we want to expose to the Environment objects.
         self.register([self.get_access_token])
 
+        check_required_params(
+            self.config, {
+                "keyVaultName",
+                "certName",
+                "spClientId",
+            }
+        )
+
         # This parameter can come from command line as strings, so conversion is needed.
         self._req_interval = float(self.config.get("tokenRequestInterval", self._REQ_INTERVAL))
+        keyvault_name = self.config.get("keyVaultName")
+        cert_name = self.config.get("certName")
+        sp_client_id = self.config.get("spClientId")
+        tenant_id = self.config.get("tenant", self._TENTANT_ID)
 
         self._access_token = "RENEW *NOW*"
         self._token_expiration_ts = datetime.datetime.now()  # Typically, some future timestamp.
+
+        # Login as ourselves
+        local_user_cred = azure_id.AzureCliCredential()
+
+        # Get a client for fetching cert info.
+        keyvault_secrets_client = keyvault_secrets.SecretClient(
+            vault_url=f"https://{keyvault_name}.vault.azure.net",
+            credential=local_user_cred,
+        )
+
+        # The certificate private key data is stored as hidden "Secret" (not Key strangely)
+        #  in PKCS12 format, but we need to decode it.
+        secret = keyvault_secrets_client.get_secret(cert_name)
+        assert secret.value is not None
+        cert_bytes = b64decode(secret.value)
+
+        # Reauthenticate as the service principal.
+        self._sp_cred = azure_id.CertificateCredential(tenant_id=tenant_id, client_id=sp_client_id, certificate_data=cert_bytes)
 
     def get_access_token(self) -> str:
         """
@@ -61,10 +95,8 @@ class AzureAuthService(Service, SupportsAuth):
         _LOG.debug("Time to renew the token: %.2f sec.", ts_diff)
         if ts_diff < self._req_interval:
             _LOG.debug("Request new accessToken")
-            # TODO: Use azure-identity SDK and a key valut instead of `az` CLI.
-            res = json.loads(subprocess.check_output(
-                'az account get-access-token', shell=True, text=True))
-            self._token_expiration_ts = datetime.datetime.fromisoformat(res["expiresOn"])
-            self._access_token = res["accessToken"]
+            res = self._sp_cred.get_token("https://management.azure.com/.default")
+            self._token_expiration_ts = datetime.datetime.fromtimestamp(res.expires_on)
+            self._access_token = res.token
             _LOG.info("Got new accessToken. Expiration time: %s", self._token_expiration_ts)
         return self._access_token
--- a/mlos_bench/mlos_bench/config/services/remote/azure/service-auth.jsonc
+++ b/mlos_bench/mlos_bench/config/services/remote/azure/service-auth.jsonc
@@ -1,6 +1,9 @@
 {
     "class": "mlos_bench.services.remote.azure.AzureAuthService",
     "config": {
-        "tokenRequestInterval": 300  // seconds
+        "tokenRequestInterval": 300,  // seconds
+        "keyVaultName": "PLACEHOLDER; e.g. mlos-autotune-kv",
+        "certName": "PLACEHOLDER; e.g. mlos-autotune-sp-cert",
+        "spClientId": "PLACEHOLDER;"
     }
 }
--- a/mlos_bench/setup.py
+++ b/mlos_bench/setup.py
@@ -27,7 +27,7 @@ except LookupError as e:
 
 extra_requires: Dict[str, List[str]] = {    # pylint: disable=consider-using-namedtuple-or-dataclass
     # Additional tools for extra functionality.
-    'azure': ['azure-storage-file-share'],
+    'azure': ['azure-storage-file-share', 'azure-identity', 'azure-keyvault'],
     'storage-sql-duckdb': ['sqlalchemy', 'duckdb_engine'],
     'storage-sql-mysql': ['sqlalchemy', 'mysql-connector-python'],
     'storage-sql-postgres': ['sqlalchemy', 'psycopg2'],


eujing:
@microsoft-github-policy-service agree company="Microsoft"

bpkroth:
Thanks for taking this on @eujing.  I will try and take a look at this and give more feedback on Monday.

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
```diff
@@ -47,12 +51,49 @@ def __init__(self,
...
+                self.config, {
+                    "keyVaultName",
+                    "certName",
+                    "spClientId",
+                }
+            )
+
         # This parameter can come from command line as strings, so conversion is needed.
         self._req_interval = float(self.config.get("tokenRequestInterval", self._REQ_INTERVAL))
+        keyvault_name = self.config.get("keyVaultName")
```
motus, line 65: if the key is required, then maybe write
```suggestion
        keyvault_name = self.config["keyVaultName"]
```

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
```diff
@@ -24,6 +27,7 @@ class AzureAuthService(Service, SupportsAuth):
     """
 
     _REQ_INTERVAL = 300   # = 5 min
+    _TENTANT_ID = "72f988bf-86f1-41af-91ab-2d7cd011db47"
```
motus, line 30: maybe we should not hardcode it. let's move it to the external config outside of git

mlos_bench/mlos_bench/config/schemas/services/remote/azure/azure-auth-service-subschema.json
```diff
@@ -23,6 +23,18 @@
...
+                        },
+                        "keyVaultName": {
+                            "description": "Name of key vault to retrieve the certificate from, for authorizing the Service Principal",
+                            "type": "string"
+                        },
+                        "certName": {
+                            "description": "Name of the Service Principal's certificate in the key vault",
+                            "type": "string"
+                        },
+                        "spClientId": {
```
motus, line 35: Can we have a config file with these keys at `config/services/remote/azure/service-auth.jsonc`?
Maybe use the placeholder values like in `config/services/remote/azure/service-fileshare.jsonc`

motus, line 35: Also, can you add to `mlos_bench/README.md` some text on how to get this data? (or add a script, if applicable)

motus, line 35: Also, @bpkroth would probably want some test JSON configs with and without these keys

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
```diff
@@ -47,12 +51,49 @@ def __init__(self,
...
+        cert_name = self.config.get("certName")
+        sp_client_id = self.config.get("spClientId")
+        tenant_id = self.config.get("tenant", self._TENTANT_ID)
 
         self._access_token = "RENEW *NOW*"
         self._token_expiration_ts = datetime.datetime.now()  # Typically, some future timestamp.
 
+        # Login as ourselves
+        local_user_cred = azure_id.AzureCliCredential()
+        self._cred = local_user_cred
```
motus, line 75: Why not just
```suggestion
        self._cred = azure_id.AzureCliCredential()
```
and use `self._cred` everywhere? pylint might be complaining about the number of local variables already

motus (changes requested):
Left a few comments. We'll need a new version of `config/services/remote/azure/service-auth.jsonc` (with placeholders, not real values!) in our private repo as well (plus the instructions on how to generate a working config)

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
```diff
@@ -47,12 +51,49 @@ def __init__(self,
...
+                    "certName",
+                    "spClientId",
+                }
+            )
+
         # This parameter can come from command line as strings, so conversion is needed.
         self._req_interval = float(self.config.get("tokenRequestInterval", self._REQ_INTERVAL))
+        keyvault_name = self.config.get("keyVaultName")
+        cert_name = self.config.get("certName")
+        sp_client_id = self.config.get("spClientId")
```
motus, line 67: if the key is required, then maybe write
```suggestion
        keyvault_name = self.config["keyVaultName"]
        cert_name = self.config["certName"]
        sp_client_id = self.config["spClientId"]
```

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
motus, line 30: also, a typo:
```suggestion
    _TENANT_ID = "..."
```

mlos_bench/mlos_bench/config/schemas/services/remote/azure/azure-auth-service-subschema.json
bpkroth: I had asked @eujing about creating a second PR for doing the 
1. control plane resources setup
2. RG setup

I think that would be a good place to do the helpers for this, but I do agree that somewhere (e.g., `serivces/remote/azure/README.md` or `config/services/remote/azure/README.md`) a doc that describes the general workflow in some way would be good too.  (it also helps with PR reviews ;)

bpkroth: Yep, they should be at least added to the `tests/config/schemas/services/test-cases/good/full/...` area.

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
bpkroth: Absolutely - this should not be in the repo. It should be in the globals_config or else can we look it up from subscription (since that is already required I think).

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
bpkroth: It should also be marked in the schema as required then, however, I think we also discussed making this *not* required, so that if someone doesn't have access to create an SP or assign IAM rights to an RG, then they can still use the framework as themselves as a fallback (albeit with some potential token refresh issues).
In that case, we can maybe just issue a warning, and move on.

eujing: I agree I wanted to make the use of the SP optional in case a user did not have appropriate access to use one. But if they did use the SP, then this group of parameters should be required together. I will move these around to reflect that!

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
eujing: Good point! I will remove this, and I realize the `generate-azure-credentials-config.*` scripts already include the tenant into the global config.

mlos_bench/mlos_bench/config/schemas/services/remote/azure/azure-auth-service-subschema.json
eujing: Initially I added these keys into that file with placeholder values.

However because we end up trying to use the keyvault when those keys are present, the unit tests around loading environments end up failing because those interactions are not mocked out.
I will try to see if I can modify those unit tests!

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
eujing: Removed the hardcoded value! Theres a placeholder for it now, and it should come from the global config.

mlos_bench/mlos_bench/config/schemas/services/remote/azure/azure-auth-service-subschema.json
eujing: I moved the initialization for the service principal into a separate initialization function to avoid keyvault-specific patching for those general environment loading tests.

eujing: Awesome I didnt realize there were those tests supported!
Added good (full & partial) and bad JSON configs to the collection of tests.

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
eujing: Adjusted the schema abit, now if `spClientId` is provided all the other associated params will be required as well.
However if not provided, we fall back to using the user credentials as before.

mlos_bench/mlos_bench/config/schemas/services/remote/azure/azure-auth-service-subschema.json
eujing: Added a README in `serivces/remote/azure/README.md`!
Will follow up with the second PR about the scripts as mentioned too.

eujing: Just posted the follow up PR with the scripts! #480

mlos_bench/mlos_bench/config/schemas/services/remote/azure/azure-auth-service-subschema.json
```diff
@@ -23,10 +23,35 @@
...
+                            "description": "Client ID of the Service Principal to authenticate as",
+                            "type": "string"
+                        },
+                        "keyVaultName": {
+                            "description": "Name of key vault to retrieve the certificate from, for authorizing the Service Principal",
+                            "type": "string"
+                        },
+                        "certName": {
```
bpkroth, line 35: Just a note: we might want to support `secretName` for pulling secrets as passwords instead, but I think for now only supporting certs is fine.

mlos_bench/mlos_bench/services/remote/azure/README.md
```diff
@@ -0,0 +1,37 @@
+# Azure Services
+
+## Authentication
+
+This service allows us to generate tokens for authenticating with other Azure services.
```
bpkroth, line 5:
```suggestion
The `AzureAuthService` allows us to generate tokens for authenticating with other Azure services.

The goal is to be able to support long running experiments ideally without needing prompt the user to refresh their login session to avoid timeouts.
```

mlos_bench/mlos_bench/services/remote/azure/README.md
```diff
@@ -0,0 +1,37 @@
...
+It supports authenticating as the following identities:
+
+### User
+
+As a user, the only requirement is that a user is logged in via the Azure CLI before `mlos_bench` is started i.e. `az login` or `az login --device-login`.
+This method is simpler to get started, but the drawback is that organization policies might limit the period in which we can generate access tokens under a user logged in via the CLI.
```
bpkroth, line 11:
```suggestion
This method is simpler to get started, but the drawback is that security policies might limit the period in which we can generate access tokens under a user logged in via the CLI.
```

mlos_bench/mlos_bench/services/remote/azure/README.md
```diff
@@ -0,0 +1,37 @@
...
+This would lead to early termination of an experiment once tokens cannot be generated anymore without another interactive login by the user.
```
bpkroth, line 12:
```suggestion
This would lead to early termination of an experiment once tokens cannot be generated anymore without another interactive login by the user.

To avoid this issue, we also support automated authentication as a Service Principle, described next.
```

mlos_bench/mlos_bench/services/remote/azure/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Service Principal (SP)
+
+Similar to above, it also requires a user to be logged in first the Azure CLI to kick off the flow for the SP.
+However, this method addresses the drawback above to some extent as the login for a SP can be set to expire much later i.e. when the associated certificate expires in a year.
```
bpkroth, line 17:
```suggestion
However, this method addresses the drawback above to some extent as the since we can configure the system to automatically reauthenticate as the SP on demand and restrict its access to fewer resources to mitigate security concerns so that experiments can run as long as necessary.
```

mlos_bench/mlos_bench/services/remote/azure/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+The requirements of this method are:
+- User logged in to Azure CLI
+- An existing SP with appropriate access to the target resource group i.e. `Contributor` role.
```
bpkroth, line 21:
```suggestion
- An existing SP with appropriate access to the target resource group (e.g., `Contributor` role).
  > See [scripts/setup-rg/README.md](./scripts/setup-rg/README.md) for details on setting up the SP.
```

mlos_bench/mlos_bench/services/remote/azure/README.md
```diff
@@ -0,0 +1,37 @@
...
+- Key vault for storing certificate.
+- A certificate associated with the SP, which is stored in the key vault.
+- User has access to read secrets / certificates from the key vault i.e. `Key Vault Administrator` role.
```
bpkroth, line 24:
```suggestion
- User has access to read the necessary secrets/certificates from the key vault (e.g., `Key Vault Administrator` role).
```

Though we should validate whether we can get away with some lesser privilege as well.

mlos_bench/mlos_bench/services/remote/azure/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+The flow of authentication is as follows:
```
bpkroth, line 26:
```suggestion
Once the SP is setup, the flow of authentication within `mlos_bench` is as follows:
```

mlos_bench/mlos_bench/services/remote/azure/README.md
```diff
@@ -0,0 +1,37 @@
...
+1. On initialization, retrieve the SP's certificate from the key vault as the current user.
+2. Log in as the SP using the certificate
+3. During runtime when tokens are requested, we request them as the SP
```
bpkroth, line 29:
```suggestion
3. During runtime when tokens are requested, we request them as the SP, reauthenticating as necessary using the certificate which is stored only in memory.
```

mlos_bench/mlos_bench/services/remote/azure/README.md
```diff
@@ -0,0 +1,37 @@
...
+The requirements of this method are:
+- User logged in to Azure CLI
+- An existing SP with appropriate access to the target resource group i.e. `Contributor` role.
+- Key vault for storing certificate.
+- A certificate associated with the SP, which is stored in the key vault.
+- User has access to read secrets / certificates from the key vault i.e. `Key Vault Administrator` role.
+
+The flow of authentication is as follows:
+1. On initialization, retrieve the SP's certificate from the key vault as the current user.
+2. Log in as the SP using the certificate
```
bpkroth, line 28:
```suggestion
2. Log in as the SP using the certificate.
```

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
```diff
@@ -53,18 +56,60 @@ def __init__(self,
         self._access_token = "RENEW *NOW*"
         self._token_expiration_ts = datetime.datetime.now()  # Typically, some future timestamp.
 
+        # Login as ourselves
+        self._cred = azure_id.AzureCliCredential()
+
+    def _init_sp(self):
+        # Perform this initialization outside of __init__ so that environment loading tests
+        # dont need to specifically mock keyvault interactions out
```
bpkroth, line 64:
```suggestion
        # don't need to specifically mock keyvault interactions out
```

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
```diff
@@ -53,18 +56,60 @@ def __init__(self,
...
+
+        # Already logged in as SP
+        if isinstance(self._cred, azure_id.CertificateCredential):
+            return
+
+        check_required_params(
```
bpkroth, line 70: It would be good to do this early in the process if possible.
Basically, we want to have an init time config error message rather than a late runtime config error message whenever it's feasible to do so.

bpkroth: I think what you could do is make this conditional on `spClientId` being present in the config, and then issue the `check_required_params` check.

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
```diff
@@ -53,18 +56,60 @@ def __init__(self,
...
         _LOG.debug("Time to renew the token: %.2f sec.", ts_diff)
         if ts_diff < self._req_interval:
             _LOG.debug("Request new accessToken")
-            # TODO: Use azure-identity SDK and a key valut instead of `az` CLI.
-            res = json.loads(subprocess.check_output(
-                'az account get-access-token', shell=True, text=True))
-            self._token_expiration_ts = datetime.datetime.fromisoformat(res["expiresOn"])
-            self._access_token = res["accessToken"]
+            res = self._cred.get_token("https://management.azure.com/.default")
+            self._token_expiration_ts = datetime.datetime.fromtimestamp(res.expires_on)
```
bpkroth, line 112: Please validate that this is in UTC or doesn't have other sorts of timezone issues.

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
```diff
@@ -53,18 +56,60 @@ def __init__(self,
...
+        )
+
+        # The certificate private key data is stored as hidden "Secret" (not Key strangely)
+        #  in PKCS12 format, but we need to decode it.
+        secret = keyvault_secrets_client.get_secret(cert_name)
+        assert secret.value is not None
+        cert_bytes = b64decode(secret.value)
+
+        # Reauthenticate as the service principal.
+        self._cred = azure_id.CertificateCredential(tenant_id=tenant_id, client_id=sp_client_id, certificate_data=cert_bytes)
```
bpkroth, line 97: Is there an expiration on this portion, or is this sufficient to re-login automatically whenever we call the `get_token` on this?

bpkroth: Basically, wondering if the check at the top of `if isinstance(self._cred, CertificateCredential)` is sufficient to avoid re-checking the keyvault repeatedly.

eujing: @bpkroth If I understand what you mean correctly, the user's `AzureCliCredentials` will expire as usual in ~1 day (50 iterations), but on my side I have observed the SP's `CertificateCredential` last for >3 days with a run of 200 iterations without expiring and running into any errors.

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
eujing: Good point. We get the expiration time as a [Unix timestamp](https://learn.microsoft.com/en-us/python/api/azure-core/azure.core.credentials.accesstoken?view=azure-python#azure-core-credentials-accesstoken-expires-on) but the usual datetime operations are in local timezone. Changed the datetime operations to their corresponding UTC versions!

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
eujing: Makes sense! Moved the check up earlier into `__init__`

mlos_bench/mlos_bench/services/remote/azure/README.md
```diff
@@ -0,0 +1,43 @@
...
+- A certificate associated with the SP, which is stored in the key vault.
+- User has access to read the necessary secrets/certificates from the key vault (e.g., `Key Vault Administrator` role).
+
+Once the SP is setup, the flow of authentication within `mlos_bench` is as follows:
+1. On initialization, retrieve the SP's certificate from the key vault as the current user and keep it in memory.
+2. Log in as the SP using the certificate retrieved from the key vault.
+3. During runtime when tokens are requested, we request them as the SP, reauthenticating as necessary using the certificate which is stored only in memory.
+
+Although the user's login will expire earlier, it is only required on initialization.
+We retain the SP's credentials in memory while `mlos_bench` is running so that no interactive logins are required until the process ends.
```
bpkroth, line 38: These two sentences are a little redundant now given the text above.  Could probably remove them.

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
bpkroth: Sounds good.  We can adjust later if we run into issues I guess.

bpkroth (approved):
LGTM, thanks!

mlos_bench/mlos_bench/services/remote/azure/README.md
eujing: Makes sense, removed them!

bpkroth (approved):

motus (approved):
Looks good! Thank you for figuring it out!

PR #477, opened 2023-08-04
Fix SMAC optimizer logging issue

### LINKED ISSUE #475: SMAC optimizer messes up mlos_bench logging
SMAC optimizer completely overrides our logging setup and installs its own formatter, output handler, and so on. As a result, as soon as SMAC optimizer is initialized, mlos_bench stops writing to its log file, and all logging goes to stdout, in different format, and at different log level (always INFO). We need to find a way to make SMAC use our logger instead of setting up its own from scratch

motus:
By default, SMAC overrides the logging configuration and that prevents mlos_bench from writhing to a log file in proper format. This PR fixes the issue and also adds a test configuration file to run the optimizer.

Closes #475

### DIFF
--- a/mlos_bench/mlos_bench/tests/config/cli/mock-opt.jsonc
+++ b/mlos_bench/mlos_bench/tests/config/cli/mock-opt.jsonc
@@ -9,14 +9,12 @@
 
     "config_path": [
         // relative to the root of the repo (for now), where this is expected to be executed from
-        "mlos_bench/mlos_bench/tests/config"
+        "mlos_bench/mlos_bench/tests/config",
+        "mlos_bench/mlos_bench/config"
     ],
 
     "environment": "environments/mock/mock_env.jsonc",
-
-    "tunable_values": [
-        "tunable-values/tunable-values-example.jsonc"
-    ],
+    "optimizer": "optimizers/mlos_core_default_opt.jsonc",
 
     // "globals": ["global_config.json"],
 
@@ -25,6 +23,6 @@
 
     "teardown": true,
 
-    // "log_file": "mock-bench.log",
-    "log_level": "INFO"
+    // "log_file": "mock-opt.log",
+    "log_level": "DEBUG"
 }
--- a/mlos_bench/mlos_bench/tests/launcher_test.py
+++ b/mlos_bench/mlos_bench/tests/launcher_test.py
@@ -47,8 +47,6 @@ def _launch_main_app(root_path: str, local_exec_service: LocalExecService,
     with local_exec_service.temp_dir_context() as temp_dir:
 
         log_path = path_join(temp_dir, "mock-test.log")
-        if os.path.exists(log_path):
-            os.unlink(log_path)
         (return_code, _stdout, _stderr) = local_exec_service.local_exec(
             [f"./mlos_bench/mlos_bench/run.py {cli_config} --log_file '{log_path}'"],
             cwd=root_path)
--- a/mlos_bench/mlos_bench/tests/launcher_test.py
+++ b/mlos_bench/mlos_bench/tests/launcher_test.py
@@ -46,7 +46,9 @@ def _launch_main_app(root_path: str, local_exec_service: LocalExecService,
     """
     with local_exec_service.temp_dir_context() as temp_dir:
 
-        log_path = path_join(temp_dir, "mock-bench.log")
+        log_path = path_join(temp_dir, "mock-test.log")
+        if os.path.exists(log_path):
+            os.unlink(log_path)
         (return_code, _stdout, _stderr) = local_exec_service.local_exec(
             [f"./mlos_bench/mlos_bench/run.py {cli_config} --log_file '{log_path}'"],
             cwd=root_path)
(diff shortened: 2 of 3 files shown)


mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -124,6 +124,7 @@ def __init__(self, *,  # pylint: disable=too-many-locals
             random_design=random_design,
             config_selector=config_selector,
             overwrite=True,
+            logging_level=False,  # Use the existing logger
```
bpkroth, line 127: Since it's supposed to be possible to use `mlos_core` independently of `mlos_bench` can we:
- in `mlos_core` extend the constructor params to accept a `logging_level`
  - if it's `None` (default), then issue a warning that `SMAC` will be interacting with the logger somehow
- in `mlos_bench` we set that value for them by default

Thoughts?

motus: By the way, that SMAC `logging_level` parameter is actually quite a mess! I've looked at the source code, and it can actually be of one of *four* different types:
* `False`: do not configure the logger (i.e., use the default / already configured logger)
* `None`: reconfigure the logger with SMAC defaults (use INFO level, write to stdout only, reset the output format, etc.)
* `int`: (e.g., `20`): reconfigure the logger with SMAC defaults, but set the logging level to the given threshold
* `str`: A path to the YAML configuration file (!) that contains logging parameters.

Now imagine every optimizer that we use in `mlos_core` being so creative about their logging! We'll end up implementing our own adapters to configure each optimizer's logging, etc., etc. I would much prefer each optimizer rely on some preconfigured logging instead. Every other way is just not composable

bpkroth: Ack.  You've convinced me.

mlos_bench/mlos_bench/tests/config/cli/mock-opt.jsonc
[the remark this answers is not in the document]
motus: yes, we can easily verify the log format in the `launcher_test` - I will add the check to this PR now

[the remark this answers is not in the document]
bpkroth: FYI, I cherry picked a little of this into #470 and augmented this config to use a `global_test_config.jsonc` that overrides the default number of iterations for testing purposes.

mlos_bench/mlos_bench/tests/launcher_test.py
bpkroth, line 1: This is still only using the `mlos-bench.jsonc` config, right?
If we borrow those other changes from #470, then we can also run `mock-opt.jsonc` for a small number of iterations as a separate mock end-to-end check as well.

bpkroth (approved):
Approved for now.  I can add that other config as a test in #470 next.

mlos_bench/mlos_bench/tests/launcher_test.py
motus: oh, that's right! totally forgot about it. will add more unit tests now

mlos_bench/mlos_bench/tests/launcher_test.py
```diff
@@ -37,27 +39,70 @@ def local_exec_service() -> LocalExecService:
...
-                ln for ln in best_score_lines
-                if " best score: 65.67" in ln
-            ]) == 1
+        try:
+            iter_expected = iter(re_expected)
+            re_log = re.compile(next(iter_expected))
+            with open(log_path, "rt", encoding="utf-8") as fh_out:
+                for ln in fh_out:
+                    if re_log.match(ln):
+                        re_log = re.compile(next(iter_expected))
```
bpkroth, line 62: This opens and reads the file multiple times for each regex.

What about something more like the following:

```python
regexes = {pattern: re.compile(pattern) for pattern in patterns}
with open(...) as fh:
   for line in fh:
      for pattern in regexes:
          regex = regexes[pattern]
          if regex.match(line):
              del(regexes[pattern])
      if not regexes:
          break
if regexes:
    raise Exception("The following regexes were not matched: " + str(regexes))
```

mlos_bench/mlos_bench/tests/launcher_test.py
```diff
@@ -37,27 +39,70 @@ def local_exec_service() -> LocalExecService:
...
     """
     with local_exec_service.temp_dir_context() as temp_dir:
 
-        log_path = path_join(temp_dir, "mock-bench.log")
-        cmd = "./mlos_bench/mlos_bench/run.py" + \
-              " --config mlos_bench/mlos_bench/tests/config/cli/mock-bench.jsonc" + \
-              f" --log_file '{log_path}'"
-        (return_code, _stdout, _stderr) = local_exec_service.local_exec([cmd], cwd=root_path)
-
+        log_path = path_join(temp_dir, "mock-test.log")
```
bpkroth, line 50:
```suggestion
        # Test developers note: for local debugging,
        # uncomment the following line to use a known file path that can be examined:
        # temp_dir = '/tmp'
        log_path = path_join(temp_dir, "mock-test.log")
```

mlos_bench/mlos_bench/tests/launcher_test.py
bpkroth: Nm, I missed the `next(iter_...` part *inside* the line loop.
Yours is actually even better than mine since you only need to evaluate a single regex (with expected order) for each line during the leading portion of the file.
Kudos!

PR #478, opened 2023-08-07
Implement "read_telemetry_file" feature

motus:
Summary of changes:
* Add `"read_telemetry_file"` parameter to `LocalEnv` implement the telemetry reading functionality
* Allow failures when downloading (intermediate) files in `LocalFileShareEnv`
* Translate Azure-specific exceptions in `AzureFileShareService` to the generic ones to keep Environment classes cloud-agnostic
* Bugfix: implement `CompositeEnv.status()` properly

**TODO:**
* [x] Fix the JSON schema to support `"read_telemetry_file"`

### DIFF
--- a/mlos_bench/mlos_bench/environments/README.md
+++ b/mlos_bench/mlos_bench/environments/README.md
@@ -7,7 +7,7 @@ An [`Environment`](./base_environment.py) is a class that represents a system th
 It is responsible for setting up the system, running the benchmark, and tearing down the system.
 Each `Environment` object also keeps track of the current state of the system, and can be used to query the system for metrics.
 
-Environments have [`Tunable`](../tunables/tunable.py) parameters and [`TunableGroups`](../tunables/tunable_groups.py) for controlling their configuration.
+Environments can have [`Tunable`](../tunables/tunable.py) parameters and [`TunableGroups`](../tunables/tunable_groups.py) for controlling their configuration.
 It is generally expected that all [`Tunable`](../tunables/tunable.py) parameters within an `Environment` will have the same cost to change.
 
 Common and platform-specific functionality of the environments is implemented in [`Service`](../services/) classes.
--- a/mlos_bench/mlos_bench/environments/README.md
+++ b/mlos_bench/mlos_bench/environments/README.md
@@ -20,7 +20,7 @@ Each Environment has several stages that it goes through:
 - `run`
 - `teardown`
 
-One cal also query the current state of the system via the `.status()` method.
+One can also query the current state of the system via the `.status()` method.
 Our current implementation of the `Environment` classes is synchronous; that means, a `.status()` method can only be used *after* `.run()` method has been called.
 
 Once we implement an asynchronous mode of operation, the `.status()` method will be usable at any time during the `Environment` object lifecycle.
--- a/mlos_bench/mlos_bench/environments/README.md
+++ b/mlos_bench/mlos_bench/environments/README.md
@@ -4,7 +4,8 @@
 
 This directory contains the code for the [`Environment`](./base_environment.py) classes used in the [`mlos_bench`](../../../mlos_bench/) benchmarking automation framework.
 An [`Environment`](./base_environment.py) is a class that represents a system that can be benchmarked.
-It is responsible for setting up the system, running the benchmark, and tearing down the system.
+It is responsible for setting up a portion of the system, running the benchmark (or some other command), and tearing down that portion of the system.
+A `CompositeEnvironment` can be used to stack these together to represent the entire system under evaluation (e.g., VM, OS, App, etc.)
 Each `Environment` object also keeps track of the current state of the system, and can be used to query the system for metrics.
 
 Environments can have [`Tunable`](../tunables/tunable.py) parameters and [`TunableGroups`](../tunables/tunable_groups.py) for controlling their configuration.
--- a/mlos_bench/mlos_bench/environments/README.md
+++ b/mlos_bench/mlos_bench/environments/README.md
@@ -10,6 +10,7 @@ Each `Environment` object also keeps track of the current state of the system, a
 
 Environments can have [`Tunable`](../tunables/tunable.py) parameters and [`TunableGroups`](../tunables/tunable_groups.py) for controlling their configuration.
 It is generally expected that all [`Tunable`](../tunables/tunable.py) parameters within an `Environment` will have the same cost to change.
+> It may therefore make sense to split a portion of the system logically across `Environments` (e.g., boot-time vs. runtime settings).
 
 Common and platform-specific functionality of the environments is implemented in [`Service`](../services/) classes.
 
--- a/mlos_bench/mlos_bench/environments/README.md
+++ b/mlos_bench/mlos_bench/environments/README.md
@@ -1,6 +1,18 @@
 # Environments
 
-This directory contains the code for the `Environments` used in the [`mlos_bench`](../../../mlos_bench/) benchmarking automation framework.
+## Overview
+
+This directory contains the code for the [`Environment`](./base_environment.py) classes used in the [`mlos_bench`](../../../mlos_bench/) benchmarking automation framework.
+An [`Environment`](./base_environment.py) is a class that represents a system that can be benchmarked.
+It is responsible for setting up the system, running the benchmark, and tearing down the system.
+Each `Environment` object also keeps track of the current state of the system, and can be used to query the system for metrics.
+
+Environments have [`Tunable`](../tunables/tunable.py) parameters and [`TunableGroups`](../tunables/tunable_groups.py) for controlling their configuration.
+It is generally expected that all [`Tunable`](../tunables/tunable.py) parameters within an `Environment` will have the same cost to change.
+
+Common and platform-specific functionality of the environments is implemented in [`Service`](../services/) classes.
+
+## Lifecycle
 
 Each Environment has several stages that it goes through:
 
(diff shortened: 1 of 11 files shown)


mlos_bench/mlos_bench/environments/composite_env.py
```diff
@@ -203,3 +204,28 @@ def run(self) -> Tuple[Status, Optional[dict]]:
...
+        if not status.is_ready():
+            return (status, telemetry)
+
+        joint_telemetry = []
+        for env_context in self._child_contexts:
+            (status, telemetry) = env_context.status()
+            joint_telemetry.extend(telemetry)
+            _LOG.debug("Child env. status: %s :: %s", env_context, status)
+            if not status.is_good():
+                break
```
bpkroth, line 228: Shouldn't we continue to accumulate the telemetry from additional children, but still return the worst status?

mlos_bench/mlos_bench/environments/local/local_env.py
```diff
@@ -187,6 +192,30 @@ def run(self) -> Tuple[Status, Optional[dict]]:
...
+                    self._read_telemetry_file, extra_paths=[self._temp_dir]))
+        except FileNotFoundError as ex:
+            _LOG.warning("Telemetry CSV file not found: %s :: %s", self._read_telemetry_file, ex)
+            return (status, [])
+
+        _LOG.debug("Read telemetry data:\n%s", data)
+        if list(data.columns) != ["timestamp", "metric", "value"]:
+            _LOG.warning(
+                'Telemetry CSV file should have columns ["timestamp", "metric", "value"] :: %s',
+                self._read_telemetry_file)
```
bpkroth, line 215: Could also support non-headered CSVs if the columns schema looks to still line up, right?

mlos_bench/mlos_bench/environments/local/local_env.py
bpkroth, line 1: Possible for us to have any tests for this?

mlos_bench/mlos_bench/run.py
```diff
@@ -149,13 +149,14 @@ def _run(env_context: Environment, opt: Optimizer,
         opt.register(trial.tunables, Status.FAILED)
         return
 
+    (status, results) = env_context.run()  # Block and wait for the final result.
+    _LOG.info("Results: %s :: %s\n%s", trial.tunables, status, results)
+
     # In async mode (TODO), poll the environment for status and telemetry
     # and update the storage with the intermediate results.
-    (status, telemetry) = env_context.status()
+    (_, telemetry) = env_context.status()
```
bpkroth, line 157: Why is status dropped here?
We update it in the next line, right?

mlos_bench/mlos_bench/environments/local/local_env.py
```diff
@@ -187,6 +192,30 @@ def run(self) -> Tuple[Status, Optional[dict]]:
         _LOG.info("Local run complete: %s ::\n%s", self, data_dict)
         return (Status.SUCCEEDED, data_dict) if data_dict else (Status.FAILED, None)
 
+    def status(self) -> Tuple[Status, List[Tuple[datetime, str, Any]]]:
```
bpkroth, line 195: We should have some documentation in a `README.md` somewhere that explains how this expected to work.  For instance:

```md
## Environment

Generally something that's tunable ...

### Lifecycle

1. `setup`
2. `run`
3. Async `status` polling during `run` phase.
    - Environments can optionally provide a `status` command in their config that can be executed to periodically generate telemetry info in csv format into a file in the shared storage area to be consumed asynchornously, or else return it on `stdout` to be consumed directly.
    - Data will then be stored in the DB for later analysis.
5. Results reporting/processing.
6. `teardown`
```

bpkroth: Obviously a lot more to be filled in here.

mlos_bench/mlos_bench/environments/composite_env.py
motus: yes, I was thinking about it, too. we might still need to know what environment has that status, though.. anyway, let's just collect as much data as possible and we can think about the rest later

also, now when I am looking at the code - we should probably do the same for the final metrics. rn we return the data only from the _last_ environment, but it might be good to collect different metrics from different environments. One example is OtterTune's microbenchmark to gauge the IO performance before running the actual DB trial. We might not need to collect all that in one environment

mlos_bench/mlos_bench/environments/local/local_env.py
motus: yep, that's why I only print a warning and then proceed with reading the data

mlos_bench/mlos_bench/run.py
motus: yes, we just use the status from `.run()` that we get a few lines above. Once we switch to the async mode, we'll query the `.status()` while `.run()` is still in progress, and *then* we'll use the status from it

mlos_bench/mlos_bench/environments/local/local_env.py
motus: yes, I'll do that with the async version. rn I just want a working implementation of `.status()` and `.update_telemetry()` so I have something to use for debugging when working on the async stuff

mlos_bench/mlos_bench/environments/composite_env.py
bpkroth: Yes, exactly.  This was a bit of what I was talking about for providing precanned system metrics collection environments.

mlos_bench/mlos_bench/run.py
bpkroth: Maybe add a small comment to that effect.

mlos_bench/mlos_bench/environments/local/local_env.py
bpkroth: That's understandable.  I guess it'd help me while reviewing PRs to see what the overall design goal is, even if it's incomplete for now.

mlos_bench/mlos_bench/environments/local/local_env.py
motus: uh, let's do it in a separate PR. Turns out, we don't have any tests for `LocalEnv` at all (only for `local_exec` service). I'll add tests for `LocalEnv` and check both `"read_telemetry_file"` and `"read_results_file"` there

mlos_bench/mlos_bench/environments/README.md
```diff
@@ -1,20 +1,32 @@
...
+## Lifecycle
 
 Each Environment has several stages that it goes through:
 
 - `setup`
 - `run`
 - `teardown`
 
-Which are implemented using [`Services`](../services/).
+One cal also query the current state of the system via the `.status()` method.
```
bpkroth, line 23:
```suggestion
One can also query the current state of the system via the `.status()` method.
```

mlos_bench/mlos_bench/environments/README.md
```diff
@@ -1,20 +1,32 @@
 # Environments
 
-This directory contains the code for the `Environments` used in the [`mlos_bench`](../../../mlos_bench/) benchmarking automation framework.
+## Overview
+
+This directory contains the code for the [`Environment`](./base_environment.py) classes used in the [`mlos_bench`](../../../mlos_bench/) benchmarking automation framework.
+An [`Environment`](./base_environment.py) is a class that represents a system that can be benchmarked.
+It is responsible for setting up the system, running the benchmark, and tearing down the system.
```
bpkroth, line 7:
```suggestion
It is responsible for setting up a portion of the system, running the benchmark (or some other command), and tearing down that portion of the system.
A `CompositeEnvironment` can be used to stack these together to represent the entire system under evaluation (e.g., VM, OS, App, etc.)
```

bpkroth: And maybe remove the line before it.

mlos_bench/mlos_bench/environments/README.md
```diff
@@ -1,20 +1,32 @@
...
+Each `Environment` object also keeps track of the current state of the system, and can be used to query the system for metrics.
+
+Environments have [`Tunable`](../tunables/tunable.py) parameters and [`TunableGroups`](../tunables/tunable_groups.py) for controlling their configuration.
```
bpkroth, line 10:
```suggestion
Environments can have [`Tunable`](../tunables/tunable.py) parameters and [`TunableGroups`](../tunables/tunable_groups.py) for controlling their configuration.
```

mlos_bench/mlos_bench/environments/README.md
```diff
@@ -1,20 +1,32 @@
...
+It is generally expected that all [`Tunable`](../tunables/tunable.py) parameters within an `Environment` will have the same cost to change.
```
bpkroth, line 11:
```suggestion
It is generally expected that all [`Tunable`](../tunables/tunable.py) parameters within an `Environment` will have the same cost to change.
> It may therefore make sense to split a portion of the system logically across `Environments` (e.g., boot-time vs. runtime settings).
```

mlos_bench/mlos_bench/environments/composite_env.py
```diff
@@ -192,14 +193,46 @@ def run(self) -> Tuple[Status, Optional[dict]]:
...
+            return (status, telemetry)
+
+        joint_telemetry = []
+        final_status = None
+        for env_context in self._child_contexts:
+            (status, telemetry) = env_context.status()
+            _LOG.debug("Child env. status: %s :: %s", env_context, status)
+            joint_telemetry.extend(telemetry)
+            if not status.is_good() and final_status is None:
+                final_status = status
```
bpkroth, line 234: Wonder if there's a `max(final_status, status)` version of this that propogates the "worst" status seen, but I think this is good enough for now.

bpkroth (approved):

mlos_bench/mlos_bench/environments/composite_env.py
motus: I was thinking about it - but the meaning of "best" and "worst" status can vary across different contexts.. it is very likely I'll get back to this issue, though - when we implement the environment resetting properly

PR #479, opened 2023-08-09
Random seed improvements

bpkroth:
Porting a subset of changes from #470 to their own PR for independent testing.

- Pass seed along to ConfigSpace creation when converting.
- Set `np.random.seed` inside flaml per recommendations.
- Make sure RandomSearch in SMAC passes seed along from config.
- Consolidate seed specifications in `mlos_bench` tests

### DIFF
--- a/mlos_bench/mlos_bench/optimizers/base_optimizer.py
+++ b/mlos_bench/mlos_bench/optimizers/base_optimizer.py
@@ -58,7 +58,7 @@ class Optimizer(metaclass=ABCMeta):     # pylint: disable=too-many-instance-attr
 
     def __repr__(self) -> str:
         opt_direction = 'min' if self._opt_sign > 0 else 'max'
-        return f"{self.__class__.__name__}:{opt_direction}({self._opt_target})"
+        return f"{self.__class__.__name__}:{opt_direction}({self._opt_target})(config={self._config})"
 
     @property
     def target(self) -> str:
--- a/mlos_bench/mlos_bench/optimizers/convert_configspace.py
+++ b/mlos_bench/mlos_bench/optimizers/convert_configspace.py
@@ -58,7 +58,7 @@ def _tunable_to_hyperparameter(
         raise TypeError(f"Undefined Parameter Type: {tunable.type}")
 
 
-def tunable_groups_to_configspace(tunables: TunableGroups) -> ConfigurationSpace:
+def tunable_groups_to_configspace(tunables: TunableGroups, seed: Optional[int] = None) -> ConfigurationSpace:
     """
     Convert TunableGroups to  hyperparameters in ConfigurationSpace.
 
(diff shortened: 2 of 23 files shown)


mlos_bench/mlos_bench/optimizers/base_optimizer.py
```diff
@@ -58,7 +58,7 @@ def __init__(self,
 
     def __repr__(self) -> str:
         opt_direction = 'min' if self._opt_sign > 0 else 'max'
-        return f"{self.__class__.__name__}:{opt_direction}({self._opt_target})"
+        return f"{self.__class__.__name__}:{opt_direction}({self._opt_target})(config={self._config})"
```
bpkroth, line 61: Include the seed and other config details when we print it.

mlos_bench/mlos_bench/optimizers/convert_configspace.py
```diff
@@ -67,12 +67,15 @@ def tunable_groups_to_configspace(tunables: TunableGroups) -> ConfigurationSpace
...
+    seed : Optional[int]
+        Random seed to use.
+
     Returns
     -------
     configspace : ConfigurationSpace
         A new ConfigurationSpace instance that corresponds to the input TunableGroups.
     """
-    space = ConfigurationSpace()
+    space = ConfigurationSpace(seed=seed)
```
bpkroth, line 78: Include the seed when we convert a config space.

mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
```diff
@@ -52,9 +55,12 @@ def __init__(self,
             optimizer_type=opt_type,
             optimizer_kwargs=self._config,
             space_adapter_type=space_adapter_type,
-            space_adapter_kwargs=space_adapter_config
+            space_adapter_kwargs=space_adapter_config,
         )
 
+    def __repr__(self) -> str:
+        return f"{super().__repr__()}({self._opt.__class__.__name__})"
```
bpkroth, line 62: Include the mlos_core optimizer type in our string form.

mlos_bench/mlos_bench/tests/environments/mock_env_test.py
```diff
@@ -25,7 +25,7 @@ def test_mock_env_default(mock_env: MockEnv, tunable_groups: TunableGroups) -> N
         (status, data) = env_context.run()
         assert status.is_succeeded()
         assert data is not None
-        assert data["score"] == pytest.approx(93.22, 0.01)
+        assert data["score"] == pytest.approx(72.92, 0.01)
```
bpkroth, line 28: Consolidated seed.

mlos_bench/mlos_bench/tests/__init__.py
```diff
@@ -12,6 +12,11 @@
 from mlos_bench.util import get_class_from_name
 
 
+# A common seed to use to avoid tracking down race conditions and intermingling
+# issues of seeds across tests that run in non-deterministic parallel orders.
+SEED = 42
```
bpkroth, line 17: In a real running version of mlos_bench, I don't think we'd ever have more than one seed running at a time, so while this may mask places where we might not be setting the seed correctly everywhere, on the other hand I think it's OK to set it to the same value as long as it happens towards the top.
Can consider to do this inside launch.py or run.py in the future as well if we want.

motus (approved):
Looks good! Thank you for figuring it out!

PR #480, opened 2023-08-09
Automate Control Plane RG Setup

eujing:
This PR adds an initial script for automating the set up of the control plane resource group.
More details in the included README.


scripts/setup-rg/setup-rg.ps1
bpkroth, line 1: Let's have an `sh` version as well, or even better, a shell syntax agnostic one.

scripts/setup-rg/rg-template.json
```diff
@@ -0,0 +1,376 @@
...
+        "properties": {
+            "securityRules": [
+                {
+                    "name": "CorpSSH",
+                    "properties": {
+                        "description": "Corp. IP SSH access",
+                        "protocol": "Tcp",
+                        "sourcePortRange": "*",
+                        "destinationPortRange": "22",
+                        "sourceAddressPrefix": "131.107.0.0/16",
```
bpkroth, line 80: These IPs should not be exposed externally.  Let's keep that inside some `env.sh` files that can be sourced locally to help parameterize which IPs to by default allow connections from.
Ideally, automatically adding the caller IP would be very nice too.

scripts/setup-rg/README.md
bpkroth, line 1: Proposal:
Let's reorg all of these into the following structure:

```
mlos_core/mlos_core/config/services/remote/azure/scripts/setup-rg/...
```

A convenience could also be to add a symlink:

```
scripts/azure/setup-rg -> ../../mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/
```

bpkroth: Should also add a link to this README.md from one of the other ones (e.g. `mlos_bench/mlos_bench/README.md`)

scripts/setup-rg/README.md
```diff
@@ -0,0 +1,66 @@
...
+1. Starting in this current directory, ensure that we are logged in to Azure CLI.
+
+    ```sh
+    az login
+    ```
+
+2. Make a copy of the ARM parameters file.
+
+    ```sh
+    cp rg-template.example.parameters.json rg-template.<your name>.parameters.json
```
bpkroth, line 16:
```suggestion
    cp rg-template.example.parameters.json rg-template.parameters.json
```

scripts/setup-rg/README.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+where `$armParams` follows the same usage as `--parameters` in [az deployment group create](https://learn.microsoft.com/en-us/cli/azure/deployment/group?view=azure-cli-latest#az-deployment-group-create-examples).
+
+## Workflow
+
+The high-level flow for what this script automates is as follows:
+
+1. Assign `Contributor` access to the Service Principal (SP) for write access over resources.
+    For now we assume the experiment's resources are provisioned in the same RG as the control plane, so the access is granted to the same RG.
```
bpkroth, line 51: We probably need a PR for the control plane setup too.

scripts/setup-rg/README.md
```diff
@@ -0,0 +1,66 @@
+# Resource Group Setup
+
+These scripts are helpers for setting up a new resource group (RG) as the control plane for MLOS.
```
bpkroth, line 3:
```suggestion
These scripts are helpers for setting up a new resource group (RG) in Azure as the *control plane* for MLOS.

The *control plane RG* is a container for the *persistent* resources of MLOS (results/metrics storage, scheduler VM, notebook interface, etc.).
```

scripts/setup-rg/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+2. Make a copy of the ARM parameters file.
+
+    ```sh
+    cp rg-template.example.parameters.json rg-template.<your name>.parameters.json
+    ```
+
+3. Modify the ARM parameters in the newly created file as needed, especially the `PLACEHOLDER` values.
+
+4. Execute the main script and follow prompts:
```
bpkroth, line 21: Let's not give directions that rely on prompts.  Only CLI args.  Thanks!

scripts/setup-rg/README.md
```diff
@@ -0,0 +1,66 @@
...
+    -certName $certName
+```
+
+where `$armParams` follows the same usage as `--parameters` in [az deployment group create](https://learn.microsoft.com/en-us/cli/azure/deployment/group?view=azure-cli-latest#az-deployment-group-create-examples).
+
+## Workflow
+
+The high-level flow for what this script automates is as follows:
+
+1. Assign `Contributor` access to the Service Principal (SP) for write access over resources.
```
bpkroth, line 50: Does it optionally create the SP?

scripts/setup-rg/README.md
bpkroth, line 51:
```suggestion
    Ideally, experiment resources are placed in their own RG.
    When that isn't possible, they can also be placed in the control plane RG, in which case the SP can optionally be given access to the control plane RG as well.
```

bpkroth: With associated script fixups to make that indeed optional.

scripts/setup-rg/README.md
```diff
@@ -0,0 +1,66 @@
...
+    For now we assume the experiment's resources are provisioned in the same RG as the control plane, so the access is granted to the same RG.
+
+2. Provision control plane resources into the RG.
+    This includes:
+    - Networking (public IP, security group, vnet, subnet, network interface)
```
bpkroth, line 55: Most of these probably shouldn't be there in the control plane.

scripts/setup-rg/README.md
```diff
@@ -0,0 +1,66 @@
...
+    - Key Vault
+    - Storage (storage account, file share, MySQL Flex server)
+    - VM
```
bpkroth, line 58:
```suggestion
    - VM for running the `mlos_bench` scheduler.
```

scripts/setup-rg/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+The high-level flow for what this script automates is as follows:
+
+1. Assign `Contributor` access to the Service Principal (SP) for write access over resources.
+    For now we assume the experiment's resources are provisioned in the same RG as the control plane, so the access is granted to the same RG.
+
+2. Provision control plane resources into the RG.
+    This includes:
+    - Networking (public IP, security group, vnet, subnet, network interface)
+    - Key Vault
```
bpkroth, line 56:
```suggestion
    - Key Vault for storing the SP credentials.
```

scripts/setup-rg/README.md
bpkroth: Actually, the NSG might be able to be reused across both experiment resources and the control plane.  We'd have to test it.
But not across locations, so still probably better to just keep them separate.

scripts/setup-rg/README.md
```diff
@@ -0,0 +1,66 @@
...
+    - Storage (storage account, file share, MySQL Flex server)
```
bpkroth, line 57:
```suggestion
    - Storage (storage account, file share, DB)
```

- DB need not be MySQL, so we should make that optional as well (can certainly default to MySQL though and let it get parameterized).
- Storage needs to be location dependent, and embedded in the name.
  As such, we should make sure this script can be idempotent or split so that we can independently setup storage in multiple locations without it complaining about the DB portion or scheduler VM already existing.

In fact as far as that goes, we could even have multiple scheduler VMs at some point.

scripts/setup-rg/README.md
```diff
@@ -0,0 +1,66 @@
...
+    - VM
+
+3. Assign `Key Vault Administrator` access to the current user.
+    This allows the current user to retrieve secrets / certificates from the VM once it is set up.
+    Ensure to log in as the same user in the VM.
+
+4. If not existing yet, generate a certificate with `certName` in the key vault.
```
bpkroth, line 64: Shouldn't the script either allow us to:
- create the SP and store its generated cert in the keyvault, or
- reference an existing SP with provided cert?

I think the middle case someone else created the SP, but then we reset its password/cert and store that in the keyvault could break other uses of the SP potentially.

scripts/setup-rg/rg-template.example.parameters.json
```diff
@@ -0,0 +1,30 @@
...
+            "value": "mlos-mysql-autotune"
+        },
+        "vmSKU": {
+            "value": "Standard_D2s_v3"
+        },
+        "vmAdminUsername": {
+            "value": "PLACEHOLDER"
+        },
+        "sshPublicKey": {
+            "value": "PLACEHOLDER"
```
bpkroth, line 15: Can we make this a list?  e.g., in case multiple users want to login?

scripts/setup-rg/rg-template.example.parameters.json
```diff
@@ -0,0 +1,30 @@
...
+        },
+        "fileShareName": {
+            "value": "mlos-file-share"
+        },
+        "resultsDbName": {
+            "value": "mlos-mysql-autotune"
```
bpkroth, line 21: Let's also strip `mysql` from these template names.

scripts/setup-rg/rg-template.example.parameters.json
```diff
@@ -0,0 +1,30 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentParameters.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "projectPrefix": {
+            "value": "mlos-mysql-autotune"
```
bpkroth, line 6:
```suggestion
            "value": "mlos-autotune"
```

Here's a cool thing about this all: we don't need to have separate storage/db/scheduler for different projects, so I don't think we should embed the name in the control pane portion of things.

The project name can be a part of a set of experiments, but basically each group that wants to run an mlos instance could share it across projects, or even across an org in order to pool benchmark data for wide spread analysis.
That's something GSL has talked about for instance, despite the fact that we may work on many different target projects.

Make sense?

scripts/setup-rg/rg-template.json
```diff
@@ -0,0 +1,376 @@
...
+        },
+        "resultsDbAdminPassword": {
+            "type": "securestring"
+        }
+    },
+    "functions": [],
+    "variables": {
+        "storageAccountName": "[toLower(concat(replace(parameters('projectPrefix'), '-', ''), 'storage'))]",
+        "vmName": "[concat(parameters('projectPrefix'), '-vm')]",
+        "networkInterfaceName": "[concat(variables('vmName'), '-ni-', uniqueString(resourceGroup().id))]",
```
bpkroth, line 44: why append the RG to the NIC name?

scripts/setup-rg/rg-template.json
```diff
@@ -0,0 +1,376 @@
...
+        "location": "[resourceGroup().location]",
+        "properties": {
+            "securityRules": [
+                {
+                    "name": "CorpSSH",
+                    "properties": {
+                        "description": "Corp. IP SSH access",
+                        "protocol": "Tcp",
+                        "sourcePortRange": "*",
+                        "destinationPortRange": "22",
```
bpkroth, line 79: Internally, we'll also need to open to some other alt port, so please make this a variable that we can change via the template params too.

scripts/setup-rg/rg-template.json
```diff
@@ -0,0 +1,376 @@
...
+        "type": "Microsoft.Network/networkInterfaces",
+        "apiVersion": "2020-11-01",
+        "location": "[resourceGroup().location]",
+        "dependsOn": [
+            "[resourceId('Microsoft.Network/publicIPAddresses', variables('publicIpName'))]",
+            "[resourceId('Microsoft.Network/virtualNetworks', variables('vnetName'))]",
+            "[resourceId('Microsoft.Network/networkSecurityGroups', variables('nsgName'))]"
+        ],
+        "tags": {
+            "displayName": "ubuntuVM1-NetworkInterface"
```
bpkroth, line 141: ?

scripts/setup-rg/rg-template.json
```diff
@@ -0,0 +1,376 @@
...
+            "displayName": "[variables('storageAccountName')]"
+        },
+        "location": "[resourceGroup().location]",
+        "dependsOn": [
+            "[resourceId('Microsoft.Network/virtualNetworks', variables('vnetName'))]"
+        ],
+        "kind": "StorageV2",
+        "sku": {
+            "name": "Standard_LRS",
+            "tier": "Standard"
```
bpkroth, line 266: May want this to be optionally premium as well, so please provide a default of standard and option to override in the template params.

scripts/setup-rg/rg-template.json
```diff
@@ -0,0 +1,376 @@
...
+    {
+        "type": "Microsoft.DBforMySQL/flexibleServers/firewallRules",
+        "apiVersion": "2021-12-01-preview",
+        "name": "[concat(parameters('resultsDbName'), '/CorpIps')]",
+        "dependsOn": [
+            "[resourceId('Microsoft.DBforMySQL/flexibleServers', parameters('resultsDbName'))]"
+        ],
+        "properties": {
+            "startIpAddress": "131.107.0.0",
+            "endIpAddress": "131.107.255.255"
```
bpkroth, line 348: Same issue about embeded IPs.

scripts/setup-rg/rg-template.example.parameters.json
bpkroth: Could call it `mlos-autotune-db`

scripts/setup-rg/README.md
bpkroth: Nm, in the templates that gets hard to support.

We can stick with a MySQL Flex DB for now, but let's do the following:

1. Separate out the template so that some of these resources can be deployed, in an appropriate order, separately.
2. Update the script to allow skip deploying the DB in case someone wants to provide an alternative implementation option there.  Our default supported implementation can be MySQL for now.
3. Label it generically as "results DB" instead.

scripts/setup-rg/README.md
motus, line 1: [services/remote/azure/README.md](../../blob/main/mlos_bench/mlos_bench/services/remote/azure/README.md) refers to this README - can you add a link to this file URL instead of just mentioning a file name?

Also, maybe describe it at [mlos_bench/README.md](../../blob/main/mlos_bench/README.md) because this is where we outline all steps required to run `mlos_bench`

scripts/setup-rg/README.md
eujing: Now optionally creates the SP!

scripts/setup-rg/README.md
eujing: - Split out an ARM template for the results DB that users can optionally provision.
- Kept storage in the original template, but now allows for users to set a list of locations to create one storage for each, with location is appended to the storage name. This was ease the updating of the control vnet's service endpoints too.

scripts/setup-rg/README.md
eujing: I agree! Now the flow is:
- If cert name does not exist in the keyvault already, create or update a SP with the access, generate a cert and store in the keyvault.
- Otherwise, create or update a SP with the access only, and verify that the SP's cert is linked with the one in the keyvault already via thumbprint.

scripts/setup-rg/rg-template.example.parameters.json
eujing: Makes sense! Accidentally left `mysql` in a bunch of these names, but it is intended to be generic.

scripts/setup-rg/rg-template.json
eujing: It seemed to be a pattern from the ARM documentation, wasnt sure if NI names needed to be unique this way. I realize in other templates we have this was not needed, I will look into removing this!

scripts/setup-rg/rg-template.json
eujing: Good catch! Replaced with the actual name.

scripts/setup-rg/rg-template.json
eujing: Good point, replaced with a parameter that is standard by default!

scripts/setup-rg/rg-template.example.parameters.json
eujing: Made it into a list!

scripts/setup-rg/rg-template.json
eujing: Made it into a parameter!

scripts/setup-rg/README.md
eujing: I agree! I meant for these to be solely for the control VM, and experiments will set up their own as needed. Will rephrase this to `Control VM's networking`

scripts/setup-rg/rg-template.json
eujing: Parameterized all these specific IPs into the ARM parameter files!

scripts/setup-rg/README.md
bpkroth: On the last point, will that still work if we want to add additional locations after the fact, or do they all needed to be added at the initial creation time?

eujing: It will still work after the fact, it will incrementally add the storages!

scripts/setup-rg/rg-template.json
eujing: 1. In the update, this has been renamed to "UserSSH"!
2. It seems only singletons, ranges or "*" are supported with `destinationPortRange`. However there is also a list version `destinationPortRanges` that can accept multiple of these. Do we want to support that instead?

scripts/setup-rg/setup-rg.ps1
eujing: Added the `sh` version!

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Quickstart
+
+1. Starting in this current directory, ensure that we are logged in to Azure CLI.
+
+    ```sh
+    az login
+    ```
+
+1. Make a copy of the control plane ARM parameters file.
```
bpkroth, line 14: The numbering on these are all `1.` which won't render well.  Please have a look in the markdown linter inside the devcontainer.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+    ```sh
+    cp rg-template.example.parameters.json rg-template.parameters.json
+    ```
+
+1. (Optional) Make a copy of the results DB parameters file, if planning to provision a results DB.
```
bpkroth, line 20:
```suggestion
3. (Optional) Make a copy of the results DB parameters file, if planning to provision a results DB (suggested).
```

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/README.md
```diff
@@ -0,0 +1,68 @@
...
+    ```sh
+    cp results-db/mysql-template.parameters.example.json results-db/mysql-template.parameters.json
+    ```
+
+1. Modify the ARM parameters in the newly created files as needed, especially the `PLACEHOLDER` values.
+
+1. Execute the main script with CLI args as follows:
+
+    ```shell
+    ./setup-rg.ps1 `
```
bpkroth, line 31: Please provide alternative examples for sh version as well.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Workflow
+
+The high-level flow for what this script automates is as follows:
+
+1. Assign `Contributor` access to the Service Principal (SP) for write access over resources.
+    Ideally, experiment resources are placed in their own RG.
+    When that isn't possible, they can also be placed in the control plane RG, in which case the SP can optionally be given access to the control plane RG as well.
+
+1. Provision control plane resources into the RG.
```
bpkroth, line 49: Again with the numbering issues.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.sh
```diff
@@ -0,0 +1,148 @@
...
+# Argument parsing
+while [[ "$#" -gt 0 ]]; do
+case $1 in
+    # Main control plane ARM template params
+    --controlPlaneArmParameters)
+        controlPlaneArmParameters="$2"
+        shift 2
+        ;;
+    # Results DB ARM template params
+    --resultsDbArmParameters)
```
bpkroth, line 18:
```suggestion
    --resultsDbArmParametersFile)
```

See also comments from README

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/results-db/mysql-template.json
```diff
@@ -0,0 +1,94 @@
...
+        },
+        "properties": {
+            "administratorLogin": "[parameters('dbAdminUsername')]",
+            "administratorLoginPassword": "[parameters('dbAdminPassword')]",
+            "storage": {
+                "storageSizeGB": 20,
+                "iops": 360,
+                "autoGrow": "Enabled",
+                "autoIoScaling": "Disabled",
+                "logOnDisk": "Disabled"
```
bpkroth, line 44: Can we leave those two settings unspecified to let whatever upstream defaults are take effect?
That way it's less for us to maintain and users hopefully get whateever enhancements happen on the upstream service for free.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/results-db/mysql-template.parameters.example.json
```diff
@@ -0,0 +1,24 @@
...
+            "value": "mlos"
+        },
+        "dbAdminPassword": {
+            "value": "PLACEHOLDER"
+        },
+        "dbFirewallRules": {
+            "value": [
+                {
+                    "name": "PLACEHOLDER",
+                    "startIpAddress": "123.123.0.0",
```
bpkroth, line 18: A better set of non-IPs to use would be something like the RFC1918 (private) addresses:

192.168.0.0
192.168.255.255

That way if they forget to change those they aren't accidentally opening things up to the public internet.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/rg-template.json
```diff
@@ -0,0 +1,346 @@
...
+        },
+        "vmSKU": {
+            "type": "string",
+            "defaultValue": "Standard_D2s_v3",
+            "metadata": {
+                "description": "VM SKU"
+            }
+        },
+        "vmAdminUsername": {
+            "type": "string"
```
bpkroth, line 20: does it make sense to give a default value here?

e.g., `mlos-admin`?

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/rg-template.json
```diff
@@ -0,0 +1,346 @@
...
+                "Standard_ZRS"
+            ]
+        }
+    },
+    "functions": [],
+    "variables": {
+        "storageAccountName": "[toLower(concat(replace(parameters('projectPrefix'), '-', '')))]",
+        "vmName": "[concat(parameters('projectPrefix'), '-vm')]",
+        "networkInterfaceName": "[concat(variables('vmName'), '-ni-', uniqueString(resourceGroup().id))]",
+        "vnetName": "[concat(variables('vmName'), '-vnet')]",
```
bpkroth, line 66: Let's name the vnet something less tied to just the single VM name.
That way we can add more VMs/services (additional schedulers, frontends, storage, etc.) to it in the future if necessary.

bpkroth: > This will require fixups throughout the rest of this config.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/rg-template.json
```diff
@@ -0,0 +1,346 @@
...
+            "[resourceId('Microsoft.Network/virtualNetworks', variables('vnetName'))]"
+        ],
+        "tags": {
+            "displayName": "[variables('kvName')]"
+        },
+        "properties": {
+            "enabledForDeployment": true,
+            "enabledForTemplateDeployment": true,
+            "enabledForDiskEncryption": true,
+            "tenantId": "72f988bf-86f1-41af-91ab-2d7cd011db47",
```
bpkroth, line 199: Please don't encode tenantIds.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/rg-template.json
```diff
@@ -0,0 +1,346 @@
...
+        }
+    },
+    "functions": [],
+    "variables": {
+        "storageAccountName": "[toLower(concat(replace(parameters('projectPrefix'), '-', '')))]",
+        "vmName": "[concat(parameters('projectPrefix'), '-vm')]",
+        "networkInterfaceName": "[concat(variables('vmName'), '-ni-', uniqueString(resourceGroup().id))]",
+        "vnetName": "[concat(variables('vmName'), '-vnet')]",
+        "publicIpName": "[concat(variables('vmName'), '-ip')]",
+        "nsgName": "[concat(variables('vmName'), '-nsg')]",
```
bpkroth, line 68:
```suggestion
        "nsgName": "[concat(variables('projectPrefix'), '-nsg')]",
```

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/rg-template.json
```diff
@@ -0,0 +1,346 @@
...
+                        ]
+                    }
+                }
+            },
+            "storageProfile": {
+                "imageReference": {
+                    "publisher": "Canonical",
+                    "offer": "0001-com-ubuntu-server-jammy",
+                    "sku": "22_04-lts-gen2",
+                    "version": "latest"
```
bpkroth, line 255: Any way to remove the version from this entirely and just let it take "Ubuntu Server LTS Latest"?

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/rg-template.json
```diff
@@ -0,0 +1,346 @@
...
+                "[resourceGroup().location]"
+            ]
+        },
+        "fileShareName": {
+            "type": "string",
+            "defaultValue": "mlos-file-share"
+        },
+        "vmLocation": {
+            "type": "string",
+            "defaultValue": "[resourceGroup().location]"
```
bpkroth, line 37: Will this create a VM in each location that we need storage as well?

bpkroth: Probably only need the scheduler VM in a single location, but we will need storage in each location that we want to run experiments.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.ps1
```diff
@@ -0,0 +1,115 @@
+
+param(
+    # Main control plane ARM template params
+    [Parameter(Mandatory=$True)]
+    [string] $controlPlaneArmParameters,
```
bpkroth, line 5:
```suggestion
    [string] $controlPlaneArmParametersFile,
```

Per comments elsewhere

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.ps1
```diff
@@ -0,0 +1,115 @@
...
+
+# Provision resources into the resource group with ARM template
+Write-Output "Provisioning control plane resources..."
+$deploymentResults = az deployment group create `
+    --resource-group $resourceGroupName `
+    --template-file .\rg-template.json `
+    --parameters $controlPlaneArmParameters `
+    | ConvertFrom-JSON
+
+if (!$?) {
```
bpkroth, line 28: Is that sufficient, or do you also need to check `$LASTEXITCODE`?

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.ps1
```diff
@@ -0,0 +1,115 @@
...
+}
+
+# Conditional provisioning of results DB
+if ($resultsDbArmParameters) {
+    Write-Output "Provisioning results DB..."
+    $dbDeploymentResults = az deployment group create `
+        --resource-group $resourceGroupName `
+        --template-file ./results-db/mysql-template.json `
+        --parameters $resultsDbArmParameters `
+        | ConvertFrom-JSON
```
bpkroth, line 40:
```suggestion
        -o json `
        | ConvertFrom-JSON
```

per comments above

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.ps1
```diff
@@ -0,0 +1,115 @@
...
+    --scope $kvId
+
+# Check if cert of same name exists in keyvault already
+$certThumbprint = az keyvault certificate show `
+    --name $certName `
+    --vault-name $kvName `
+    --query "x509ThumbprintHex" --output tsv `
+    2>$null
+
+if (!$?) {
```
bpkroth, line 78: Makes a little more sense to me to check that the output of the command that you depend on (e.g., `$certThumbprint`) looks sane (I guess just non-empty in this case) instead of the exit code of the last command, but maybe you need both?

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.sh
```diff
@@ -0,0 +1,148 @@
...
+
+# Provision resources into the resource group with ARM template
+echo "Provisioning control plane resources..."
+deploymentResults=$(az deployment group create \
+    --resource-group "$resourceGroupName" \
+    --template-file ./rg-template.json \
+    --parameters "$controlPlaneArmParameters" \
+    )
+
+if [[ ! $? ]]; then
```
bpkroth, line 58:
```suggestion
if [[ $? -ne 0 ]]; then
```

bpkroth: `! $?` in bash is going to evaluate to "not 1 (true)` which is the opposite of what you want here.  Better to be explicit in this case.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.sh
```diff
@@ -0,0 +1,148 @@
...
+        vmIpAddress=$(echo "$deploymentResults" | jq -r ".properties.outputs.vmIpAddress.value")
+
+        # VM IP access for results DB
+        az mysql flexible-server firewall-rule update \
+            --resource-group "$resourceGroupName" \
+            --name "$dbName" \
+            --rule-name "AllowVM-$vmName" \
+            --start-ip-address "$vmIpAddress" \
+            --end-ip-address "$vmIpAddress" \
+
```
bpkroth, line 86:
```suggestion
```

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.sh
```diff
@@ -0,0 +1,148 @@
...
+        vmName=$(echo "$deploymentResults" | jq -r ".properties.outputs.vmName.value")
+        vmIpAddress=$(echo "$deploymentResults" | jq -r ".properties.outputs.vmIpAddress.value")
+
+        # VM IP access for results DB
+        az mysql flexible-server firewall-rule update \
+            --resource-group "$resourceGroupName" \
+            --name "$dbName" \
+            --rule-name "AllowVM-$vmName" \
+            --start-ip-address "$vmIpAddress" \
+            --end-ip-address "$vmIpAddress" \
```
bpkroth, line 85:
```suggestion
            --end-ip-address "$vmIpAddress"
```

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.sh
```diff
@@ -0,0 +1,148 @@
...
+currentUserAlias=$(az account show --query "user.name" --output tsv)
+resourceGroupId=$(az group show --name "$resourceGroupName" --query "id" --output tsv)
+
+# Assign 'Key Vault Administrator' access to current user
+kvName=$(echo "$deploymentResults" | jq -r ".properties.outputs.kvName.value")
+kvId=$(az keyvault show --name "$kvName" --resource-group "$resourceGroupName" --query "id" --output tsv)
+az role assignment create \
+    --assignee "$currentUserAlias" \
+    --role "Key Vault Administrator" \
+    --scope "$kvId" \
```
bpkroth, line 99:
```suggestion
    --scope "$kvId"
```

Please remove these trailing `\` characters from the end of a multi-line command - it will cause weird problems.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.sh
```diff
@@ -0,0 +1,148 @@
...
+
+# Check if cert of same name exists in keyvault already
+certThumbprint=$(az keyvault certificate show \
+    --name "$certName" \
+    --vault-name "$kvName" \
+    --query "x509ThumbprintHex" --output tsv \
+    2> /dev/null \
+    || echo "NOCERT")
```
bpkroth, line 107:
```suggestion
    || echo "NOCERT" \
    )
```
This is a complete nit, but you use both bare closing parens `)` on their own line, and closing `)` on the line with the end of the command.  Let's just be consistent about which one we use within a file.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/rg-template.json
eujing: The VM will not follow the storage locations, only a single VM in this location as specified.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/rg-template.json
```diff
@@ -0,0 +1,346 @@
...
+                "Standard_LRS",
+                "Standard_RAGRS",
+                "Standard_RAGZRS",
+                "Standard_ZRS"
+            ]
+        }
+    },
+    "functions": [],
+    "variables": {
+        "storageAccountName": "[toLower(concat(replace(parameters('projectPrefix'), '-', '')))]",
```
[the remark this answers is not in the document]
eujing: Ah good catch! Actually I meant to totally remove the concat because it used to add a `storage` suffix, but that was going over the character limit after adding location suffix so I removed the suffix but forgot to remove the concat. Will remove!

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/README.md
eujing: Added sh example!

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/rg-template.json
```diff
@@ -0,0 +1,346 @@
...
+        ],
+        "tags": {
+            "displayName": "[variables('vmName')]"
+        },
+        "properties": {
+            "hardwareProfile": {
+                "vmSize": "[parameters('vmSKU')]"
+            },
+            "osProfile": {
+                "computerName": "[variables('vmName')]",
```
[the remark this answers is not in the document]
eujing: Added `vmCustomData` as a ARM parameter, defaulting `null` when empty to address the issue of empty string cloud-init and repeated deployments as discussed!

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/README.md
eujing: Changed back to actual numbers

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/README.md
```diff
@@ -0,0 +1,68 @@
...
+    ```shell
+    ./setup-rg.ps1 `
+        -controlPlaneArmParameters $controlPlaneArmParams `
+        -resultsDbArmParameters $resultsDbArmParams  # If provisioning results DB, otherwise omit `
+        -servicePrincipalName $servicePrincipalName `
+        -resourceGroupName $resourceGroupName `
+        -certName $certName
+    ```
+
+    where `$*ArmParams` can be the corresponding `*.parameters.json` and from before. However, it also follows the same usage as `--parameters` in [az deployment group create](https://learn.microsoft.com/en-us/cli/azure/deployment/group?view=azure-cli-latest#az-deployment-group-create-examples).
```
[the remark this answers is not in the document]
eujing: Changed the naming as suggested!

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/results-db/mysql-template.json
eujing: Removed these settings to be unspecified.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/results-db/mysql-template.parameters.example.json
eujing: That makes sense! Changed them to your suggestion.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/rg-template.json
eujing: Agreed, changed these to the project prefix instead.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/rg-template.json
eujing: Good catch this was from a generated template, changed it to an ARM script to pull the `tenantId` off the subscription.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.ps1
```diff
@@ -0,0 +1,115 @@
...
+    [int] $certExpirationYears = 1
+)
+
+# Provision resources into the resource group with ARM template
+Write-Output "Provisioning control plane resources..."
+$deploymentResults = az deployment group create `
+    --resource-group $resourceGroupName `
+    --template-file .\rg-template.json `
+    --parameters $controlPlaneArmParameters `
+    | ConvertFrom-JSON
```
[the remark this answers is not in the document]
eujing: Good point! Added the explicit output for all of these statements expecting JSON, across both the powershell and bash scripts.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.ps1
eujing: Agreed, modified this to be similar to the bash version where the value becomes `NOCERT` on failure of the keyvault command.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.ps1
eujing: I feel this is sufficient, since `$?` is always a boolean indicating success of the last command instead of explicit exit codes.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/rg-template.json
eujing: There doesn't seem to be a simple way to do this without resorting to external querying like:
```sh
> az vm image list --all --publisher Canonical \
  | jq '[.[] | select(.sku | endswith("lts-gen2"))] | max_by(.version)'

{
  "architecture": "x64",
  "offer": "0001-com-ubuntu-server-jammy-daily",
  "publisher": "Canonical",
  "sku": "22_04-daily-lts-gen2",
  "urn": "Canonical:0001-com-ubuntu-server-jammy-daily:22_04-daily-lts-gen2:22.04.202308280",
  "version": "22.04.202308280"
}
```
which would complicate the script if trying to use as a ARM parameter. What do you think about this sort of querying?

bpkroth: LGTM, thanks!

bpkroth: Tho, let's just leave a note about this in the comments for now and leave it for a future PR.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.ps1
```diff
@@ -0,0 +1,118 @@
+
```
bpkroth, line 1: I think you might need to run `make licenseheaders` to apply the standard license header to this file before it will pass the CI checks.

bpkroth (approved):
Looks reasonable to me.  Let's fix the remaining CI issues and get it in.  We can fix additional bugs/improvements in future PRs.

bpkroth:
Have a look here for excluding your new arm template json configs from schema checking.

https://github.com/microsoft/MLOS/blob/6997c2ab8ad00dc14db1b66c363f0316567a93e3/mlos_bench/mlos_bench/tests/config/services/test_load_service_config_examples.py#L28-L33

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.ps1
bpkroth: I've run into trouble with that in the past, but it might have been due to behavior like this:

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_automatic_variables?view=powershell-7.3#section-1

> Until PowerShell 7, wrapping a statement within parentheses `(...)`, subexpression syntax `$(...)`, or an array expression `@(...)` always reset `$?` to `True`. For example, `(Write-Error)` shows `$?` as `True`. This behavior changed in PowerShell 7, so that `$?` always reflects the actual success of the last command run in these expressions.

eujing: Good to know! Will add a note about this in the README about using this script recommending Powershell 7

eujing:
> Have a look here for excluding your new arm template json configs from schema checking.
> 
> https://github.com/microsoft/MLOS/blob/6997c2ab8ad00dc14db1b66c363f0316567a93e3/mlos_bench/mlos_bench/tests/config/services/test_load_service_config_examples.py#L28-L33

Got it! I had to change the filtering away from in-place removing as I was getting some issues with removed paths still appearing in the list during iteration.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.ps1
eujing: The `licenseheaders` tool does not seem to recognize `.ps1` files despite us including it as an extension, but I manually added the license header in

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/rg-template.json
eujing: Added a TODO with this!

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.ps1
bpkroth: You are correct.  I guess we'd just been doing that manually.

I've made a PR to fix that upstream:
https://github.com/johann-petrak/licenseheaders/pull/78

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.ps1
bpkroth: Could add a `#Requires -Version 7` header to the top.

eujing: Added it! Didn't realize that was a thing, thanks for letting me know!

PR #481, opened 2023-08-10
Implement unit tests for `LocalEnv`

motus:
verify that `read_telemetry_data` and `read_results_data` logic works.

Also:
* Fix a bug in handling the dates in `read_telemetry_data`
* Fix a pylint issue in `launcher_test`

**TODO:**
* [x] Fix a bug in `local_exec` service on Windows


mlos_bench/mlos_bench/tests/environments/local_env_test.py
```diff
@@ -0,0 +1,68 @@
...
+                f"echo '\"{time_str1}\",mem_usage,10240' >> telemetry.csv",
+                f"echo '\"{time_str2}\",cpu_load,0.8' >> telemetry.csv",
+                f"echo '\"{time_str2}\",mem_usage,20480' >> telemetry.csv",
+            ],
+            "read_results_file": "output.csv",
+            "read_telemetry_file": "telemetry.csv",
+        },
+        tunables=tunable_groups,
+        service=LocalExecService(parent=ConfigPersistenceService()),
+    )
```
bpkroth, line 47: Regarding the failures on Windows, I'd made a comment about that here:
https://github.com/microsoft/MLOS/pull/461#discussion_r1277930027

bpkroth: Basically, the `echo 'metric,value' > output.csv` in Windows results in `metric,value ` (trailing space), and `pd.read_csv()` gets a column name of `value ` instead of `value`.

The fix I'd done there originally, was to special case that test and rename the columns to strip the trailing spaces on Windows like so:

```python
if sys.platform == 'win32':
    data.rename(str.rstrip, axis='columns', inplace=True)
```

I think I'd proposed doing that directly in the `local_exec` Env, but we decided to wait on it.

Could do that now and then remove the special case in that other test.

Thoughts?

mlos_bench/mlos_bench/tests/environments/local_env_test.py
```diff
@@ -0,0 +1,68 @@
...
+                "echo 'metric,value' > output.csv",
+                "echo 'latency,10' >> output.csv",
+                "echo 'throughput,66' >> output.csv",
+                "echo 'score,0.9' >> output.csv",
+                "echo '-------------------'",  # This output does not go anywhere
+                "echo 'timestamp,metric,value' > telemetry.csv",
+                f"echo '\"{time_str1}\",cpu_load,0.65' >> telemetry.csv",
+                f"echo '\"{time_str1}\",mem_usage,10240' >> telemetry.csv",
+                f"echo '\"{time_str2}\",cpu_load,0.8' >> telemetry.csv",
+                f"echo '\"{time_str2}\",mem_usage,20480' >> telemetry.csv",
```
bpkroth, line 40: Also, this is probably fine for now, but I suspect in the future we might want to separate `run` and `status` as separate commands so we can have a separate polling script and not force the client to shove all of that into a single `run` script that would then require them to implement the asynchronous polling.

mlos_bench/mlos_bench/tests/environments/local_env_test.py
```diff
@@ -0,0 +1,68 @@
...
+        (status, data) = env_context.run()
+        assert status.is_succeeded()
+        assert data == {
+            "latency": 10.0,
+            "throughput": 66.0,
+            "score": 0.9,
+        }
+
+        (status, telemetry) = env_context.status()
+        assert status.is_good()
```
bpkroth, line 62: For a separate test, seems like a good place to test the `CompositeEnv` rollup logic of status/metrics as well, especially when errors are encountered in an intermediate child Env.

mlos_bench/mlos_bench/tests/environments/local_env_test.py
motus: @bpkroth yes, I guess we have to strip the spaces when we read the data. I'll add your code to the PR.

That leads us to another problem: the lines
```python
"echo 'timestamp,metric,value' > telemetry.csv",
f"echo '\"{time_str1}\",cpu_load,0.65' >> telemetry.csv",
f"echo '\"{time_str1}\",mem_usage,10240' >> telemetry.csv",
f"echo '\"{time_str2}\",cpu_load,0.8' >> telemetry.csv",
f"echo '\"{time_str2}\",mem_usage,20480' >> telemetry.csv",
```
currently produce the following string:
```
timestamp,metric,value 
"\"2023-08-10 20:18:10\",cpu_load,0.65" 
"\"2023-08-10 20:18:10\",mem_usage,10240" 
"\"2023-08-10 20:19:10\",cpu_load,0.8" 
"\"2023-08-10 20:19:10\",mem_usage,20480" 
```
(with trailing spaces, of course)
do you have any idea how to fix that?
note that the timestamp can be like `2023-08-07 14:59:12,593000` - i.e., with a comma inside

mlos_bench/mlos_bench/tests/environments/local_env_test.py
motus: absolutely!

mlos_bench/mlos_bench/tests/environments/local_env_test.py
motus: yes, this is a big deal, and we need to test the status and data aggregation thoroughly. I plan to write a bunch of unit tests for it when I start working on the environment reset logic - this will be a central part of it

mlos_bench/mlos_bench/tests/environments/local_env_test.py
bpkroth: I think you can just drop the quotes.

bpkroth: Oh wait, the timestamp includes a comma?

motus: and
```python
"echo 'timestamp,metric,value' > telemetry.csv",
f"echo '{time_str1},cpu_load,0.65' >> telemetry.csv",
f"echo '{time_str1},mem_usage,10240' >> telemetry.csv",
f"echo '{time_str2},cpu_load,0.8' >> telemetry.csv",
f"echo '{time_str2},mem_usage,20480' >> telemetry.csv",
```
gives
```csv
timestamp,metric,value 
"2023-08-10 20:28:37,cpu_load,0.65" 
"2023-08-10 20:28:37,mem_usage,10240" 
"2023-08-10 20:29:37,cpu_load,0.8" 
"2023-08-10 20:29:37,mem_usage,20480" 
```

motus: @bpkroth I agree, all those Windows shenanigans are probably not relevant to the real applications. Still, I just want to come up with a simple and human-readable unit test that works in both Linux and Windows

bpkroth: Try this:

```
"echo timestamp,metric,value > telemetry.csv",
f"echo {time_str1},cpu_load,0.65 >> telemetry.csv",
f"echo {time_str1},mem_usage,10240 >> telemetry.csv",
f"echo {time_str2},cpu_load,0.8 >> telemetry.csv",
f"echo {time_str2},mem_usage,20480 >> telemetry.csv",
```

(no quotes)

Works locally at least, though I haven't tested it in the Python stack yet.

bpkroth (approved):

mlos_bench/mlos_bench/tests/environments/local_env_test.py
motus: wee! it works! Thank you!

mlos_bench/mlos_bench/environments/local/local_env.py
```diff
@@ -182,6 +183,9 @@ def run(self) -> Tuple[Status, Optional[dict]]:
             self._config_loader_service.resolve_path(
                 self._read_results_file, extra_paths=[self._temp_dir]))
 
+        if sys.platform == 'win32':
+            data.rename(str.rstrip, axis='columns', inplace=True)
```
bpkroth, line 187: @motus could you go back and add a comment about why we have to do this in #483

PR #483, opened 2023-08-11
Unit test to read experimental data in wide format

### LINKED ISSUE #462: mlos_bench: read results file format tests
@poojanilangekar, could you please add some tests for the changes in #460 so that we make sure that the code accepts both input file formats we expect to be able to support?  Thanks!

motus:
Make sure we can read results in both tall and wide formats. Together with (already merged) PR #481, this closes issue #462.

Closes #462


mlos_bench/mlos_bench/tests/environments/local_env_test.py
```diff
@@ -66,3 +66,30 @@ def test_local_env(tunable_groups: TunableGroups) -> None:
...
+
+def test_local_env_wide(tunable_groups: TunableGroups) -> None:
+    """
+    Produce benchmark data in wide format and read it.
+    """
+    local_env = LocalEnv(
+        name="Test Local Env",
+        config={
+            "run": [
+                "echo 'latency,throughput,score' > output.csv",
```
bpkroth, line 79: Can we add one more that doesn't include the header?
And also a negative test that has the wrong number of columns or wrong format of the columns?

mlos_bench/mlos_bench/environments/local/local_env.py
```diff
@@ -181,20 +181,25 @@ def run(self) -> Tuple[Status, Optional[dict]]:
 
         data: pandas.DataFrame = pandas.read_csv(
             self._config_loader_service.resolve_path(
-                self._read_results_file, extra_paths=[self._temp_dir]))
+                self._read_results_file, extra_paths=[self._temp_dir]),
+            index_col=False,
+        )
 
         if sys.platform == 'win32':
```
bpkroth, line 188: I left a comment in #481 but I think it was after it got merged.
Can you please add a quick comment about why we need this special case for win32 here (I know it's not related to this PR).

bpkroth:
```suggestion
        # Windows cmd interpretation of > redirect symbols can leave trailing spaces in
        # the final column, which leads to misnamed columns.
        # For now, we simply strip trailing spaces from column names to account for that.
        if sys.platform == 'win32':
```

mlos_bench/mlos_bench/environments/local/local_env.py
```diff
@@ -204,25 +209,26 @@ def status(self) -> Tuple[Status, List[Tuple[datetime, str, Any]]]:
...
             # FIXME: We should not be assuming that the only output file type is a CSV.
-            data: pandas.DataFrame = pandas.read_csv(
-                self._config_loader_service.resolve_path(
-                    self._read_telemetry_file, extra_paths=[self._temp_dir]),
-                parse_dates=[0],
-            )
+            data: pandas.DataFrame = pandas.read_csv(fname, index_col=False, parse_dates=[0])
+
+            if sys.platform == 'win32':
+                data.rename(str.rstrip, axis='columns', inplace=True)
```
bpkroth, line 219: Maybe turn this into a small stub function (e.g., `normalize_column_names`) that can be called.
Then we can consolidate the comment about it too.

mlos_bench/mlos_bench/environments/local/local_env.py
```diff
@@ -204,25 +209,26 @@ def status(self) -> Tuple[Status, List[Tuple[datetime, str, Any]]]:
...
+
+            col_names = ["timestamp", "metric", "value"]
```
bpkroth, line 221:
```suggestion
            expected_col_names = ["timestamp", "metric", "value"]
```

mlos_bench/mlos_bench/environments/local/local_env.py
```diff
@@ -204,25 +209,26 @@ def status(self) -> Tuple[Status, List[Tuple[datetime, str, Any]]]:
...
+            if len(data.columns) != len(col_names):
+                raise ValueError('Telemetry data must have columns {col_names}')
+            elif list(data.columns) != col_names:
+                # Assume no header - this is ok for telemetry data.
```
bpkroth, line 225: How will this respond if there are headers, but with different names?

mlos_bench/mlos_bench/tests/environments/local_env_test.py
```diff
@@ -66,3 +68,159 @@ def test_local_env(tunable_groups: TunableGroups) -> None:
...
+    """
+    local_env = LocalEnv(
+        name="Test Local Env",
+        config={
+            "run": [
+                # Error: field 1 must be a timestamp
+                "echo 1,cpu_load,0.65 > telemetry.csv",
+                "echo 2,mem_usage,10240 >> telemetry.csv",
+                "echo 3,cpu_load,0.8 >> telemetry.csv",
+                "echo 4,mem_usage,20480 >> telemetry.csv",
```
bpkroth, line 162: Couldn't these be listed as unix epoch based timestamps (i.e., just ints)?

mlos_bench/mlos_bench/environments/local/local_env.py
motus: it should replace the names with those in `col_names` - but yes, it's a good idea to add a unit test for it

mlos_bench/mlos_bench/tests/environments/local_env_test.py
motus: pandas does not recognize an epoch-based int as a timestamp - and I don't want to overengineer. Let's just assume that if pandas cannot automatically infer a datetime, it's an error (btw pandas actually does a really good job recognizing all possible timestamp formats)

mlos_bench/mlos_bench/environments/local/local_env.py
motus: I like that

mlos_bench/mlos_bench/tests/environments/local_env_test.py
bpkroth: benchbase outputs unix timestamps with microsecond precision, fwiw

bpkroth (approved):

PR #484, opened 2023-08-12
Unit tests for `shell_env_vars` logic

### LINKED ISSUE #386: FIxup bugs from #383 - Add end to end environment variable test script

motus:
Summary of changes:
* Add unit tests for `LocalEnv` to make sure that we pass environment variables into a local script correctly.
* Split unit tests for `LocalEnv` into three separate files for maintainability.

P.S. Turned out, we pass the environment variables into a local script correctly - we just have to use the cmd syntax `%VAR%` on Windows instead of `$VAR`. We now have unit tests for both cases.

-----------
> It looks like `local_exec` service fails to set the environment variables on Windows.
> This PR adds a unit test to reproduce the problem.
> We need to debug the issue and push the fix into this or other PR.
> 
> P.S. `local_env_test.py` has too many tests now, so in this PR we split it into three separate files that have tests for the results, telemetry, and environment variables, respectively.

~~**NOTE:** Merge after #483~~

Closes #386


bpkroth:
> It looks like `local_exec` service fails to set the environment variables on Windows. We need to debug the issue.
> 
> **NOTE:** Merge after #483
> 
> Closes #386

I was actually seeing cases where we don't propagate environment variables to Linux shell scripts either.
I'll cook a smaller example up and we can address that in a different PR though.
Thanks for adding the tests!

mlos_bench/mlos_bench/tests/environments/local_env_test.py
```diff
@@ -66,3 +68,132 @@ def test_local_env(tunable_groups: TunableGroups) -> None:
...
+                "other_arg": 222,  # NOT passed into "shell_env_params"
+            },
+            "tunable_params": ["kernel"],
+            "shell_env_params": [
+                "const_arg",                # From "const_arg"
+                "kernel_sched_latency_ns",  # From "tunable_params"
+            ],
+            "run": [
+                "echo const_arg,other_arg,unknown_arg,kernel_sched_latency_ns > output.csv",
+                "echo $const_arg,$other_arg,$unknown_arg,$kernel_sched_latency_ns >> output.csv",
```
bpkroth, line 183: Would be good to also test this in a composite nested env.
That's where I've run into issues previously.

motus: yes, I plan to add many more tests for the composite environment

bpkroth: I assume that's for a future PR then?

bpkroth (approved):

mlos_bench/mlos_bench/tests/environments/local_env_test.py
motus: yes

PR #485, opened 2023-08-14
Better type annotation for `Environment.run()` methods

motus:
Nothing but a single change in type annotation of `Environment.run()`:
`dict` replaced with `Dict[str, float]`
That makes mypy happier in other parts of the code


bpkroth (approved):
I think this is ok.  Part of me has a nagging thought about places where we might use a RemoteEnv to `run` something that isn't a benchmark, but maybe that belongs in `setup`, `teardown` instead.

PR #489, opened 2023-08-16
Switch back to FLAML as the default optimizer

bpkroth:
We've found that SMAC has some non-determinism and incomplete integration, which causes some issues with our pipelines and experiment results, so for now we switch back to FLAML as the default.


motus (approved):
FLAML FTW! 😺

bpkroth:
> FLAML FTW! 😺

I looked under the rock.  Both of them need additional integrations unfortunately ...

PR #490, opened 2023-08-16
Expand dollar variables in the globals

motus:
This feature greatly simplifies our configs for complex multi-VM setups.

P.S. I cannot come up with any simple and elegant unit tests for this feature. Technically, we can define some dollar variables in `global_test_config.jsonc` and check their values in the log in `launcher_test.py`, but we don't have that many parameters in `MockEnv` and `MockOptimizer` that we can expand. I'll think about it and do it in the next PR.

### DIFF
--- a/mlos_bench/mlos_bench/launcher.py
+++ b/mlos_bench/mlos_bench/launcher.py
@@ -12,6 +12,8 @@ command line.
 
 import logging
 import argparse
+
+from string import Template
 from typing import Any, Dict, Iterable, List, Optional, Tuple, Type
 
 from mlos_bench.config.schemas import ConfigSchema
@@ -84,6 +86,7 @@ class Launcher:
             args_rest,
             {key: val for (key, val) in config.items() if key not in vars(args)},
         )
+        self.global_config = self._expand_vars(self.global_config)
 
         env_path = args.environment or config.get("environment")
         if not env_path:
@@ -226,6 +229,20 @@ class Launcher:
         if config_path:
             global_config["config_path"] = config_path
         return global_config
+
+    def _expand_vars(self, value: Any) -> Any:
+        """
+        Expand dollar variables in the globals.
+
+        NOTE: `self.global_config` must be set.
+        """
+        if isinstance(value, str):
+            return Template(value).safe_substitute(self.global_config)
+        if isinstance(value, dict):
+            return {key: self._expand_vars(val) for (key, val) in value.items()}
+        if isinstance(value, list):
+            return [self._expand_vars(val) for val in value]
+        return value
 
     def _init_tunable_values(self, random_init: bool, seed: Optional[int],
                              args_tunables: Optional[str]) -> TunableGroups:
--- a/mlos_bench/mlos_bench/config/experiments/experiment_RedisBench.jsonc
+++ b/mlos_bench/mlos_bench/config/experiments/experiment_RedisBench.jsonc
@@ -2,8 +2,8 @@
     "experiment_id": "RedisBench",
     "trial_id": 1,
 
-    // TODO: Generate these names from $experiment_id
-    "deploymentName": "RedisBench",
+    // Any global parameter can be used as a dollar variable in the global config:
+    "deploymentName": "$experiment_id",
     "vmName": "os-autotune-linux-vm",
 
     "resourceGroup": "os-autotune",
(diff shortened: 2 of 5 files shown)


mlos_bench/mlos_bench/launcher.py
```diff
@@ -227,6 +230,20 @@ def _load_config(self,
             global_config["config_path"] = config_path
         return global_config
 
+    def _expand_vars(self, value: Any) -> Any:
+        """
+        Expand dollar variables in the globals.
```
bpkroth, line 235: First, thanks for this!
I had a thought about this during some local hacking I was doing along the same lines.
What do you think about doing this recursively until all `$variables` are expanded or can't be (in which case, throw an error)?
The idea is basically, to support things like this, where one value might also include the results of a template from another one:

- template vmServerName with `"vmServerName": "vmServer-$experiment-$location"`
- template vnetName with `"vnetName": "vnet-$vmServerName"`

mlos_bench/mlos_bench/launcher.py
```diff
@@ -227,6 +230,20 @@ def _load_config(self,
             global_config["config_path"] = config_path
         return global_config
 
+    def _expand_vars(self, value: Any) -> Any:
```
bpkroth, line 233: We do this in several places now.  Is it possible for us to consolidate this into `mlos_bench.util` and fixup all the places where we do `$var` expansion to call it there?  (see also comments about recursive calls below)

bpkroth:
> This feature greatly simplifies our configs for complex multi-VM setups.
> 
> P.S. I cannot come up with any simple and elegant unit tests for this feature. Technically, we can define some dollar variables in `global_test_config.jsonc` and check their values in the log in `launcher_test.py`, but we don't have that many parameters in `MockEnv` and `MockOptimizer` that we can expand. I'll think about it and do it in the next PR.

Could we do something like this for testing:
https://til.simonwillison.net/pytest/pytest-argparse

Basically, let's stop using `parse_args()` without an argument, and instead explicitly pass `sys.args` to via the `__init__` function and on to `_parse_args()`
That would allow us to write a test function on the `Launcher` and the data it returns from the args/files, by passing `['--foo', 'bar']` style explicit lists of args.

Make sense?

mlos_bench/mlos_bench/launcher.py
motus: yes, we process the parameters in the order they appear in config. E.g., in the following fragment
```python
{
    "experiment_id": "RedisBench",
    "location": "westus2",
    "vmServerName": "vmServer-$experiment_id-$location",
    "vnetName": "vnet-$vmServerName",
}
```
`vnetName` would expand to `vnet-vmServer-RedisBench-westus2`

mlos_bench/mlos_bench/launcher.py
motus: I might do it later. Right now, we use variable expansion in four different contexts and at different stages of the Environment life cycle:
* Global parameters: we do the expansion once at the startup time; both input and output are the same global parameters
* `LocalFileShareEnv`: we substitute `const_args` and tunables (plus the special variable `$PWD`) inside the file paths for upload and download
* `tunable_params_map`: we substitute the _lists_ of covariant groups' names inside the `tunable_params`
* `LocalEnv` and `RemoteEnv`: we pass `shell_env_params` as shell environment variables of the executable script

Note that all these cases use _different_ parameters for the expansions, and the expansion mechanisms are different, too (e.g., string substitution in globals vs. list expansion for the tunables). That makes it hard to have a single unified function for all use cases. I would argue that adding string substitution for the globals covers most of our needs for the flexible configs for the nearest future; later we can rethink the way we handle the dollar variables in our configs and come up with a more elegant solution

motus:
> let's stop using `parse_args()` without an argument, and instead explicitly pass `sys.args` to via the `__init__` function and on to `_parse_args()`

I love it! This way we can write much more granular unit tests for the Launcher. let's do it in a separate PR because there will be many more unit tests, not all related to the variable expansion

bpkroth (approved):

mlos_bench/mlos_bench/launcher.py
bpkroth: Ah, if the order manages it, then it should be ok I think.  Thanks!  That's how arm templates work too I believe.

bpkroth: Fixed in #493

PR #491, opened 2023-08-17
Support customData file to pass to ARM templates

bpkroth:
Adds support to allow reading `customData` out of a file to pass to the ARM templates.
This way we can code `cloud-init` files (for example) separately with all the usual syntax help, rather than embed them in JSON configs.


motus (approved):
Looks good!

PR #493, opened 2023-08-18
CLI args testing and improvements

bpkroth:
- Support multiple instances of `--globals` and other arguments to build a list.
- Provide some common mistype aliases for CLI args.
- [x] Adds some tests (additional ones to be added later)
  See Also testing comments in #490


mlos_bench/mlos_bench/launcher.py
```diff
@@ -138,7 +139,7 @@ def _parse_args(parser: argparse.ArgumentParser) -> Tuple[argparse.Namespace, Li
                  ' Set to DEBUG for debug, WARNING for warnings only.')
 
         parser.add_argument(
-            '--config_path', nargs="+", required=False,
+            '--config_path', nargs="+", action='append', required=False,
```
motus, line 142:
```suggestion
            '--config_path', nargs="+", action='extend', required=False,
```

mlos_bench/mlos_bench/launcher.py
```diff
@@ -165,13 +166,13 @@ def _parse_args(parser: argparse.ArgumentParser) -> Tuple[argparse.Namespace, Li
             help='Seed to use with --random_init')
 
         parser.add_argument(
-            '--tunable_values', nargs="+", required=False,
+            '--tunable_values', nargs="+", action='append', required=False,
```
motus, line 169:
```suggestion
            '--tunable_values', nargs="+", action='extend', required=False,
```

mlos_bench/mlos_bench/launcher.py
```diff
@@ -165,13 +166,13 @@ def _parse_args(parser: argparse.ArgumentParser) -> Tuple[argparse.Namespace, Li
...
             help='Path to one or more JSON files that contain values of the tunable' +
                  ' parameters. This can be used for a single trial (when no --optimizer' +
                  ' is specified) or as default values for the first run in optimization.')
 
         parser.add_argument(
-            '--globals', nargs="+", required=False,
+            '--globals', nargs="+", action='append', required=False,
```
motus, line 175:
```suggestion
            '--globals', nargs="+", action='extend', required=False,
```

mlos_bench/mlos_bench/launcher.py
```diff
@@ -180,7 +181,11 @@ def _parse_args(parser: argparse.ArgumentParser) -> Tuple[argparse.Namespace, Li
             dest='teardown', action='store_false',
             help='Disable teardown of the environment after the benchmark.')
 
-        return parser.parse_known_args()
+        (args, args_rest) = parser.parse_known_args()
+        args.config_path = list(chain(*args.config_path))
+        args.tunable_values = list(chain(*args.tunable_values))
+        args.globals = list(chain(*args.globals))
+        return (args, args_rest)
```
motus, line 188:
```suggestion
        return parser.parse_known_args()
```
`action='extend'` should flatten the lists automatically

motus (changes requested):
I think if we use `action='extend'` then the result will be in the desired format automatically. Can you please check? Thank you!

mlos_bench/mlos_bench/launcher.py
```diff
@@ -129,16 +129,17 @@ def _parse_args(parser: argparse.ArgumentParser) -> Tuple[argparse.Namespace, Li
                  ' for additional config examples for this and other arguments.')
 
         parser.add_argument(
-            '--log_file', required=False,
+            '--log_file', '--log-file', required=False,
```
motus, line 132: I don't like mixing `-` and `_` in the option name, but the idea is to keep it consistent across the config parameters, internal argparse args names, and command-line options. Any config parameter can be a command line option, so after your change we'll end up with command lines like `--experiment_id foo --log-level DEBUG` and it would be hard to explain to the user why we can't have `--experiment-id` instead of `--experiment_id` but it's ok for `--log-level`

motus: so I would keep underscores everywhere for consistency, although I totally agree that this is not the most aesthetically pleasing solution

bpkroth: I kept mistyping them locally, so I added aliases.  I'd like to keep them if possible.  I don't think they should add too much confusion - it's just in the cli args.

`--experiment-id` should probably move to the `globals` tbh, or we could just special treat it and convert.

mlos_bench/mlos_bench/launcher.py
```diff
@@ -237,9 +251,14 @@ def _expand_vars(self, value: Any) -> Any:
         NOTE: `self.global_config` must be set.
         """
         if isinstance(value, str):
+            if "$" in value:
+                _LOG.warning("Expand vars: %s", value)
```
bpkroth, line 255: - [ ] Remove this.  Was debugging something.

mlos_bench/mlos_bench/launcher.py
```diff
@@ -290,7 +309,11 @@ def _load_storage(self, args_storage: Optional[str]) -> Storage:
             # pylint: disable=import-outside-toplevel
             from mlos_bench.storage.sql.storage import SqlStorage
             return SqlStorage(self.tunables, service=self._parent_service,
-                              config={"drivername": "sqlite", "database": ":memory:"})
+                              config={
+                                  "drivername": "sqlite",
+                                  "database": ":memory:",
+                                  "lazy_schema_create": True,
```
bpkroth, line 315: This was slowing down tests a little bit and is unnecessary often.

mlos_bench/mlos_bench/launcher.py
```diff
@@ -237,9 +251,14 @@ def _expand_vars(self, value: Any) -> Any:
...
         if isinstance(value, str):
+            if "$" in value:
+                _LOG.warning("Expand vars: %s", value)
             return Template(value).safe_substitute(self.global_config)
         if isinstance(value, dict):
-            return {key: self._expand_vars(val) for (key, val) in value.items()}
+            # Note: we use a loop instead of dict comprehension in order to
+            # allow secondary expansion of subsequent values immediately.
+            for (key, val) in value.items():
+                value[key] = self._expand_vars(val)
```
motus, line 261: that's
```suggestion
            value = {key: self._expand_vars(val) for (key, val) in value.items()}
```

motus: oh wait never mind, that won't work if there is a chain of dependencies

mlos_bench/mlos_bench/tests/config/__init__.py
```diff
@@ -28,8 +37,8 @@ def locate_config_examples(config_examples_dir: str) -> List[str]:
     """
     assert os.path.isdir(config_examples_dir)
     config_examples = []
-    for root, _, files in os.walk(config_examples_dir):
-        for file in files:
+    for root, _, _files in os.walk(config_examples_dir):
+        for file in _files:
             if file.endswith(".json") or file.endswith(".jsonc"):
                 config_examples.append(path_join(root, file))
```
motus, line 43:
```suggestion
    config_examples.extend(glob.glob("**/*.json*", root_dir=config_examples_dir, recursive=True))
```
would that work?

mlos_bench/mlos_bench/tests/config/globals/test_load_global_config_examples.py
```diff
@@ -0,0 +1,45 @@
...
+def filter_configs(configs_to_filter: List[str]) -> List[str]:
+    """If necessary, filter out json files that aren't for the module we're testing."""
+    return configs_to_filter
+
+
+configs = []
+# configs += filter_configs(locate_config_examples(path_join(ConfigPersistenceService.BUILTIN_CONFIG_PATH, CONFIG_TYPE)))
+configs += filter_configs(locate_config_examples(path_join(ConfigPersistenceService.BUILTIN_CONFIG_PATH, "experiments")))
+configs += filter_configs(locate_config_examples(path_join(BUILTIN_TEST_CONFIG_PATH, CONFIG_TYPE)))
+configs += filter_configs(locate_config_examples(path_join(BUILTIN_TEST_CONFIG_PATH, "experiments")))
```
motus, line 37:
```suggestion
configs = [
    # *filter_configs(locate_config_examples(path_join(ConfigPersistenceService.BUILTIN_CONFIG_PATH, CONFIG_TYPE))),
    *filter_configs(locate_config_examples(path_join(ConfigPersistenceService.BUILTIN_CONFIG_PATH, "experiments"))),
    *filter_configs(locate_config_examples(path_join(BUILTIN_TEST_CONFIG_PATH, CONFIG_TYPE))),
    *filter_configs(locate_config_examples(path_join(BUILTIN_TEST_CONFIG_PATH, "experiments"))),
]
```

mlos_bench/mlos_bench/launcher.py
bpkroth: Exactly, hence the change/comment.  I'm still not sure this is the best way, but it accomplishes the desired behavior we'd discussed in a previous PR as demonstrated in the tests we're adding here.

mlos_bench/mlos_bench/launcher.py
bpkroth: Part of the trouble as I see it is that we accept unknown arguments as globals, so if you accidentally use `--config-path` and we only support `--config_path`, then we get a global of `'config-path'` populated instead of setting the config paths or an error as one might expect.

mlos_bench/mlos_bench/tests/config/__init__.py
bpkroth: Probably?  I didn't really change this logic, just the variable names due to the `files` name conflict.
I can give it a shot though.

bpkroth: Nm.  `root_dir` option doesn't exist in older versions of python

motus (approved):
Looks good!

PR #494, opened 2023-08-23
FIO Environment config and scripts

motus:
First version of FIO environment for Linux. We can stack it together with other application-specific environments and let mlos_bench merge the results.


mlos_bench/mlos_bench/config/environments/apps/fio/fio.jsonc
```diff
@@ -0,0 +1,70 @@
...
+                        "mountPoint"
+                    ],
+                    "setup": [
+                        "sudo apt-get update",
+                        "sudo apt-get -y install fio"
+                    ],
+                    "run": [
+                        "mkdir -p /tmp/mlos_bench/fio/",
+                        "sudo mkdir -p $mountPoint/$experiment_id/$trial_id/output/fio/",
+                        "sudo fio --output-format=json $mountPoint/$experiment_id/$trial_id/scripts/latency-profile.fio > /tmp/mlos_bench/fio/latency-profile.out.json",
```
bpkroth, line 40: The location fio runs from is often important (e.g., on the VM's data or ephemeral disk instead of the OS disk).
Can we provide a parameter to help control for that?

mlos_bench/mlos_bench/config/environments/apps/fio/scripts/remote/latency-profile.fio
```diff
@@ -0,0 +1,21 @@
...
+
+[global]
+bs=4k
+rw=randrw
+random_generator=lfsr
+direct=1
+ioengine=libaio
+iodepth=128
+# Set max acceptable latency to 500msec
+latency_target=500000
```
bpkroth, line 14: That seems really high.  Maybe something lower like 50ms?

mlos_bench/mlos_bench/config/environments/apps/fio/scripts/remote/latency-profile.fio
```diff
@@ -0,0 +1,21 @@
...
+# profile over a 20s window
+latency_window=20000000
+# 99.0% of IOs must be below the target
+latency_percentile=99.0
+
+[device]
+filename=/dev/sda
```
bpkroth, line 21: This should probably be controllable.  One option would be to let it run in the CWD, and adjust that in the `run` script above.

mlos_bench/mlos_bench/config/environments/apps/fio/fio.jsonc
```diff
@@ -0,0 +1,70 @@
...
+                "config": {
+                    "required_args": [
+                        "vmName",
+                        "experiment_id",
+                        "trial_id",
+                        "mountPoint"
+                    ],
+                    "setup": [
+                        "sudo apt-get update",
+                        "sudo apt-get -y install fio"
```
bpkroth, line 35: This is fine for now, but one thing about needing to support Mariner in the future is we're going to have to special case which package manager to use in a lot of cases and that will probably mean converting these to more complicated scripts we download/run.

bpkroth:
> First version of FIO environment for Linux. We can stack it together with other application-specific environments and let mlos_bench merge the results.

What do you mean by "merge the results".  Is this basically an example to show how we can manage uploading multiple metrics from different sources?

mlos_bench/mlos_bench/config/environments/apps/fio/fio.jsonc
```diff
@@ -0,0 +1,69 @@
...
+                    ],
+                    "run": [
+                        "mkdir -p $mountPoint/$experiment_id/$trial_id/output/fio/",
+                        "sudo fio --output-format=json $mountPoint/$experiment_id/$trial_id/scripts/readwrite.fio.ini > /dev/null",  // warm-up
```
bpkroth, line 39: I thought there was a warmup option for fio configs?

bpkroth, line 39:
```suggestion
                        "cd ${fioDir:-$CWD} && sudo fio --output-format=json $mountPoint/$experiment_id/$trial_id/scripts/readwrite.fio.ini > /dev/null",  // warm-up
```

bpkroth: Then we could expose `fioDir` as an optional global to adjust the path.

mlos_bench/mlos_bench/config/environments/apps/fio/fio.jsonc
motus: done

mlos_bench/mlos_bench/config/environments/apps/fio/scripts/remote/latency-profile.fio
```diff
@@ -0,0 +1,21 @@
...
+bs=4k
+rw=randrw
+random_generator=lfsr
+direct=1
+ioengine=libaio
+iodepth=128
+# Set max acceptable latency to 500msec
+latency_target=500000
+# profile over a 20s window
+latency_window=20000000
```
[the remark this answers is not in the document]
motus: done

mlos_bench/mlos_bench/config/environments/apps/fio/scripts/remote/latency-profile.fio
motus: yep. also, I read the fio docs and this is probably not the best benchmark for our needs. I'll use a config that Microsoft recommends to benchmark our disks in the cloud, from here: [Azure/VMs/Benchmark a disk](https://learn.microsoft.com/en-us/azure/virtual-machines/disks-benchmarks#maximum-read-and-write-iops)

mlos_bench/mlos_bench/config/environments/apps/fio/fio.jsonc
motus: yep. this config is open source, so I think `apt-get` should be fine for the sample config

mlos_bench/mlos_bench/config/environments/apps/fio/fio.jsonc
motus: I can't find it in fio docs. 🤔
we can do it properly in the next PR when we figure it out

motus:
> What do you mean by "merge the results". Is this basically an example to show how we can manage uploading multiple metrics from different sources?

exactly. e.g., by stacking up `fio.jsonc` and `redis.jsonc` we automatically get metrics from both benchmarks at our root composite environment. we might not use FIO metrics directly for the optimization at this point, but they will be saved in the DB and can be used later as a side signal etc.

> Are we able to identify pre/post runs easily from this? That would be useful for normalization perhaps.

oh, that's a very good point! I'll add some option like `--prefix fio.before` to the script that parses FIO results so we can distinguish between the two

mlos_bench/mlos_bench/config/environments/apps/fio/scripts/remote/latency-profile.fio
motus: yes, I'll move it to the CLI level and expose the parameter in `const_args`

mlos_bench/mlos_bench/config/environments/apps/fio/fio.jsonc
```diff
@@ -0,0 +1,72 @@
...
+                "class": "mlos_bench.environments.remote.RemoteEnv",
+                "config": {
+                    "required_args": [
+                        "vmName",
+                        "experiment_id",
+                        "trial_id",
+                        "mountPoint"
+                    ],
+                    "const_args": {
+                        "fio_target": "/dev/sda"
```
bpkroth, line 34: Did you test this?
I'm pretty sure that will break whatever filesystem is on the disk.
Probably better to use an actual file on top of a filesystem.
Without that I think this is just going to directly write to the block device that the OS is on and seriously corrupt the filesystem.

bpkroth: Indeed this is a problem.  I know this because I tried it on a VM of mine and had to reinstall it 😢

motus: oops, thank you for double checking it. I was lucky to not experience this issue on my VM. I was under the impression that it reads and writes the same data (after all, `/dev/sda` comes from the official example). Anyway, I'll fix it in a second

motus: done

bpkroth: Yeah, fio is often used to test raw block device performance, so that example makes sense.
Whether or not you ran into it is probably a matter of how long the VM was up and whether or not you accessed the areas of the FS in a non-cached fashion.  I noticed it in `dmesg` output.

mlos_bench/mlos_bench/config/environments/apps/fio/fio.jsonc
```diff
@@ -0,0 +1,77 @@
...
+                    "const_args": {
+                        "fio_target": "/fio_test.tmp"
+                    },
+                    "setup": [
+                        "sudo apt-get update",
+                        "sudo apt-get -y install fio"
+                    ],
+                    "run": [
+                        "mkdir -p $mountPoint/$experiment_id/$trial_id/output/$fio_prefix/",
+                        "sudo fio --output-format=json --filename=$fio_target --runtime=20 $mountPoint/$experiment_id/$trial_id/scripts/readwrite.fio.ini > /dev/null",  // warm-up
```
bpkroth, line 48: Future version could also make the fio config file a variable and allow uploading it.
Let's make that a future PR though.

mlos_bench/mlos_bench/config/environments/apps/fio/fio.jsonc
```diff
@@ -0,0 +1,77 @@
...
+                    "download": [
+                        {
+                            "from": "$experiment_id/$trial_id/output/$fio_prefix",
+                            "to": "$PWD/$fio_prefix/"
+                        }
+                    ],
+                    "run": [
+                        "environments/apps/fio/scripts/local/process_fio_results.py --prefix $fio_prefix $fio_prefix/readwrite.json readwrite.csv"
+                    ],
+                    "read_results_file": "readwrite.csv"
```
bpkroth, line 72:
```suggestion
                    "read_results_file": "fio.results.csv"
```

And in that version, maybe something like this.

mlos_bench/mlos_bench/config/environments/apps/fio/fio.jsonc
bpkroth: Made #500 to track that.

mlos_bench/mlos_bench/config/environments/apps/fio/fio.jsonc
bpkroth: Made #500 to track that.

Issue #468, opened 2023-08-02
ARM template auto-shutdown and cleanup helper scripts


bpkroth:
Often times we will run with `teardown: false` and tweak experiments for different variations.  This means that after a while we may have stale resources running and eating resources in the cloud (Azure).

To fix that I propose:
1. [ ] Simply enable an auto-shutdown policy that somehow (*) detects idle resources and shutsdown the VM.
   That could help with compute costs, but not allocation costs.
   - [ ] (*) Not sure if the native version does this or if it's just a schedule, in which case we'd have to write something to run on the VM
   - [ ] Could also configure the ARM templates to do cascading deletion of certain resources, but probably not all (e.g., shared storage, shared vnets, shared db, etc.)
2. [ ] We need a script that could look at a shared database instance for all experiment definitions, and can determine which resources may no longer be necessary, and propose deletions.
   - [ ] We might need to add additional metadata to help track that.

PR #497, opened 2023-08-24
Change VM ops terminology to Host ops for cleaner Service type separation

bpkroth:
Reorg prep for SSH support.
Mostly renames to consolidate non-differences between VMs and pre-existing hosts.
SSH remote Service support to follow in subsequent PRs.


mlos_bench/mlos_bench/services/remote/azure/azure_services.py
```diff
@@ -81,6 +83,17 @@ class AzureVMService(Service, SupportsVMOps, SupportsRemoteExec):
...
+    # From: https://docs.microsoft.com/en-us/rest/api/compute/virtual-machines/deallocate
+    _URL_DEPROVISION = (
+        "https://management.azure.com" +
+        "/subscriptions/{subscription}" +
+        "/resourceGroups/{resource_group}" +
+        "/providers/Microsoft.Compute" +
+        "/virtualMachines/{vm_name}" +
+        "/delete" +
+        "?api-version=2022-03-01"
+    )
```
bpkroth, line 95: This is maybe something to move to a separate PR, but we'd previously been too loose with our terminology - VM deprovisioning actually deletes the VM, whereas deallocation releases the compute resources associated with that VM without actually deleting the VM object (e.g., it's disks).
See Also #468 for related commentary about cleaning up after ourselves for cost reduction reasons.

motus: yeah, let's do it in a separate PR. rn I am mostly concerned about keeping the framework usable after all our refactorings. I am fine with the changes as long as I can still run our existing configs at least for Redis and MySQL

bpkroth: I left a comment there referencing a new issue ( #498 ) and restored the original URL for now.

motus (approved):
Looks quite neat

PR #503, opened 2023-08-30
Extract benchmark results straight from the local or remote stdout

motus:
~~**TODO:**~~
* [x] Fix the JSON schema
* [x] Add unit tests

**Not** in this PR:
* Extract telemetry from stdout (will do in async mode later)

### DIFF
--- a/mlos_bench/mlos_bench/config/schemas/environments/common-environment-subschemas.json
+++ b/mlos_bench/mlos_bench/config/schemas/environments/common-environment-subschemas.json
@@ -80,7 +80,7 @@
                     "description": "The command lines to execute for the teardown phase.",
                     "$ref": "#/$defs/command_lines"
                 },
-                "parse_results_stdout": {
+                "results_stdout_pattern": {
                     "description": "A regex to parse the stdout of the run phase for results.",
                     "type": "string"
                 }
--- a/mlos_bench/mlos_bench/config/schemas/environments/common-environment-subschemas.json
+++ b/mlos_bench/mlos_bench/config/schemas/environments/common-environment-subschemas.json
@@ -79,6 +79,10 @@
                 "teardown": {
                     "description": "The command lines to execute for the teardown phase.",
                     "$ref": "#/$defs/command_lines"
+                },
+                "parse_results_stdout": {
+                    "description": "A regex to parse the stdout of the run phase for results.",
+                    "type": "string"
                 }
             }
         },
--- a/mlos_bench/mlos_bench/environments/script_env.py
+++ b/mlos_bench/mlos_bench/environments/script_env.py
@@ -100,7 +100,7 @@ class ScriptEnv(Environment, metaclass=abc.ABCMeta):
 
         return {key_sub: str(self._params[key]) for (key_sub, key) in rename.items()}
 
-    def _extract_stdout_results(self, stdout: str) -> Dict[str, str]:
+    def _extract_stdout_results(self, stdout: str) -> Dict[str, float]:
         """
         Extract the results from the stdout of the script.
 
@@ -111,10 +111,10 @@ class ScriptEnv(Environment, metaclass=abc.ABCMeta):
 
         Returns
         -------
-        results : Dict[str, str]
+        results : Dict[str, float]
             A dictionary of results extracted from the stdout.
         """
         if not self._parse_results_stdout:
             return {}
         _LOG.debug("Extract regex: '%s' from: '%s'", self._parse_results_stdout, stdout)
-        return dict(self._parse_results_stdout.findall(stdout))
+        return {key: float(val) for (key, val) in self._parse_results_stdout.findall(stdout)}
--- a/mlos_bench/mlos_bench/environments/script_env.py
+++ b/mlos_bench/mlos_bench/environments/script_env.py
@@ -71,9 +71,9 @@ class ScriptEnv(Environment, metaclass=abc.ABCMeta):
         self._shell_env_params: Optional[Iterable[str]] = self.config.get("shell_env_params")
         self._shell_env_params_rename: Dict[str, str] = self.config.get("shell_env_params_rename", {})
 
-        parse_results_stdout = self.config.get("parse_results_stdout")
-        self._parse_results_stdout: Optional[re.Pattern[str]] = \
-            re.compile(parse_results_stdout) if parse_results_stdout else None
+        results_stdout_pattern = self.config.get("results_stdout_pattern")
+        self._results_stdout_pattern: Optional[re.Pattern[str]] = \
+            re.compile(results_stdout_pattern) if results_stdout_pattern else None
 
     def _get_env_params(self) -> Dict[str, str]:
         """
@@ -114,7 +114,7 @@ class ScriptEnv(Environment, metaclass=abc.ABCMeta):
         results : Dict[str, float]
             A dictionary of results extracted from the stdout.
         """
-        if not self._parse_results_stdout:
+        if not self._results_stdout_pattern:
             return {}
-        _LOG.debug("Extract regex: '%s' from: '%s'", self._parse_results_stdout, stdout)
-        return {key: float(val) for (key, val) in self._parse_results_stdout.findall(stdout)}
+        _LOG.debug("Extract regex: '%s' from: '%s'", self._results_stdout_pattern, stdout)
+        return {key: float(val) for (key, val) in self._results_stdout_pattern.findall(stdout)}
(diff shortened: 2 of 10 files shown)


mlos_bench/mlos_bench/config/schemas/environments/common-environment-subschemas.json
```diff
@@ -79,6 +79,10 @@
                 "teardown": {
                     "description": "The command lines to execute for the teardown phase.",
                     "$ref": "#/$defs/command_lines"
+                },
+                "parse_results_stdout": {
```
bpkroth, line 83:
```suggestion
                "results_stdout_regex": {
```

or 

```suggestion
                "results_stdout_pattern": {
```

Minor naming suggestion to help understand what the item is quickly.

mlos_bench/mlos_bench/environments/script_env.py
```diff
@@ -68,6 +71,10 @@ def __init__(self,
         self._shell_env_params: Optional[Iterable[str]] = self.config.get("shell_env_params")
         self._shell_env_params_rename: Dict[str, str] = self.config.get("shell_env_params_rename", {})
 
+        parse_results_stdout = self.config.get("parse_results_stdout")
```
bpkroth, line 74:
```suggestion
        results_stdout = self.config.get("results_stdout_regex")
```

mlos_bench/mlos_bench/environments/script_env.py
```diff
@@ -92,3 +99,22 @@ def _get_env_params(self) -> Dict[str, str]:
...
+
+        Returns
+        -------
+        results : Dict[str, float]
+            A dictionary of results extracted from the stdout.
+        """
+        if not self._parse_results_stdout:
+            return {}
+        _LOG.debug("Extract regex: '%s' from: '%s'", self._parse_results_stdout, stdout)
+        return {key: float(val) for (key, val) in self._parse_results_stdout.findall(stdout)}
```
bpkroth, line 120: A thought: this will only support full word matches, correct?

Are groupings or replacements possible?

For instance (in Perl psuedocode) 

```python
   pat = re.sub(r'(foo)?bar(baz)?', r'\1\2')
   # returns one of foobaz, foo, baz
   # whereas findall() would return foobar, barbaz, foobarbaz
```

I don't know what type of output you have in mind for this, but I can imagine cases where there are extra characters in the output that we want to stip and reformat.

Could even be an array of those search/replace patterns actually.

mlos_bench/mlos_bench/tests/config/schemas/environments/test-cases/good/full/local_fileshare_env-full.jsonc
```diff
@@ -29,6 +29,8 @@
         "dump_params_file": "/tmp/dump_params.json",
         "dump_meta_file": "/tmp/dump_params_meta.json",
 
+        "parse_results_stdout": "(\\w+),([0-9.]+)",
+
         "read_results_file": "/tmp/results.csv",
```
bpkroth, line 34: Are these mutually exclusive?

Also, does parse results stdout belong in fileshare services, or just the scriptenv (remote/local exec)?

mlos_bench/mlos_bench/tests/environments/local/local_env_stdout_test.py
```diff
@@ -0,0 +1,66 @@
...
+    Print benchmark results to stdout and capture them in the LocalEnv.
+    """
+    local_env = create_local_env(tunable_groups, {
+        "run": [
+            "echo 'Benchmark results:'",  # This line should be ignored
+            "echo 'latency,111'",
+            "echo 'throughput,222'",
+            "echo 'score,0.999'",
+        ],
+        "parse_results_stdout": r"(\w+),([0-9.]+)",
```
bpkroth, line 24: Ah, this is actually more like a grep pattern, not parse/extract.  Is that correct?

bpkroth: When you called it parse, I was thinking of it more like a `sed`

mlos_bench/mlos_bench/config/schemas/environments/common-environment-subschemas.json
motus: I agree. I like the latter more

mlos_bench/mlos_bench/environments/script_env.py
motus: renamed to `results_stdout_pattern`

mlos_bench/mlos_bench/environments/script_env.py
motus: yes, we can easily strip the extra characters. In fact, this is the regex that I tested on our actual output of SysBench+MySQL:
```python
In [5]: re.findall(r"results-cumulative,[^,]*,[^,]*,(\w+),([0-9.]+)", stdout)
Out[5]:
[('time', '600.095825'),
 ('threads', '0'),
 ('events', '2074.051638'),
 ('total', '2074.051638'),
 # . . . .
]
```
where `stdout` looks like
```
Enable succeeded:
[stdout]
results-intermediate,2023-08-30 00:13:51,560.00,latency_pct,66.838153
results-intermediate,2023-08-30 00:13:51,560.00,errors,0.000000
. . . . . . .
results-cumulative,2023-08-30 00:14:31,600.10,time,600.095825
results-cumulative,2023-08-30 00:14:31,600.10,threads,0
results-cumulative,2023-08-30 00:14:31,600.10,events,2074.051638
results-cumulative,2023-08-30 00:14:31,600.10,total,2074.051638
. . . . . . .
```
that is, `re.findall()` extracts all tuples of regex groups from the text. The only thing we _cannot_ do is reordering of the groups, but I would not bother about it now. After all, this is intended to be a quick solution that a) fixes a bug that @eujing had discovered, and b) gives an ability to quickly extract the data from stdout without using additional scripts or an extra LocalEnv. We can always add more functionality later if we really need to

mlos_bench/mlos_bench/tests/config/schemas/environments/test-cases/good/full/local_fileshare_env-full.jsonc
motus: No, currently we just merge the results from stdout and the file when both parameters are present.

Parsing the results does not belong to the fileshare services - e.g. RemoteEnv has no access to fileshare (only the LocalFileShareEnv does), but we might want to parse its stdout

mlos_bench/mlos_bench/tests/environments/local/local_env_stdout_test.py
motus: yes, it's like grep in a sense that we can use it to match only some specific parts of the output, but then it must have two groups to extract the metric id and the value. This way we can discard large portions of stdout that do not contain the data we want to extract

mlos_bench/mlos_bench/environments/script_env.py
motus: oh, sorry I have not immediately realized what you meant. Yes, technically the user can write a regex that would have multiple overlapping matches. I don't really know how to make it foolproof - I would rather assume that the user really knows what they are doing :)

mlos_bench/mlos_bench/services/remote/azure/azure_services.py
```diff
@@ -789,10 +789,13 @@ def get_remote_exec_results(self, config: dict) -> Tuple[Status, dict]:
...
+            A dict can have an "stdout" key with the remote output.
         """
         _LOG.info("Check the results on VM: %s", config.get("vmName"))
         (status, result) = self.wait_host_operation(config)
-        if status.is_succeeded():
-            return (status, result.get("properties", {}).get("output", {}))
-        else:
+        _LOG.debug("Result: %s :: %s", status, result)
+        if not status.is_succeeded():
+            # TODO: Extract the telemetry and status from stdout, if available
```
bpkroth, line 798: Is that happening here, or a future PR?

mlos_bench/mlos_bench/environments/script_env.py
bpkroth: Would be good to file that bug to give more context.

mlos_bench/mlos_bench/tests/environments/local/local_env_stdout_test.py
bpkroth: Ack, makes more sense now with the example you posted above.  Thanks!

bpkroth (approved):

mlos_bench/mlos_bench/services/remote/azure/azure_services.py
motus: that will be the future PR

Issue #505, opened 2023-09-07
AzureFileShareService fails to upload a directory if path does not already exists on the shared storage


motus:
e.g. the following *directory* upload fails:
```javascript
{
    "from": "environments/apps/mysql/scripts/remote/common/",
    "to": "$experiment_id/$trial_id/scripts"
}
```
if path `$experiment_id/$trial_id/` does not already exist on the file share.

Uploading a *single file* works, though - and I currently use it as a workaround, e.g.:
```javascript
{
    // FIXME: A workaround for the bug in Azure FileShare Service.
    // Copy one known local file to trigger `mkdir -p` on the file share.
    "from": "environments/apps/mysql/scripts/remote/common/common.sh",
    "to": "$experiment_id/$trial_id/scripts/common.sh"
},
{
    "from": "environments/apps/mysql/scripts/remote/common/",
    "to": "$experiment_id/$trial_id/scripts"
}
```
@eujing want to look at it?

PR #506, opened 2023-09-07
Fix azure fileshare uploading of directories

eujing:
Fix for #505 when parent folders do not exist in the remote file share yet during an upload.


motus (approved):
yay, it works! Thanks for the fix!

PR #508, opened 2023-09-14
CI fixups

bpkroth:
- Permissions fixup for recent CI issues.
  See Also: https://github.com/microsoft/MLOS/actions/runs/6179287329
- Add support to run an uncached devcontainer build by including `NO_CACHE=true` in one of the last 10 commit messages being pushed.
- Slight modification to `licenseheaders` rule for more file types.
  See Also: https://github.com/johann-petrak/licenseheaders/pull/78

PR #509, opened 2023-09-18
Fixups to commit message checks to only happen during PR pushes

bpkroth:
Handles bugs from #508.

Also try and fix a quoting issues.


motus (approved):
perfect, thanks!

PR #510, opened 2023-09-18
Add SSH Support

### LINKED ISSUE #379: ssh environment support

bpkroth:
- Adds SSH support for remote exec and file copy operations via `asyncssy` library.  Closes #379 
- This requires an event loop thread to operate the async calls in the background without blocking the rest of our main thread operations or changing our APIs dramatically.  Instead we interact with all operations as futures after that.
- Add json schema config support for SSH

- Adds test infrastructure for running SSH servers inside containers via `pytest-docker` which uses `docker compose` to start/stop them as fixtures.  This *should* work in all dev environments (Linux host, WSL host, Windows host), but is quite sensitive to networking setups.

  - [x] test the background thread cleanup logic
  - [x] test error handling of a broken ssh endpoint
  - [x] test basic exec commands
  - [x] test host operations (e.g., reboot)
  - [x] test basic copy commands
  - [x] test for single connection reuse across services
  - [x] test multi-server support
  - [x] use random ports to avoid conflicts
  - provide some config examples for using this
  - add remote_env exec tests with this (increase code coverage)
    - Will do these in a future PR instead - #521

### DIFF
--- a/.devcontainer/scripts/run-devcontainer.sh
+++ b/.devcontainer/scripts/run-devcontainer.sh
@@ -19,18 +19,29 @@ cd "$repo_root"
 
 container_name="$repo_name.$(stat -c%i "$repo_root/")"
 
+# Be sure to use the host workspace folder if available.
+workspace_root=${LOCAL_WORKSPACE_FOLDER:-$repo_root}
+
+if [ -e /var/run/docker-host.sock ]; then
+    docker_gid=$(stat -c%g /var/run/docker-host.sock)
+else
+    docker_gid=$(stat -c%g /var/run/docker.sock)
+fi
+
+set -x
 mkdir -p "/tmp/$container_name/dc/shellhistory"
 docker run -it --rm \
     --name "$container_name" \
-    --user vscode \
+    --user vscode:$docker_gid \
     -v "$HOME/.azure":/dc/azure \
     -v "/tmp/$container_name/dc/shellhistory:/dc/shellhistory" \
     -v /var/run/docker.sock:/var/run/docker.sock \
-    -v "$PWD":"/workspaces/$repo_name" \
+    -v "$workspace_root":"/workspaces/$repo_name" \
     --workdir "/workspaces/$repo_name" \
     --env CONTAINER_WORKSPACE_FOLDER="/workspaces/$repo_name" \
-    --env LOCAL_WORKSPACE_FOLDER="$repo_root" \
+    --env LOCAL_WORKSPACE_FOLDER="$workspace_root" \
     --env http_proxy="${http_proxy:-}" \
     --env https_proxy="${https_proxy:-}" \
     --env no_proxy="${no_proxy:-}" \
-    mlos-devcontainer
+    mlos-devcontainer \
+    $*
--- a/.devcontainer/scripts/run-devcontainer.sh
+++ b/.devcontainer/scripts/run-devcontainer.sh
@@ -32,7 +32,8 @@ set -x
 mkdir -p "/tmp/$container_name/dc/shellhistory"
 docker run -it --rm \
     --name "$container_name" \
-    --user vscode:$docker_gid \
+    --user vscode \
+    --group-add $docker_gid \
     -v "$HOME/.azure":/dc/azure \
     -v "/tmp/$container_name/dc/shellhistory:/dc/shellhistory" \
     -v /var/run/docker.sock:/var/run/docker.sock \
--- a/.devcontainer/scripts/run-devcontainer.sh
+++ b/.devcontainer/scripts/run-devcontainer.sh
@@ -33,6 +33,7 @@ mkdir -p "/tmp/$container_name/dc/shellhistory"
 docker run -it --rm \
     --name "$container_name" \
     --user vscode \
+    --env USER=vscode \
     --group-add $docker_gid \
     -v "$HOME/.azure":/dc/azure \
     -v "/tmp/$container_name/dc/shellhistory:/dc/shellhistory" \
--- a/.github/workflows/devcontainer.yml
+++ b/.github/workflows/devcontainer.yml
@@ -127,6 +127,13 @@ jobs:
         # Now actually run the tests.
         docker exec --user vscode mlos-devcontainer make CONDA_INFO_LEVEL=-v test
 
+    - name: Debugging - Output Disk Usage
+      run: |
+        set -x
+        df -h
+        sudo -n du -hsc /var/lib/* /var/cache/* /tmp/*
+        du -hsc *
+
     - name: Generate and test binary distribution files
       timeout-minutes: 10
       run: |
--- a/.github/workflows/devcontainer.yml
+++ b/.github/workflows/devcontainer.yml
@@ -136,10 +136,11 @@ jobs:
     - name: Debugging - Output Disk Usage
       run: |
         set -x
-        df -h
-        sudo -n du -hsc /var/lib/* /var/cache/* /tmp/*
-        du -hsc *
-        docker image ls
+        df -h | sort -h
+        sudo df -h /var/lib/docker/* | sort -h | uniq
+        sudo -n du -hsc /var/lib/* /var/cache/* /tmp/* | sort -h
+        sudo du -hsc /mnt/* | sort -h
+        docker image ls | awk '{ printf "%s\t%s\n", $NF, $0 }' | sort -h
 
     - name: Generate and test binary distribution files
       timeout-minutes: 10
--- a/.github/workflows/devcontainer.yml
+++ b/.github/workflows/devcontainer.yml
@@ -91,7 +91,7 @@ jobs:
           --env PYTEST_EXTRA_OPTIONS=$PYTEST_EXTRA_OPTIONS \
           --workdir /workspaces/MLOS \
           --add-host host.docker.internal:host-gateway \
-          --name mlos-devcontainer mlos-devcontainer sleep infinity
+          --name mlos-devcontainer mlos-devcontainer sleep 1800
     - name: Fixup vscode uid/gid in the running container
       timeout-minutes: 3
       run: |
@@ -99,9 +99,6 @@ jobs:
         docker exec --user root mlos-devcontainer groupmod --non-unique -g `id -g` vscode
         docker exec --user root mlos-devcontainer usermod --non-unique -u `id -u` -g `id -g` vscode
         docker exec --user root mlos-devcontainer chown -R vscode:vscode /home/vscode
-        docker exec --user root mlos-devcontainer mkdir -p /opt/conda/pkgs/cache /var/cache/pip
-        docker exec --user root mlos-devcontainer chown -R vscode /opt/conda/pkgs/cache /var/cache/pip
-        docker exec --user root mlos-devcontainer chown -R vscode /opt/conda/envs/mlos
 
     - name: Update the conda env in the devcontainer
       timeout-minutes: 10
@@ -143,6 +140,7 @@ jobs:
       timeout-minutes: 3
       run: |
         set -x
+        git --no-pager diff --exit-code
         docker exec --user vscode mlos-devcontainer make CONDA_INFO_LEVEL=-v devcontainer
 
     - name: Generate docs and test check them
--- a/.github/workflows/devcontainer.yml
+++ b/.github/workflows/devcontainer.yml
@@ -149,7 +149,7 @@ jobs:
         docker exec --user vscode mlos-devcontainer make CONDA_INFO_LEVEL=-v dist dist-test
 
     - name: Test rebuilding the devcontainer in the devcontainer
-      timeout-minutes: 3
+      timeout-minutes: 7
       run: |
         set -x
         docker exec --user vscode mlos-devcontainer make CONDA_INFO_LEVEL=-v devcontainer
--- a/.github/workflows/devcontainer.yml
+++ b/.github/workflows/devcontainer.yml
@@ -51,6 +51,11 @@ jobs:
       run: |
         echo "NO_CACHE=true" >> $GITHUB_ENV
 
+    - name: Run pytest with debug logging enabled for nightly runs
+      if: github.event_name == 'schedule'
+      run: |
+        echo "PYTEST_EXTRA_OPTIONS=--log-level=DEBUG" >> $GITHUB_ENV
+
     - name: Log some environment variables for debugging
       run: |
         set -x
@@ -83,6 +88,7 @@ jobs:
           --volume $(pwd):/workspaces/MLOS \
           --env CONTAINER_WORKSPACE_FOLDER=/workspaces/MLOS \
           --env LOCAL_WORKSPACE_FOLDER=$(pwd) \
+          --env PYTEST_EXTRA_OPTIONS=$PYTEST_EXTRA_OPTIONS \
           --workdir /workspaces/MLOS \
           --add-host host.docker.internal:host-gateway \
           --name mlos-devcontainer mlos-devcontainer sleep infinity
--- a/.github/workflows/devcontainer.yml
+++ b/.github/workflows/devcontainer.yml
@@ -100,17 +100,21 @@ jobs:
         docker exec --user root mlos-devcontainer usermod --non-unique -u `id -u` -g `id -g` vscode
         docker exec --user root mlos-devcontainer chown -R vscode:vscode /home/vscode
 
+    - name: Print some debug info from inside the container
+      run: |
+        docker exec --user vscode --env USER=vscode mlos-devcontainer printenv
+
     - name: Update the conda env in the devcontainer
       timeout-minutes: 10
       run: |
         set -x
-        docker exec --user vscode mlos-devcontainer make CONDA_INFO_LEVEL=-v conda-env
+        docker exec --user vscode --env USER=vscode mlos-devcontainer make CONDA_INFO_LEVEL=-v conda-env
 
     - name: Check for missing licenseheaders
       timeout-minutes: 3
       run: |
         set -x
-        docker exec --user vscode mlos-devcontainer make CONDA_INFO_LEVEL=-v licenseheaders
+        docker exec --user vscode --env USER=vscode mlos-devcontainer make CONDA_INFO_LEVEL=-v licenseheaders
         # licenseheaders changes the contents of the files, so make this check fail if there are any changes detected
         git --no-pager diff --exit-code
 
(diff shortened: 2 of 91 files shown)


conftest.py
```diff
@@ -24,4 +24,4 @@ def pytest_configure(config: pytest.Config) -> None:   # pylint: disable=unused-
         import matplotlib   # pylint: disable=import-outside-toplevel
         matplotlib.rcParams['backend'] = 'agg'
         warn(UserWarning('DISPLAY environment variable is set, which can cause problems in some setups (e.g. WSL). '
-            + f'Adjusting matplotlib backend to "{matplotlib.rcParams["backend"]}" to compensate.'))
+                         + f'Adjusting matplotlib backend to "{matplotlib.rcParams["backend"]}" to compensate.'))
```
bpkroth, line 27: Unrelated pylint quibble.

mlos_bench/mlos_bench/config/schemas/services/remote/azure/azure-auth-service-subschema.json
```diff
@@ -55,5 +55,5 @@
             "unevaluatedProperties": false
         }
     },
-    "required": ["class", "config"]
+    "required": ["class"]
```
bpkroth, line 58: Don't make empty configs required.

mlos_bench/mlos_bench/services/types/fileshare_type.py
```diff
@@ -31,12 +33,14 @@ def download(self, remote_path: str, local_path: str, recursive: bool = True) ->
             if True (the default), download the entire directory tree.
         """
 
-    def upload(self, local_path: str, remote_path: str, recursive: bool = True) -> None:
+    def upload(self, params: dict, local_path: str, remote_path: str, recursive: bool = True) -> None:
```
bpkroth, line 36: Required for ssh fileshare service.
- [ ] Might want to break this out as well.

mlos_bench/mlos_bench/tests/__init__.py
```diff
@@ -9,9 +9,18 @@
...
+import shutil
+
+import pytest
+
 from mlos_bench.util import get_class_from_name
 
 
+# A decorator for tests that require docker.
+DOCKER = shutil.which('docker')
+requires_docker = pytest.mark.skipif(not DOCKER, reason='Docker is not available on this system.')
```
bpkroth, line 22:
```suggestion
# Use with @requires_docker above a `test_...` function.
requires_docker = pytest.mark.skipif(not DOCKER, reason='Docker is not available on this system.')
```

mlos_bench/mlos_bench/tests/services/remote/ssh/ssh_hacking_test.py
bpkroth, line 1: Probably remove this file in the final PR.

mlos_bench/mlos_bench/tests/services/remote/ssh/conftest.py
```diff
@@ -0,0 +1,74 @@
...
+    # On Linux, if we're running in a docker container, we can use the
+    # --add-host (extra_hosts in docker-compose.yml) to refer to the host IP.
+    if resolve_host_name('host.docker.internal'):
+        return 'host.docker.internal'
+    # Otherwise, assume we're executing directly inside conda on the host.
+    return 'localhost'
+
+
+@pytest.fixture(scope="session")
+def ssh_test_server(ssh_test_server_hostname: str,
```
bpkroth, line 41: - [ ] Add a second test server.

conftest.py
bpkroth: Maybe move to a separate PR

mlos_bench/mlos_bench/tests/services/remote/ssh/ssh_hacking_test.py
bpkroth, line 1: Remove this file?

mlos_bench/mlos_bench/config/schemas/services/remote/azure/azure-auth-service-subschema.json
bpkroth: See #518

mlos_bench/mlos_bench/services/types/fileshare_type.py
bpkroth: Moved to #517

mlos_bench/mlos_bench/config/schemas/config_schemas.py
```diff
@@ -137,5 +137,5 @@ def validate(self, config: dict) -> None:
         else:
             jsonschema.Draft202012Validator(
                 schema=self.schema,
-                registry=SCHEMA_STORE.registry,     # type: ignore[call-arg]
```
bpkroth, line 140: removes an unnecessary call (unrelated)

mlos_bench/mlos_bench/tests/config/cli/test_load_cli_config_examples.py
```diff
@@ -54,6 +54,7 @@ def filter_configs(configs_to_filter: List[str]) -> List[str]:
 @pytest.mark.parametrize("config_path", configs)
 def test_load_cli_config_examples(config_loader_service: ConfigPersistenceService, config_path: str) -> None:
     """Tests loading a config example."""
+    # pylint: disable=too-complex
```
bpkroth, line 57: Remove, unnecessary?

bpkroth: Nope, it is.  Though I'm not sure why it's showing up here since I didn't change it ...

mlos_bench/mlos_bench/tests/config/__init__.py
```diff
@@ -8,7 +8,6 @@
 
 from typing import Callable, List, Optional
 
-from glob import iglob
```
bpkroth, line 11: Unrelated pylint catch.

bpkroth:
Runs fine locally from what I've been able to tell.  Debugging CI environment issues moreso.

.devcontainer/scripts/run-devcontainer.sh
```diff
@@ -19,18 +19,29 @@ cd "$repo_root"
 
 container_name="$repo_name.$(stat -c%i "$repo_root/")"
 
+# Be sure to use the host workspace folder if available.
+workspace_root=${LOCAL_WORKSPACE_FOLDER:-$repo_root}
+
+if [ -e /var/run/docker-host.sock ]; then
```
bpkroth, line 25: This makes it so we can use this script for local testing even inside the devcontainer.

.github/workflows/devcontainer.yml
```diff
@@ -84,6 +84,7 @@ jobs:
           --env CONTAINER_WORKSPACE_FOLDER=/workspaces/MLOS \
           --env LOCAL_WORKSPACE_FOLDER=$(pwd) \
           --workdir /workspaces/MLOS \
+          --add-host host.docker.internal:host-gateway \
```
bpkroth, line 87: Mimics the changes to `devcontainer/docker-compose.yml` so we can know which endpoint to talk to for tests.

See Also: `mlos_bench/tests/services/remote/ssh/conftest.py`

mlos_bench/mlos_bench/tests/services/remote/ssh/test_ssh_host_service.py
```diff
@@ -0,0 +1,207 @@
...
+        script=["hostname"],
+        config=alt_test_server.to_ssh_service_config(),
+        env_params={
+            "UNUSED": "unused",  # unused, making sure it doesn't carry over with cached connections
+        },
+    )
+    assert status == Status.PENDING
+    assert "asyncRemoteExecResultsFuture" in results_info
+    status, results = ssh_host_service.get_remote_exec_results(results_info)
+    assert status == Status.SUCCEEDED
```
motus, line 84:
```suggestion
    assert status.is_succeeded()
```
if we change all checks like this, we can get by without
```python
from mlos_bench.environments.status import Status
```

mlos_bench/mlos_bench/tests/services/remote/ssh/test_ssh_host_service.py
```diff
@@ -0,0 +1,207 @@
...
+    # Note: rebooting changes the port number unfortunately, but makes it
+    # easier to check for success.
+    # Also, it may cause issues with other parallel unit tests, so we run it as
+    # a part of the same unit test for now.
+    alt_test_server_ssh_service_config = alt_test_server.to_ssh_service_config()
+    (status, results_info) = ssh_host_service.remote_exec(
+        script=[
+            "echo \"sleeping...\"",
+            "sleep 30",
+            "echo \"shouldn't reach this point\""
```
motus, line 156:
```suggestion
            'echo "sleeping..."',
            'sleep 30',
            'echo "shouldn't reach this point"'
```
but maybe `\"` is more explicit

motus (approved):
First pass. Will review the unit tests more thoroughly tomorrow

mlos_bench/mlos_bench/tests/services/remote/ssh/test_ssh_fileshare.py
```diff
@@ -0,0 +1,186 @@
...
+        temp_file.writelines(lines)
+        temp_file.flush()
+        temp_file.close()
+
+        ssh_fileshare_service.upload(
+            params=config,
+            local_path=temp_file.name,
+            remote_path=remote_file_path,
+        )
+        os.unlink(temp_file.name)
```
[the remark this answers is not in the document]
motus: btw, why doesn't `contextlib` has something generic like
```python
@contextlib.contextmanager
def on_exit(func_cleanup):
    try:
        yield
    finally:
        func_cleanup()
```

Then we would write 
```python
with tempfile.NamedTemporaryFile(mode='w+t', encoding='utf-8', delete=False) as temp_file, \
     on_exit(lambda: os.unlink(temp_file.name)):
    # ...
```

mlos_bench/mlos_bench/services/remote/ssh/ssh_host_service.py
```diff
@@ -0,0 +1,261 @@
...
+            required_keys=[
+                "ssh_hostname",
+            ]
+        )
+        cmd_opts = ' '.join([f"'{cmd}'" for cmd in cmd_opts_list])
+        script = r"if [[ $EUID -ne 0 ]]; then sudo=$(command -v sudo); sudo=${sudo:+$sudo -n}; fi; " \
+            + f"for cmd in {cmd_opts}; do " \
+            + r"  $sudo $cmd && exit 0;" \
+            + r"done;" \
+            + r"echo 'ERROR: Failed to shutdown/reboot the system.'; exit 1"
```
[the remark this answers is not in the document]
bpkroth: I'll double check this, but I was worried about how sending newlines across the wire would work.

mlos_bench/mlos_bench/services/remote/ssh/ssh_service.py
```diff
@@ -0,0 +1,393 @@
...
+                SshService._event_loop_thread_ssh_client_cache = SshClientCache()
+                assert SshService._event_loop is None
+                SshService._event_loop = asyncio.new_event_loop()
+                SshService._event_loop_thread = Thread(target=SshService._run_event_loop, daemon=True)
+                SshService._event_loop_thread.start()
+            SshService._event_loop_thread_refcnt += 1
+
+        self._enabled = True
+
+    def close(self) -> None:    # pylint: disable=no-self-use
```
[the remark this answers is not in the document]
bpkroth: Yep, earlier rendition of this was a class method, but then I added the `self._enabled` check.  Thanks for checking that.

mlos_bench/mlos_bench/services/remote/ssh/ssh_service.py
```diff
@@ -0,0 +1,393 @@
...
+    See additional details in SshService comments.
+    """
+
+    def __init__(self) -> None:
+        self._cache: Dict[str, Tuple[SSHClientConnection, SshClient]] = {}
+        self._cache_lock = CoroLock()
+
+    def __del__(self) -> None:
+        self.cleanup()
+
```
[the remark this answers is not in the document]
bpkroth: I looked at that briefly, but it gets pretty complicated for our Services due to the way that mixins work.  We'd need to do something like a register enter/exit callbacks for each Service and then order may matter and ...

I can keep digging into that rock if you want, but in my testing so far this seems to work OK.

The only time it really matters to do the cleanup is on program exit anyways since all we need to do is stop a background thread.  The main reason to add the explicit `close()` call was for testing, and that's only to make the call to the cleanup code explicit so I could test it.

Thoughts?

mlos_bench/mlos_bench/tests/services/remote/ssh/test_ssh_host_service.py
bpkroth: Work for me, though that doesn't autocomplete for some reason in the editor :shrug:

mlos_bench/mlos_bench/services/remote/ssh/ssh_service.py
```diff
@@ -0,0 +1,393 @@
...
+    def __init__(self, config: dict, global_config: dict, parent: Optional[Service]):
+        # pylint: disable=too-complex
+
+        super().__init__(config, global_config, parent)
+
+        # Make sure that the value we allow overriding on a per-connection
+        # basis are present in the config so merge_parameters can do its thing.
+        self.config.setdefault('ssh_port', None)
+        assert isinstance(self.config['ssh_port'], (int, type(None)))
```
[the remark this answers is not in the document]
bpkroth: `NoneType` doesn't exist in older pythons

mlos_bench/mlos_bench/services/remote/ssh/ssh_host_service.py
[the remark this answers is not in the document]
bpkroth: Had to change the way we combine commands but I think it works.

mlos_bench/mlos_bench/tests/services/remote/ssh/test_ssh_fileshare.py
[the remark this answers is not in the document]
bpkroth: Done.

mlos_bench/mlos_bench/services/remote/ssh/ssh_service.py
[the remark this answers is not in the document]
bpkroth: @motus did you have further thoughts on this thread?  Punt for now and revisit later or try and introduce contextmanager support for all Services first?

mlos_bench/mlos_bench/services/remote/ssh/ssh_fileshare.py
motus, line 1: need to double check that the right methods get registered when we don't define the constructor explicitly.
(Note to self: write a unit test for this case)

mlos_bench/mlos_bench/services/remote/ssh/ssh_service.py
```diff
@@ -0,0 +1,391 @@
...
+    # Using a larger thread pool requires a bit more work since asyncssh
+    # requires that run() requests are submitted to the same event loop handler
+    # that the connection was made on.
+    # In that case, each background thread should get its own SshClientCache.
+
+    _event_loop: Optional[AbstractEventLoop] = None
+    _event_loop_thread: Optional[Thread] = None
+    _event_loop_thread_lock = ThreadLock()
+    _event_loop_thread_refcnt: int = 0
+    _event_loop_thread_ssh_client_cache: Optional[SshClientCache] = None
```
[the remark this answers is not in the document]
motus: or maybe even roll all these variables into a single `@dataclass` struct with enter/exit methods so we can write e.g., `SshService._CONTEXT.enter(...)` or `SshService._CONTEXT.clear_cache()`or access the stuff in it like `SshService._CONTEXT.event_loop`

mlos_bench/mlos_bench/services/remote/ssh/ssh_service.py
```diff
@@ -0,0 +1,379 @@
...
+                known_hosts_file = os.path.expanduser(self._connect_params['known_hosts'])
+                if not os.path.exists(known_hosts_file):
+                    raise ValueError(f"ssh_known_hosts_file {known_hosts_file} does not exist")
+                self._connect_params['known_hosts'] = known_hosts_file
+        if self._connect_params['known_hosts'] is None:
+            _LOG.info("%s known_hosts checking is disabled per config.", self.__class__.__name__)
+
+        if 'ssh_keepalive_interval' in self.config:
+            keepalive_internal = self.config.get('ssh_keepalive_interval')
+            self._connect_params['keepalive_interval'] = int(keepalive_internal) if keepalive_internal is not None else None
```
motus, line 241:
```suggestion
            self._connect_params['keepalive_interval'] = None if if keepalive_internal is None else int(keepalive_internal)
```
4 bytes shorter 😄

mlos_bench/mlos_bench/services/remote/ssh/ssh_service.py
```diff
@@ -0,0 +1,379 @@
...
+
+        if 'ssh_known_hosts_file' in self.config:
+            self._connect_params['known_hosts'] = self.config.get("ssh_known_hosts_file", None)
+            if isinstance(self._connect_params['known_hosts'], str):
+                known_hosts_file = os.path.expanduser(self._connect_params['known_hosts'])
+                if not os.path.exists(known_hosts_file):
+                    raise ValueError(f"ssh_known_hosts_file {known_hosts_file} does not exist")
+                self._connect_params['known_hosts'] = known_hosts_file
+        if self._connect_params['known_hosts'] is None:
+            _LOG.info("%s known_hosts checking is disabled per config.", self.__class__.__name__)
```
motus, line 237:
```suggestion
            _LOG.info("%s known_hosts checking is disabled per config.", self)
```
if we have `__repr__` or `__str__` defined (+ a case for human-readable `str()`)

mlos_bench/mlos_bench/services/remote/ssh/ssh_service.py
```diff
@@ -0,0 +1,379 @@
...
+        # Start the background thread if it's not already running.
+        with SshService._event_loop_thread_lock:
+            if not SshService._event_loop_thread:
+                assert SshService._event_loop_thread_ssh_client_cache is None
+                SshService._event_loop_thread_ssh_client_cache = SshClientCache()
+                assert SshService._event_loop is None
+                SshService._event_loop = asyncio.new_event_loop()
+                SshService._event_loop_thread = Thread(target=SshService._run_event_loop, daemon=True)
+                SshService._event_loop_thread.start()
+            SshService._event_loop_thread_refcnt += 1
```
motus, line 253: I would move it into some sort of `EventLoopContext` class method that takes thread target function as a parameter

motus (approved):
Love it!

mlos_bench/mlos_bench/services/remote/ssh/ssh_service.py
bpkroth: Not sure I follow ...

mlos_bench/mlos_bench/services/remote/ssh/ssh_service.py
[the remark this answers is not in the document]
bpkroth: I split it out a little bit differently so that it hopefully supports other future uses as well (e.g., parallel trial runners perhaps).
It keeps the event loop manager separately from the client cache, since the latter is somewhat specific to the SshService.
Let me know what you think: #555

mlos_bench/mlos_bench/services/remote/ssh/ssh_fileshare.py
bpkroth: An explicit test for this is not a bad idea, but we're already validating this with these end-to-end tests as well (i.e., file copy operations with ssh work)

mlos_bench/mlos_bench/services/remote/ssh/ssh_service.py
bpkroth: I think we don't need that any more given the way I refactored the EventLoopContext to be a separate reusable entity.

mlos_bench/mlos_bench/services/remote/ssh/ssh_fileshare.py
bpkroth: Started a stub for this here: #556

Issue #362, opened 2023-05-18
Create a multi-VM example for Azure


motus:
For instance, provide an ARM Template for two VMs: one for MySQL server and another for MySQL BenchBase client. Also, implement the corresponding scripts and configs and make sure our framework handles multi-VM setups correctly.

bpkroth:
See Also #329

bpkroth:
#455 does this

bpkroth:
Nm, I guess that was still just single VM.

PR #513, opened 2023-09-19
Unit tests for combining the results of several `LocalEnv`s in one `CompositeEnv`

### LINKED ISSUE #501: Add tests for environment variable population in shell scripts derived from composite environments
Would be good to also test this in a composite nested env.
That's where I've run into issues previously.

_Originally posted by @bpkroth in https://github.com/microsoft/MLOS/pull/484#discussion_r1293906386_

motus:
Also, rename some local functions to reflect the fact that the environment may not necessarily be an instance of `LocalEnv` and change the docstrings and annotations accordingly.

Closes #501

### DIFF
--- a/mlos_bench/mlos_bench/tests/environments/local/__init__.py
+++ b/mlos_bench/mlos_bench/tests/environments/local/__init__.py
@@ -40,6 +40,7 @@ def create_local_env(tunable_groups: TunableGroups, config: Dict[str, Any]) -> L
 
 
 def create_composite_local_env(tunable_groups: TunableGroups,
+                               params: Dict[str, Any],
                                local_configs: List[Dict[str, Any]]) -> CompositeEnv:
     """
     Create a CompositeEnv with several LocalEnv instances.
@@ -48,6 +49,8 @@ def create_composite_local_env(tunable_groups: TunableGroups,
     ----------
     tunable_groups : TunableGroups
         Tunable parameters (usually come from a fixture).
+    params: Dict[str, Any]
+        Additional config params for the CompositeEnv.
     local_configs: List[Dict[str, Any]]
         Configurations of the local environments.
 
@@ -59,6 +62,7 @@ def create_composite_local_env(tunable_groups: TunableGroups,
     return CompositeEnv(
         name="TestCompositeEnv",
         config={
+            **params,
             "children": [
                 {
                     "name": f"TestLocalEnv{i}",
--- a/mlos_bench/mlos_bench/tests/environments/local/__init__.py
+++ b/mlos_bench/mlos_bench/tests/environments/local/__init__.py
@@ -40,6 +40,7 @@ def create_local_env(tunable_groups: TunableGroups, config: Dict[str, Any]) -> L
 
 
 def create_composite_local_env(tunable_groups: TunableGroups,
+                               global_config: Dict[str, Any],
                                params: Dict[str, Any],
                                local_configs: List[Dict[str, Any]]) -> CompositeEnv:
     """
@@ -49,6 +50,8 @@ def create_composite_local_env(tunable_groups: TunableGroups,
     ----------
     tunable_groups : TunableGroups
         Tunable parameters (usually come from a fixture).
+    global_config : Dict[str, Any]
+        Global configuration parameters.
     params: Dict[str, Any]
         Additional config params for the CompositeEnv.
     local_configs: List[Dict[str, Any]]
@@ -73,6 +76,7 @@ def create_composite_local_env(tunable_groups: TunableGroups,
             ]
         },
         tunables=tunable_groups,
+        global_config=global_config,
         service=LocalExecService(parent=ConfigPersistenceService()),
     )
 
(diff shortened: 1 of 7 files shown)


mlos_bench/mlos_bench/tests/environments/local/__init__.py
```diff
@@ -30,32 +32,70 @@ def create_local_env(tunable_groups: TunableGroups, config: Dict[str, Any]) -> L
...
+                }
+                for (i, config) in enumerate(local_configs)
+            ]
+        },
+        tunables=tunable_groups,
+        service=LocalExecService(parent=ConfigPersistenceService()),
+    )
+
+
+def check_env_success(env: Environment,
```
motus, line 80: was: `check_local_env_success`, but the input env does not have to be `LocalEnv`. Hence the rename

mlos_bench/mlos_bench/tests/environments/local/composite_local_env_test.py
```diff
@@ -0,0 +1,92 @@
...
+                },
+                "shell_env_params": ["throughput", "score"],
+                "run": [
+                    "echo 'metric,value' > output.csv",
+                    f"echo 'throughput,{var_prefix}throughput{var_suffix}' >> output.csv",
+                    f"echo 'score,{var_prefix}score{var_suffix}' >> output.csv",
+                    "echo '-------------------'",  # This output does not go anywhere
+                    "echo 'timestamp,metric,value' > telemetry.csv",
+                    f"echo {time_str2},cpu_load,0.79 >> telemetry.csv",
+                    f"echo {time_str2},mem_usage,40960 >> telemetry.csv",
```
bpkroth, line 70: What happens if these two separate environments submit results with the same timestamp and same metrics?  Do we have a first writer wins policy, or do we prefix one with the environment name or something?

Seems relevant for two scenarios:
- "sidecar" style monitors that we tack on to other applications to gather generic system metrics
- multi-vm setups

motus: Yes, right now we have the "first writer wins" policy, which is implemented in the storage layer. That is, `CompositeEnv` would just collect and concatenate the telemetry records from all environments and send everything as one big `DataFrame` to the storage layer. In the database, the `trial_telemetry` table has a unique constraint of `{experiment_id, trial_id, timestamp, metric_id}`, so only the first record would get inserted. We won't crash when inserting the next one, though - it will be just ignored by the storage layer and we'll continue with inserting other recoreds

motus: We don't automatically prefix our `metric_id`s with VM name or anything - I think this is the responsibility of the user to do so, and it is pretty easy to implement in our configs/scripts. I won't overengineer it right now

motus:
> have something like `BENCHMARK=benchmark-name` in the global config, and pass that on down through via environment variables to the `benchmark-env.jsonc` without having to also chain that `BENCHMARK` arg as a `required_arg` or `const_arg` through the full other stack of configs

Oh, I see. I can easily add that case to our test

mlos_bench/mlos_bench/tests/environments/local/composite_local_env_test.py
bpkroth: Examples of doing it might be helpful later at least.

motus: >> We don't automatically prefix our `metric_ids` with VM name or anything - I think this is the responsibility of the user to do so

> Examples of doing it might be helpful later at least.

Agree. We need a multi-VM example in our main repo to demo that, though. I'll create an issue so we won't forget about it when we finally implement that multi-VM example.

bpkroth: #362 is already there. Maybe add an additional task to it.

mlos_bench/mlos_bench/tests/environments/local/composite_local_env_test.py
```diff
@@ -0,0 +1,110 @@
...
+from datetime import datetime, timedelta
+
+from mlos_bench.tunables.tunable_groups import TunableGroups
+from mlos_bench.tests.environments.local import create_composite_local_env, check_env_success
+
+
+def test_composite_env(tunable_groups: TunableGroups) -> None:
+    """
+    Produce benchmark and telemetry data in TWO local environments
+    and combine the results.
```
bpkroth, line 18:
```suggestion
    and combine the results.
    Also checks that global configs flow down at least one level of CompositeEnv
    to its children without being explicitly specified in the CompositeEnv so they
    can be used in the shell_envs by its children.
    See Also: http://github.com/microsoft/MLOS/issues/501
```

bpkroth (approved):
One small doc tweak suggested.

mlos_bench/mlos_bench/tests/environments/local/composite_local_env_test.py
bpkroth: Done.

PR #518, opened 2023-09-28
Don't require empty config element in service json schemas

bpkroth:
Adjusts service schema's to not require an empty `"config": {}` element, and instead only include it if there's some other property in it.

Split out from #510


conftest.py
```diff
@@ -24,4 +24,4 @@ def pytest_configure(config: pytest.Config) -> None:   # pylint: disable=unused-
         import matplotlib   # pylint: disable=import-outside-toplevel
         matplotlib.rcParams['backend'] = 'agg'
         warn(UserWarning('DISPLAY environment variable is set, which can cause problems in some setups (e.g. WSL). '
-            + f'Adjusting matplotlib backend to "{matplotlib.rcParams["backend"]}" to compensate.'))
+                         + f'Adjusting matplotlib backend to "{matplotlib.rcParams["backend"]}" to compensate.'))
```
bpkroth, line 27: Minor autopep8 reformat.

motus (approved):

Issue #498, opened 2023-08-25
Separate "teardown" from "cleanup" phases


bpkroth:
In the `AzureVMService` right now we have a `teardown` phase that calls `vm_deprovision` (soon to be `deprovision_host`).

However, it doesn't actually deprovision the host.  Rather, it currently calls `deallocate`:
https://github.com/microsoft/MLOS/blob/83497e2a04aa3c367c2c2579a80ca7019f2a379f/mlos_bench/mlos_bench/services/remote/azure/azure_services.py#L80
which simply releases the VM's node binding (thus freeing Compute cost from incurring), but it doesn't actually remove the VM disks, NIC, IPs, etc. (which have their own cost)

I think there's value in having a "teardown" mode that simply shutsdown and deallocates the VM
(as an aside, shutting down the VM inside the guest OS, doesn't deallocate the VM) without removing it entirely.

Additionally, some non-VM related resources could use "teardown" phases to manage cleaning up after a workload run, so the `--no-teardown` option isn't super great for that either.

Moreover, as mentioned in #468, there are other needs for more complete cleanup scripts to make sure we aren't over spending when we're done with experiment resources.

I think an approach we could take is to separate more completely the "teardown" from "cleanup" phases, the latter potentially only invoked manually when an experiment is deemed to be over, whereas we may want to do "deallocate" more frequently.

"cleanup" should probably only handle certain things like VMs and disks, but not shared storage, shared DB, shared VNets, etc.

bpkroth:
See also #404

Issue #404, opened 2023-06-20
mlos_bench: examples/abstractions for modeling reconfig vs. restart


bpkroth:
Some system parameters require reloading data or restarting to take effect.  We should make sure that we have examples at least or new abstractions for those.

bpkroth:
Another related thing:
- reallocate vs. deprovision
This is especially important for VMs as dellocation (without deleting the original resources), is slightly more expensive than a reboot, since it involves removing the VM from a node and rescheduling it (possibly/hopefully elsewhere) in the cloud system (Azure), which can be important for conducting repeat evaluations of a config across space and time for determining config robustness to different sources of noise.

PR #520, opened 2023-09-29
SaaS Environment + Azure Flex DB configuration service

motus:


mlos_bench/mlos_bench/services/remote/azure/azure_flex.py
```diff
@@ -0,0 +1,207 @@
...
+            "subscription",
+            "resourceGroup",
+            "provider",
+        })
+
+        (self._is_batch, is_flex, api_version) = {
+            "Microsoft.DBforMySQL": (True, True, "2022-01-01"),
+            "Microsoft.DBforMariaDB": (False, False, "2018-06-01"),
+            "Microsoft.DBforPostgreSQL": (False, True, "2022-12-01"),
+        }[self.config["provider"]]
```
bpkroth, line 72: This makes me a little nervous - seems like a support headache.  Is there a more generic way we can do more of this with just configs?

bpkroth, line 72: This makes me a little nervous - seems like a support headache.  Is there a more generic way we can do more of this with just configs?

motus: yes, we can move all these parameters to the config - in fact, I had it this way a few commits back. Then, say, a `[service-flex-mysql.jsonc](https://github.com/microsoft/MLOS/pull/520/files#diff-c2b0c5fd6d3c817c136510b21212d9b22212b0a9254228f7593bb0fe460b2c72)` would look like this:
```javascript
{
    "class": "mlos_bench.services.remote.azure.AzureFlexConfigService",

    "config": {

        "subscription": "PLACEHOLDER; AZURE SUBSCRIPTION ID",
        "resourceGroup": "PLACEHOLDER; e.g., os-autotune",
        "vmName": "PLACEHOLDER; e.g., mlos-vm-mysql",

        "provider": "Microsoft.DBforMySQL",
        "supportsBatchUpdate": true,
        "isFlex": true,
        "apiVersion": "2022-01-01",

        "requestTimeout": 20,   // HTTP request timeout in seconds
        "pollInterval": 20,     // Poll every 20 seconds
        "pollTimeout": 600      // Give up polling after 10 minutes
    }
}
```
I am fine with either way

bpkroth: The config only way is less friendly, but more easily fixable and expandable overtime with fewer interactions from us maintainers, so I think it might be the better way to go.
One thing we could do is bake in a few "currently known sane defaults" when some of the config values aren't provided.
Either way this would need to be well documented.

bpkroth: Also, can the Service be just `remote.azure.AzureSaaSService` instead of specifically `Flex` services?  Or is that too difficult?

motus: no, I think it makes sense. Not all DBs on Azure are Flex servers, anyway

motus: also, just pushed a new version of the service and configs

mlos_bench/mlos_bench/services/remote/azure/azure_saas.py
```diff
@@ -63,22 +63,20 @@ def __init__(self,
...
+            "supportsBatchUpdate",
+            "isFlex",
+            "apiVersion"
         })
 
-        (self._is_batch, is_flex, api_version) = {
-            "Microsoft.DBforMySQL": (True, True, "2022-01-01"),
-            "Microsoft.DBforMariaDB": (False, False, "2018-06-01"),
-            "Microsoft.DBforPostgreSQL": (False, True, "2022-12-01"),
-        }[self.config["provider"]]
```
bpkroth: I think we can keep this as a convenience fallback for when those values aren't provided in the config BTW.

mlos_bench/mlos_bench/config/schemas/services/remote/azure/azure-flex-service-subschema.json
```diff
@@ -39,12 +39,28 @@
                                 "Microsoft.DBforPostgreSQL",
                                 "Microsoft.DBforMariaDB"
                             ]
+                        },
+                        "supportsBatchUpdate": {
+                            "description": "Whether API supports updating multiple config parameters at once",
+                            "type": "boolean"
+                        },
+                        "isFlex": {
+                            "description": "Whether SaaS is an Azure Flexible Server",
```
bpkroth, line 48: Also provide a documentation URL in the description for additional details?

mlos_bench/mlos_bench/config/schemas/services/remote/azure/azure-flex-service-subschema.json
```diff
@@ -39,12 +39,28 @@
...
+                            "type": "boolean"
+                        },
+                        "apiVersion": {
+                            "description": "Version of Azure REST API, e.g., 2022-01-01-preview",
```
bpkroth, line 52: Also provide a documentation URL where one could go look these up?
Same for the `provider` param.

mlos_bench/mlos_bench/config/schemas/services/remote/azure/azure-flex-service-subschema.json
```diff
@@ -39,12 +39,28 @@
                                 "Microsoft.DBforPostgreSQL",
                                 "Microsoft.DBforMariaDB"
```
bpkroth, line 40: Change the type from `enum` to `string` and list the known values in `examples`?

mlos_bench/mlos_bench/services/remote/azure/azure_saas.py
motus: oh, you're right. Let me bring it back so we have both ways to configure things

mlos_bench/mlos_bench/services/remote/azure/azure_saas.py
```diff
@@ -20,14 +20,14 @@
...
+class AzureSaaSConfigService(Service, SupportsRemoteConfig):
     """
     Helper methods to configure Azure Flex services.
     """
 
     _REQUEST_TIMEOUT = 5  # seconds
 
-    # Azure Flex Services Configuration REST API as described in
+    # Azure SaaS DB Services Configuration REST API as described in
     # https://learn.microsoft.com/en-us/rest/api/mysql/flexibleserver/configurations
```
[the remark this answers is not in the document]
motus: done

mlos_bench/mlos_bench/config/schemas/services/remote/azure/azure-flex-service-subschema.json
motus: sounds good - I'll also implement better handling of unknown provider values on the service side

mlos_bench/mlos_bench/config/schemas/services/remote/azure/azure-flex-service-subschema.json
motus: done

mlos_bench/mlos_bench/config/schemas/services/remote/azure/azure-flex-service-subschema.json
motus: yep

motus:
Huge thanks @bpkroth for reviewing it! I'll keep it as draft for now - @eujing says that Azure API can return a flag indicating whether a config change requires a reboot, but I have not figured that stuff out yet. I'll implement it later and push it to this PR. Once it's here, I'll undraft (?) a PR 😄

mlos_bench/mlos_bench/config/schemas/environments/remote/saas-env-subschema.json
```diff
@@ -0,0 +1,27 @@
...
+    "properties": {
+        "class": {
+            "enum": [
+                "mlos_bench.environments.remote.SaaSEnv",
+                "mlos_bench.environments.remote.saas_env.SaaSEnv"
+            ]
+        },
+        "config": {
+            "type": "object",
+            "$comment": "NOTE: there aren't currently any additional properties accepted by the OSEnv.  In the future we may want to use an os_type enum to help automatic OS param tuning by default without RemoteEnv requirements.",
```
bpkroth, line 17:
```suggestion
            "$comment": "NOTE: there aren't currently any additional properties accepted by the SaaSEnv.",
```

mlos_bench/mlos_bench/config/schemas/environments/remote/saas-env-subschema.json
```diff
@@ -0,0 +1,27 @@
...
+            "allOf": [
+                {
+                    "$ref": "../common-environment-subschemas.json#/$defs/common_environment_config"
+                }
+            ],
+            "unevaluatedProperties": false
+        }
+    },
+    "required": ["class", "config"]
```
bpkroth, line 26:
```suggestion
    "required": ["class"]
```

remove `config` if not actually required, else require at least one element.

mlos_bench/mlos_bench/config/schemas/environments/remote/saas-env-subschema.json
bpkroth, line 1: Wait, is the SaaSEnv itself directly invokable and do anything, or is it just a base class or interface that the Azure version actually implements?

If so, can we just exclude it from the schemas to avoid someone accidentally trying to invoke a class that doesn't do anything?

bpkroth: There's a spot in the test infra to skip the schema requirement for certain classes (e.g., base classes)

mlos_bench/mlos_bench/config/schemas/services/remote/azure/azure-saas-service-subschema.json
```diff
@@ -0,0 +1,73 @@
...
+                    "properties": {
+                        "subscription": {
+                            "description": "Azure subscription id (typically provided in the global config in order to omit from source control).",
+                            "type": "string"
+                        },
+                        "resourceGroup": {
+                            "description": "The name of the resource group to place the deployment in (typically provided in the global config in order to omit from source control).",
+                            "type": "string"
+                        },
+                        "vmName": {
```
bpkroth, line 31: Is `vmName` the right term to use for a FlexServer instance?  Maybe just `saas_instance_name` for flexibility?

(BTW, I've found that prefixing names so they're globally unique allows for more flexible overrides across services from the `globals` configs)

mlos_bench/mlos_bench/services/remote/azure/azure_saas.py
```diff
@@ -0,0 +1,279 @@
...
+from mlos_bench.environments.status import Status
+from mlos_bench.services.base_service import Service
+from mlos_bench.services.types.authenticator_type import SupportsAuth
+from mlos_bench.services.types.remote_config_type import SupportsRemoteConfig
+from mlos_bench.util import check_required_params, merge_parameters
+
+_LOG = logging.getLogger(__name__)
+
+
+class AzureSaaSConfigService(Service, SupportsRemoteConfig):
```
bpkroth, line 23:
```suggestion
class AzureSaaSConfigService(SaaSService, SupportsRemoteConfig):
```

?

mlos_bench/mlos_bench/config/schemas/services/remote/azure/azure-saas-service-subschema.json
bpkroth: Also, is this required?

mlos_bench/mlos_bench/services/remote/azure/azure_saas.py
```diff
@@ -0,0 +1,279 @@
...
+            subscription=self.config["subscription"],
+            resource_group=self.config["resourceGroup"],
+            provider=self.config["provider"],
+            vm_name="{vm_name}",
+            server_type="flexibleServers" if is_flex else "servers",
+            update="configurations",
+            api_version=api_version,
+        )
+
+        # These parameters can come from command line as strings, so conversion is needed.
```
bpkroth, line 118: I think I fixed that globally - maybe double check if it's still required?

bpkroth: #493 ?

mlos_bench/mlos_bench/services/remote/azure/azure_saas.py
```diff
@@ -0,0 +1,279 @@
...
+        )})
+
+    def _get_headers(self) -> dict:
+        """
+        Get the headers for the REST API calls.
+        """
+        assert self._parent is not None and isinstance(self._parent, SupportsAuth), \
+            "Authorization service not provided. Include service-auth.jsonc?"
+
+        return {"Authorization": "Bearer " + self._parent.get_access_token()}
```
bpkroth, line 182: Probably a future PR, but can we abstract that somehow for all AzureServices?  Or make it part of AzureAuth or something? It seems like we have some redundancy for all of our Azure interactions.

mlos_bench/mlos_bench/config/schemas/environments/remote/saas-env-subschema.json
bpkroth: Nm I misread this as a Service, not an environment.  *derp*

mlos_bench/mlos_bench/environments/remote/saas_env.py
```diff
@@ -0,0 +1,111 @@
...
+            self._params, self._tunable_params.get_param_values())
+        if not status.is_succeeded():
+            return False
+
+        (status, res) = self._config_service.is_config_pending_restart(self._params)
+        if not status.is_succeeded():
+            return False
+
+        if res['isConfigPendingRestart']:
+
```
bpkroth, line 97:
```suggestion
```

nit

mlos_bench/mlos_bench/services/remote/azure/azure_saas.py
bpkroth: Nm, I'd misread the env as a base service earlier.

mlos_bench/mlos_bench/services/types/remote_config_type.py
bpkroth, line 1: I think this.  Hopefully we can extend it to other systems in the future.

BTW, I added some quick notes about this in #404 and #498 about extending the phases for different environments to include distinctions between reload/restart/etc.  Might want to consider incorporating that into this more generally or as a separate PR.

mlos_bench/mlos_bench/services/remote/azure/azure_saas.py
motus: great idea - I'll do it in the next PR

mlos_bench/mlos_bench/services/types/remote_config_type.py
```diff
@@ -0,0 +1,55 @@
...
+        Returns
+        -------
+        result : (Status, dict={})
+            A pair of Status and result. The result is always {}.
+            Status is one of {PENDING, SUCCEEDED, FAILED}
+        """
+
+    def is_config_pending_restart(self, config: Dict[str, Any]) -> Tuple["Status", dict]:
+        """
+        Check if the configuration of an Azure DB service requires a restart.
```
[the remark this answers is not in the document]
motus: good point! will do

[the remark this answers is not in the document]
motus: good idea! I'll fix it now

mlos_bench/mlos_bench/services/remote/azure/azure_saas.py
motus: I think it can still be a string if we pass it as `--requestTimeout=...` in the command line. Anyway, I think I'll check and maybe fix it in the subsequent PRs because we still have that `float()` conversion, e.g., in `azure_services.py` and likely in other places as well

mlos_bench/mlos_bench/config/schemas/services/remote/azure/azure-saas-service-subschema.json
motus: yeah, I used `vmName` to make this environment a drop-in replacement for our stack of local+remote+local environments for MySQL.. We can rename it later. And yes, `vmName` *is* required - good catch, thank you!

motus: oops, never mind - `vmName` is _**not**_ required - we can pass it in `const_args` to the environment instead. (and that's another reason to keep the name for now)

bpkroth: Maybe it would be better to rename all of them to `instance_name` instead of `vmName` then.  Call it future work is fine, but conceptually I think it would be cleaner.

Gets tricky though if you want to be able to mix and match both of these.  For instance, client VM for running `benchbase` or some other workload and one or more SaaS instances that act as the server backend(s).

mlos_bench/mlos_bench/services/remote/azure/azure_saas.py
bpkroth: File an issue for now?

motus: done: #581

mlos_bench/mlos_bench/services/types/remote_config_type.py
```diff
@@ -0,0 +1,56 @@
...
+    def configure(self, config: Dict[str, Any],
+                  params: Dict[str, Any]) -> Tuple["Status", dict]:
+        """
+        Update the parameters of a SaaS service in the cloud.
+
+        Parameters
+        ----------
+        config : Dict[str, Any]
+            Key/value pairs of configuration parameters (e.g., vmName).
+        params : Dict[str, Any]
```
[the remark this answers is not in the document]
motus: Agree - I've opened an issue #582 for this. I don't want to do it in this PR because this change will touch many other methods, e.g., in `AzureVMService`

mlos_bench/mlos_bench/services/remote/azure/azure_saas.py
motus: Also, using this method here

P.S. We cannot have that assertion in the constructor, because it will fail our unit tests in `test_load_service_config_examples.py`

bpkroth (approved):
Still don't love the use of `vmName` in this, but I think we can take it up later.

PR #525, opened 2023-10-04
Client-facing storage API

motus:
Minimal implementation for the demo

### DIFF
--- a/mlos_bench/mlos_bench/storage/sql/experiment_data.py
+++ b/mlos_bench/mlos_bench/storage/sql/experiment_data.py
@@ -5,9 +5,9 @@
 """
 An interface to access the benchmark data stored in SQL DB.
 """
-
 from typing import Dict
 
+import pandas
 from sqlalchemy import Engine
 
 from mlos_bench.environments.status import Status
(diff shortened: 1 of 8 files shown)


mlos_bench/mlos_bench/storage/sql/experiment_data.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+
+    def __init__(self, engine: Engine, schema: DbSchema, exp_id: str):
+        super().__init__(exp_id)
+        self._engine = engine
+        self._schema = schema
+
+    @property
+    def trials(self) -> Dict[int, TrialData]:
+        with self._engine.connect() as conn:
```
bpkroth, line 32: Maybe not here, but we should fix this to use a connection pool to avoid repeated connections if possible.

mlos_bench/mlos_bench/storage/sql/trial_data.py
```diff
@@ -0,0 +1,114 @@
...
+                )
+            )
+            return pandas.DataFrame(
+                [(row.metric_id, row.metric_value) for row in cur_results.fetchall()],
+                columns=['metric', 'value'])
+
+    @property
+    def telemetry(self) -> pandas.DataFrame:
+        """
+        Retrieve the trials' telemetry from the storage.
```
bpkroth, line 82: Let's document the expected structure of the dataframes that are returned if possible.
Ideal would be type hints, but short of that at least in the docstring.

mlos_bench/mlos_bench/storage/storage_factory.py
```diff
@@ -0,0 +1,52 @@
+#
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+#
+"""
+Factory method to create a new Storage instance from configs.
```
bpkroth, line 6: Can this replace `_load_storage` in launcher?

It would help simplify some of the `_load` and `build_generic` code which is really only used for `Storage` and `Optimizer` classes.

mlos_bench/mlos_bench/storage/sql/experiment_data.py
motus: As far as I understand, SQLAlchemy uses a connection pool under teh hood - `engine.connect()` and `engine.begin()` just get a connection from that pool and wrap it in python context so it can be returned to the pool (and committed, in case of `.begin()`) on the context `__exit__()`

mlos_bench/mlos_bench/storage/sql/trial_data.py
motus: done. I've updated all docstrings where we return DataFrames

mlos_bench/mlos_bench/storage/storage_factory.py
motus: uh, that'd be tough. The Launcher ties together a lot of disjoint components, and it is hard to split it into more or less independent parts. Because of that, I had to duplicate some of its functionality in the storage factory ☹️ 
I'll think of a better design later, but for now let's just keep the storage factory as is for the demo and make sure it looks good on the user side

mlos_bench/mlos_bench/storage/sql/experiment_data.py
motus: Agree. I will add `experiment.all_trial_{results,telemetry}` in a moment

motus: added `experiment.results` that returns trial status, config, and results in one dataframe.

mlos_bench/mlos_bench/storage/storage_factory.py
```diff
@@ -0,0 +1,63 @@
...
+from mlos_bench.config.schemas import ConfigSchema
+from mlos_bench.services.config_persistence import ConfigPersistenceService
+from mlos_bench.tunables.tunable_groups import TunableGroups
+from mlos_bench.storage.base_storage import Storage
+
+
+def from_config(config_file: str,
+                global_configs: Optional[List[str]] = None,
+                tunables: Optional[List[str]] = None,
+                **kwargs: Any) -> Storage:
```
bpkroth, line 20: As mentioned elsewhere, this requires the user to specify two things that are probably already in the `--config` file they used to launch `mlos_bench` CLI.
In the future at least it'd be nice to be able to also instantiate a Storage reader from just the `--config` file (assuming it already references the `globals_storage_auth.jsonc` file).

bpkroth (approved):
Good for now.  Future improvements later.  Thanks!

mlos_bench/mlos_bench/storage/storage_factory.py
motus: yeah, I really wanted to do exactly that, but there are too many different components involved. Let's keep it as is for the demo, and we can always come back to it later.

PR #529, opened 2023-10-09
Misc bug fixes and improvements

bpkroth:
- Adds support for expanding `temp_dir` in LocalExec service from the globals.
- Also makes sure the CWD is on the config search path by default.
- Adds the ability to expand globals from Environment variables.
  - Mostly useful to allow things like `$PWD` to be expanded to a fully qualified path.
- Fix an errant debug message.

### DIFF
--- a/mlos_bench/mlos_bench/services/local/temp_dir_context.py
+++ b/mlos_bench/mlos_bench/services/local/temp_dir_context.py
@@ -8,6 +8,7 @@ Helper functions to work with temp files locally on the scheduler side.
 
 import abc
 import logging
+import os
 from contextlib import nullcontext
 from string import Template
 from tempfile import TemporaryDirectory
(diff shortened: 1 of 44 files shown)


mlos_bench/mlos_bench/services/local/temp_dir_context.py
```diff
@@ -62,6 +70,8 @@ def temp_dir_context(self, path: Optional[str] = None) -> Union[TemporaryDirecto
         temp_dir_context : TemporaryDirectory
             Temporary directory context to use in the `with` clause.
         """
-        if path is None and self._temp_dir is None:
+        temp_dir = path or self._temp_dir
+        if temp_dir is None:
             return TemporaryDirectory()
-        return nullcontext(path or self._temp_dir)
+        os.makedirs(temp_dir, exist_ok=True)
```
motus, line 76: I don't like that we create a new directory right inside the context - and never delete it. The original behavior was to either use some existing, externally provided path (and leave it to the client to clean up) or create a new (temp) directory and take the responsibility for cleaning it. OTOH, I understand that this is probably the most convenient place to create a temp_dir, and I am not sure if the class should clean it on the context's `__exit__()` - that can easily screw things up, and it will be very hard to debug. So maybe let's keep `os.makedirs()` here. End of rant. 😄

motus (approved):
Looks good! I am not sure I like `os.makedirs()` inside the temp dir context, but that's probably the most convenient place to have it, so let's keep it

mlos_bench/mlos_bench/services/local/temp_dir_context.py
bpkroth: My thought was that if someone specifies in their config a known location, they don't want it to be cleaned up.
It's kind of known not to be a "temp" directory then.
So, in that case, rather than error out when that known location doesn't already exist, we can "just" create it for them, but leave the rest to them to manage.

PR #533, opened 2023-10-10
Reverse order of preference for environment variable expansion.

bpkroth:
Follow on to #529 

Allows for environment variables to be used in the globals config file like so:

```json
"CONTAINER_WORKSPACE_FOLDER": "$CONTAINER_WORKSPACE_FOLDER",
"LOCAL_WORKSPACE_FOLDER": "$LOCAL_WORKSPACE_FOLDER",
```

So that the value can be pulled directly from the shell environment without having to invoke `mlos_bench` with those values explicitly like this:

~`mlos_bench --config ... --globals ... --CONTAINER_WORKSPACE_FOLDER=$CONTAINER_WORKSPACE_FOLDER --LOCAL_WORKSPACE_FOLDER=$LOCAL_WORKSPACE_FOLDER`~

### DIFF
--- a/mlos_bench/mlos_bench/launcher.py
+++ b/mlos_bench/mlos_bench/launcher.py
@@ -257,8 +257,8 @@ class Launcher:
             # new set of params explicitly.
             # Since this operates by updating the global_config along the way,
             # we need to continually update the set of params to use for substitution.
-            params = dict(os.environ)
-            params.update(self.global_config)
+            params = self.global_config.copy()
+            params.update(dict(os.environ))
             return Template(value).safe_substitute(params)
         if isinstance(value, dict):
             # Note: we use a loop instead of dict comprehension in order to
--- a/mlos_core/mlos_core/optimizers/__init__.py
+++ b/mlos_core/mlos_core/optimizers/__init__.py
@@ -65,7 +65,7 @@ class OptimizerFactory:
                optimizer_type: OptimizerType = DEFAULT_OPTIMIZER_TYPE,
                optimizer_kwargs: Optional[dict] = None,
                space_adapter_type: SpaceAdapterType = SpaceAdapterType.IDENTITY,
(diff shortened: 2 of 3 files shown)


mlos_bench/mlos_bench/launcher.py
```diff
@@ -257,8 +257,8 @@ def _expand_vars(self, value: Any) -> Any:
             # new set of params explicitly.
             # Since this operates by updating the global_config along the way,
             # we need to continually update the set of params to use for substitution.
-            params = dict(os.environ)
-            params.update(self.global_config)
+            params = self.global_config.copy()
+            params.update(dict(os.environ))
```
motus, line 261:
```suggestion
            params.update(os.environ)
```
or is it there to make mypy happy?

motus (approved):
makes sense!
approved + left a nitpick, disabled automerge

mlos_bench/mlos_bench/launcher.py
bpkroth: Yeah, it's a mypy thing.

`os.Environ` as a type of `EnvironVars` or something like that instead of a `dict`

mlos_core/mlos_core/optimizers/__init__.py
```diff
@@ -65,7 +65,7 @@ def create(*,
                optimizer_type: OptimizerType = DEFAULT_OPTIMIZER_TYPE,
                optimizer_kwargs: Optional[dict] = None,
                space_adapter_type: SpaceAdapterType = SpaceAdapterType.IDENTITY,
-               space_adapter_kwargs: Optional[dict] = None) -> ConcreteOptimizer:
+               space_adapter_kwargs: Optional[dict] = None) -> ConcreteOptimizer:   # type: ignore[type-var]
```
bpkroth, line 68: FIXME (future PR): Caused by a recent update to `mypy`.

bpkroth: See Also: #535

PR #536, opened 2023-10-10
Expand const args with globals

bpkroth:
Builds off of #533 

Allows for const_args of the form:

```json
"const_args": {
  "DB_DIR": "$workdir/db",
  "DB_BAK_DIR": "$workdir/db.bak",
}
```

Which can be expanded based on the value of `$workdir` from the `globals` config (which may in turn be expanded from environment variables per prior PRs).

### DIFF
--- a/mlos_bench/mlos_bench/environments/base_environment.py
+++ b/mlos_bench/mlos_bench/environments/base_environment.py
@@ -10,6 +10,7 @@ import abc
 import json
 import logging
 from datetime import datetime
+from string import Template
 from types import TracebackType
 from typing import Any, Dict, Iterable, List, Optional, Sequence, Tuple, Type, TYPE_CHECKING, Union
 from typing_extensions import Literal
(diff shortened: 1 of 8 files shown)


mlos_bench/mlos_bench/environments/base_environment.py
```diff
@@ -172,6 +174,19 @@ def _expand_groups(groups: Iterable[str],
...
+        """
+        Expand `$var` into actual values of the variables.
+        """
+        res: Dict[str, TunableValue] = {}
+        for key, val in params.items():
+            if isinstance(val, str):
+                res[key] = Template(val).safe_substitute(global_config)
+            else:
+                res[key] = val
+        return res
```
motus, line 188: maybe something like
```suggestion
        return {
            key: Template(val).safe_substitute(global_config) if isinstance(val, str) else val
            for key, val in params.items()
        }
```

mlos_bench/mlos_bench/launcher.py
```diff
@@ -257,8 +257,8 @@ def _expand_vars(self, value: Any) -> Any:
             # new set of params explicitly.
             # Since this operates by updating the global_config along the way,
             # we need to continually update the set of params to use for substitution.
-            params = dict(os.environ)
-            params.update(self.global_config)
+            params = self.global_config.copy()
+            params.update(dict(os.environ))
```
motus, line 261: just
```suggestion
            params.update(os.environ)
```
if mypy doesn't mind

motus (approved):
looks good! left a few minor comments

mlos_bench/mlos_bench/launcher.py
bpkroth: Answered in #533

PR #538, opened 2023-10-11
bugfix: allow non-numeric values in trial results data

motus:
without this fix, our Jupyter notebook for ADU SQLite optimization demo would fail


bpkroth (approved):
LGTM - tests in a future PR.

PR #548, opened 2023-10-19
Add unit tests for multi-score updates. Allow non-numeric metrics in storage type annotations.

### LINKED ISSUE #539: Add tests for non-numeric "metrics"
Looks ok for now.  I'll make an issue to add tests for this later.

_Originally posted by @bpkroth in https://github.com/microsoft/MLOS/pull/538#discussion_r1355298477_

motus:
This is for basic functionality only. We'll have more tests in the next PR where we test the the new `storage.experiments[]` API

Closes #539

### DIFF
--- a/mlos_bench/mlos_bench/storage/base_storage.py
+++ b/mlos_bench/mlos_bench/storage/base_storage.py
@@ -294,8 +294,8 @@ class Storage(metaclass=ABCMeta):
 
         @abstractmethod
         def update(self, status: Status, timestamp: datetime,
-                   metrics: Optional[Union[Dict[str, float], float]] = None
-                   ) -> Optional[Dict[str, float]]:
+                   metrics: Optional[Union[Dict[str, Any], float]] = None
+                   ) -> Optional[Dict[str, Any]]:
             """
             Update the storage with the results of the experiment.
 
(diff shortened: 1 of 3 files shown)


mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -294,8 +294,8 @@ def config(self, global_config: Optional[Dict[str, Any]] = None) -> Dict[str, An
 
         @abstractmethod
         def update(self, status: Status, timestamp: datetime,
-                   metrics: Optional[Union[Dict[str, float], float]] = None
-                   ) -> Optional[Dict[str, float]]:
+                   metrics: Optional[Union[Dict[str, Any], float]] = None
+                   ) -> Optional[Dict[str, Any]]:
```
bpkroth, line 298: nit:
```suggestion
                   ) -> Optional[Dict[str, TunableValue]]:
```
?

bpkroth: It's a DBMS, so should be only strings, ints, and floats, right?  Maybe also nulls ...

bpkroth (approved):

mlos_bench/mlos_bench/storage/base_storage.py
motus: good catch! fixing it now

motus: mmm, maybe not. ☹️ Turned out, using `TunableValue` type spreads out in all directions to the `Environment` and `Optimizer` classes. Let's keep it small for this PR and do the bigger type change later, because I am also not 100% sure that `TunableValue`, which is a type for *input* configuration parameters should be also used for the *output* metrics and telemetry. Maybe we should have a different type there, e.g., `MetricValue`, even it is identical to the `TunableValue`

Issue #530, opened 2023-10-09
Custom LocalExec service config not taking precedence


bpkroth:
- launcher.py creates a LocalExecService with a blank config:
  https://github.com/microsoft/MLOS/blob/e1120ae1565371d9203957ef61e2bb546b838e0c/mlos_bench/mlos_bench/launcher.py#L81
- this prevents environements which may specify their own `LocalExecService` config (e.g., with `temp_dir` and `abort_on_error` configs) from taking precedence)

motus:
Leaf environments should be able to override the services that they inherit from the parent (and the Launcher). We might need to have more unit tests to verify that - and make sure that the additional services loaded in the leaf environments don't override those in the parent or siblings

bpkroth:
As described elsewhere, there's a last writer wins problem with how we chain things.

motus:
Fixed in #551

PR #551, opened 2023-10-20
Fix the nested service inclusion issue + add unit tests

motus:
**UPDATE:** Fixed! The problem was with the `.register()` method invocation in the base class constructor, so the service can inherit methods from the parent service. That call overrode methods of the service object being created with references to the parent's methods. As a result, we were losing references to the local methods before we could register them. This PR is a workaround for the issue. We will think of a more elegant way of doing it later.

-------------------

Debugging and fixing the issue described in #530

**TODO:**
* [x] Fix the original issue from #530
* [x] Add unit tests to verify the problem
* [x] ~Fix the JSON schema to accept the `CompositeEnv` fixture from unit tests~
* [x] Resolve `mypy` issues with accessing `.temp_dir_context()`

### DIFF
--- a/mlos_bench/mlos_bench/services/base_fileshare.py
+++ b/mlos_bench/mlos_bench/services/base_fileshare.py
@@ -9,7 +9,7 @@ Base class for remote file shares.
 import logging
 
 from abc import ABCMeta, abstractmethod
-from typing import Any, Dict, Optional
+from typing import Any, Callable, Dict, List, Optional, Union
 
 from mlos_bench.services.base_service import Service
 from mlos_bench.services.types.fileshare_type import SupportsFileShareOps
@@ -24,7 +24,8 @@ class FileShareService(Service, SupportsFileShareOps, metaclass=ABCMeta):
 
     def __init__(self, config: Optional[Dict[str, Any]] = None,
                  global_config: Optional[Dict[str, Any]] = None,
-                 parent: Optional[Service] = None):
+                 parent: Optional[Service] = None,
+                 methods: Union[Dict[str, Callable], List[Callable], None] = None):
         """
         Create a new file share with a given config.
 
@@ -38,11 +39,13 @@ class FileShareService(Service, SupportsFileShareOps, metaclass=ABCMeta):
             Free-format dictionary of global parameters.
         parent : Service
             Parent service that can provide mixin functions.
+        methods : Union[Dict[str, Callable], List[Callable], None]
+            New methods to register with the service.
         """
-        # IMPORTANT: Save the local methods before invoking the base class constructor
-        local_methods = [self.upload, self.download]
-        super().__init__(config, global_config, parent)
-        self.register(local_methods)
+        super().__init__(
+            config, global_config, parent,
+            self.merge_methods(methods, [self.upload, self.download])
+        )
 
     @abstractmethod
     def download(self, params: dict, remote_path: str, local_path: str, recursive: bool = True) -> None:
--- a/mlos_bench/mlos_bench/services/base_fileshare.py
+++ b/mlos_bench/mlos_bench/services/base_fileshare.py
@@ -39,12 +39,10 @@ class FileShareService(Service, SupportsFileShareOps, metaclass=ABCMeta):
         parent : Service
             Parent service that can provide mixin functions.
         """
+        # IMPORTANT: Save the local methods before invoking the base class constructor
+        local_methods = [self.upload, self.download]
         super().__init__(config, global_config, parent)
-
-        self.register([
-            self.download,
-            self.upload,
-        ])
+        self.register(local_methods)
 
     @abstractmethod
     def download(self, params: dict, remote_path: str, local_path: str, recursive: bool = True) -> None:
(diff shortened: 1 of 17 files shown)


motus (commented):
Closes #530

mlos_bench/mlos_bench/services/base_fileshare.py
```diff
@@ -39,12 +39,10 @@ def __init__(self, config: Optional[Dict[str, Any]] = None,
         parent : Service
             Parent service that can provide mixin functions.
         """
+        # IMPORTANT: Save the local methods before invoking the base class constructor
+        local_methods = [self.upload, self.download]
```
bpkroth, line 43:
```suggestion
        self._local_methods = [self.upload, self.download]
```
?

Not sure that this is super useful, but maybe allows for a little bit more introspection and debugging help later.

bpkroth: Actually, if we did that, then we could do the following in the base class:

```python
def __init__(self, ...) -> None:
    # immediately check to see if we're instantiating the base class, or it's a subclass calling super().__init__()
    if self.__class__ == Service:
        self._local_methods = []
    else:
        # Child classes must populate self._local_methods prior to calling super().__init__()
        assert self._local_methods
    ... # other init work
    # At the end automatically register the methods for that child instance.
    self.register(self._local_methods)
```

Now we could drop the `self.register()` call from all the child classes since it should be taken care of in the super class, right?

mlos_bench/mlos_bench/services/local/local_exec.py
```diff
@@ -85,9 +85,11 @@ def __init__(self,
         parent : Service
             An optional parent service that can provide mixin functions.
         """
+        # IMPORTANT: Save the local methods before invoking the base class constructor
+        local_methods = [self.local_exec]
```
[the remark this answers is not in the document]
bpkroth: See above for a possible way to address this.

mlos_bench/mlos_bench/services/base_fileshare.py
bpkroth: Hmm, that might not work if there's mulitple inheritance and bases classes building up.

bpkroth: I left an in progress version of this adaptation here for now:

https://github.com/motus/MLOS/compare/sergiym/env/composite_service...bpkroth:MLOS:sergiym/env/composite_service

mlos_bench/mlos_bench/services/config_persistence.py
bpkroth, line 1: Looks like this one isn't following the `local_methods, super(), register()` pattern.

mlos_bench/mlos_bench/services/base_fileshare.py
motus: yeah, that's a great idea! now when I think about it, we can even do without `self._local_methods` at all and just pass that list into the superclass constructor. I'll play with it now

mlos_bench/mlos_bench/tests/services/remote/mock/mock_fileshare_service.py
```diff
@@ -24,12 +24,10 @@ class MockFileShareService(FileShareService, SupportsFileShareOps):
...
                  parent: Optional[Service] = None):
+        # IMPORTANT: Save the local methods before invoking the base class constructor
+        local_methods = [self.upload, self.download]
         super().__init__(config, global_config, parent)
-
-        self.register([
-            self.download,
-            self.upload,
-        ])
+        self.register(local_methods)
```
bpkroth, line 30: This pattern isn't followed in the `AzureFileshare` service.  Does it need to be?
The base `FileShareService` class does this, but I'm not sure off hand if it's registering the abstract methods, or the subclasses methods.  If it's already happening in the base class, can we drop it here?

mlos_bench/mlos_bench/services/base_fileshare.py
bpkroth: Actually, here's a full version:
https://github.com/motus/MLOS/pull/7

motus: oops, I've just pushed my own update a minute ago :) @bpkroth can you please take a look?

mlos_bench/mlos_bench/services/config_persistence.py
motus: oops, my fault. fixed!

mlos_bench/mlos_bench/tests/services/remote/mock/mock_fileshare_service.py
motus: I don't remember whether it was a bug or a feature :) I think it would be safer to enforce registering of the local methods in every class even if the superclass already does that

bpkroth (approved):
LGTM

mlos_bench/mlos_bench/services/base_fileshare.py
bpkroth: LGTM!  Thanks :)

PR #552, opened 2023-10-23
devcontainer Fixups

### LINKED ISSUE #550: Fixup container size
Something seems to have caused a large increase in our devcontainer size recently.  Should evaluate how to fix that.

bpkroth:
- reduce devcontainer size (but only a little - see comments below)
- improve cachability
- fixup rebuild times

Closes #550

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -34,7 +34,7 @@ RUN mkdir -p ${CONDA_PKGS_DIRS} \
     && chown -R vscode:conda ${CONDA_PKGS_DIRS} \
     && chmod -R u=rwX,g=rwXs,o=rX ${CONDA_PKGS_DIRS}
 
-USER vscode
+USER vscode:conda
 
 # Upgrade conda and use strict priorities
 # Use the mamba solver (necessary for some quality of life speedups due to
@@ -46,13 +46,16 @@ RUN umask 0002 \
     && /opt/conda/bin/conda list -n base \
     && /opt/conda/bin/conda install -v -y -n base conda-libmamba-solver \
     && /opt/conda/bin/conda config --set solver libmamba \
-    && /opt/conda/bin/conda list -n base
+    && /opt/conda/bin/conda list -n base \
+    && /opt/conda/bin/conda clean -v -y -a
 
 # Update the base. This helps save space by making sure the same version
 # python is used for both the base env and mlos env.
 RUN umask 0002 \
     && /opt/conda/bin/conda update -v -y -n base -c defaults --all \
-    && /opt/conda/bin/conda clean -v -y -a
+    && /opt/conda/bin/conda update -v -y -n base -c defaults conda python \
+    && /opt/conda/bin/conda clean -v -y -a \
+    && /opt/conda/bin/conda run -n base pip cache purge
 
 # Install some additional editor packages for the base environment.
 RUN umask 0002 \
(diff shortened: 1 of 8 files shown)


bpkroth:
This commit introduces an alternative way to structure the Dockerfile in order to allow the `mlos` conda env to share files with the `base` conda env (e.g., the core python version files)
https://github.com/microsoft/MLOS/pull/552/commits/8d815a6eb3eb2708ed5513ad187e4ad8dd58b131

It says ~400MB, making it a 6.4GB image instead of the 6.8GB image it is now.  To do that though, it creates fewer larger layers.

Unfortunately, the bulk of the space (4GB) is in our own `mlos` environment, which I think has value in pre-packaging.

I'm not totally convinced the change is worth it and makes it slightly slower to rebuild/debug since the base layer, which perhaps hasn't changed, isn't reuseable.

Anyways, stashing in the comments for now in case we want to revisit.

.github/workflows/devcontainer.yml
```diff
@@ -142,6 +139,7 @@ jobs:
       timeout-minutes: 3
       run: |
         set -x
+        git --no-pager diff --exit-code
```
bpkroth, line 142: There shouldn't be any local changes to the repo - if there are it can cause cache reuse problems.

.github/workflows/devcontainer.yml
```diff
@@ -90,17 +90,14 @@ jobs:
...
     - name: Fixup vscode uid/gid in the running container
       timeout-minutes: 3
       run: |
         set -x
         docker exec --user root mlos-devcontainer groupmod --non-unique -g `id -g` vscode
         docker exec --user root mlos-devcontainer usermod --non-unique -u `id -u` -g `id -g` vscode
         docker exec --user root mlos-devcontainer chown -R vscode:vscode /home/vscode
-        docker exec --user root mlos-devcontainer mkdir -p /opt/conda/pkgs/cache /var/cache/pip
-        docker exec --user root mlos-devcontainer chown -R vscode /opt/conda/pkgs/cache /var/cache/pip
-        docker exec --user root mlos-devcontainer chown -R vscode /opt/conda/envs/mlos
```
bpkroth, line 103: Minor time improvement - we don't need to walk that filesystem when we create the files with conda as the group owner.

.github/workflows/devcontainer.yml
```diff
@@ -90,17 +90,14 @@ jobs:
           --env LOCAL_WORKSPACE_FOLDER=$(pwd) \
           --env PYTEST_EXTRA_OPTIONS=$PYTEST_EXTRA_OPTIONS \
           --workdir /workspaces/MLOS \
-          --name mlos-devcontainer mlos-devcontainer sleep infinity
+          --name mlos-devcontainer mlos-devcontainer sleep 1800
```
bpkroth, line 93: The full suite of tests should never run more than 30 minutes I think, so we can let the devcontainer exit after that and throw an error.

.devcontainer/build/build-devcontainer.sh
```diff
@@ -41,7 +41,7 @@ if [ "${NO_CACHE:-}" == 'true' ]; then
     devcontainer_build_args='--no-cache'
 else
     cache_from='mloscore.azurecr.io/mlos-devcontainer:latest'
-    devcontainer_build_args="--cache-from $cache_from"
+    devcontainer_build_args="--cache-from $cache_from --cache-from mlos-devcontainer:latest"
```
bpkroth, line 44: Allow reusing from the local cache.

.devcontainer/Dockerfile
```diff
@@ -74,12 +77,16 @@ RUN /tmp/conda-tmp/prep-deps-files.sh \
...
 FROM base AS devcontainer
+USER vscode
 COPY --from=deps-prep --chown=vscode:conda /tmp/conda-tmp/mlos_deps.yml /tmp/conda-tmp/combined.requirements.txt /tmp/conda-tmp/
 RUN umask 0002 \
-    && /opt/conda/bin/conda env create -n mlos -v -f /tmp/conda-tmp/mlos_deps.yml \
-    && /opt/conda/bin/conda run -n mlos pip install -U -r /tmp/conda-tmp/combined.requirements.txt \
-    && /opt/conda/bin/conda clean -v -y -a \
+    && sg conda -c "/opt/conda/bin/conda env create -n mlos -v -f /tmp/conda-tmp/mlos_deps.yml" \
+    && sg conda -c "/opt/conda/bin/conda run -n mlos pip install --no-cache-dir -U -r /tmp/conda-tmp/combined.requirements.txt" \
+    && sg conda -c "/opt/conda/bin/conda run -n mlos pip cache purge" \
```
bpkroth, line 88: Cull a little more container image space by removing the pip cache.

motus (approved):

PR #555, opened 2023-10-26
Add asyncio event loop context manager

bpkroth:
Extracting out from another refactor of SSH support PR #510


motus (approved):
very neat!

PR #567, opened 2023-10-27
Attempt to convert results to numeric types

bpkroth:
Split out from #557 

- [ ] Add test case

PR #569, opened 2023-10-27
Tests for converting results to numerics

bpkroth:
Adds tests to #567


mlos_bench/mlos_bench/util.py
```diff
@@ -247,3 +247,33 @@ def get_git_info(path: str = __file__) -> Tuple[str, str, str]:
...
+    if val is None:
+        return val
+    try:
+        if "." in val:
+            return float(val)
+        else:
+            return int(val)
+    except ValueError:
+        pass
+    return str(val)
```
motus, line 279: this will break on strings like `1e6` or, e.g., `user.name`. maybe do something like this?
```suggestion
    if val is None:
        return val
    try:
        val_float = float(val)
        try:
            val_int = int(val)
            return val_int if val_int == val_float else val_float
        except ValueError:
            return val_float
    except ValueError:
        return str(val)
```

motus: also, maybe test for such strings, too? I mean, `1e6` and `output.log`

bpkroth: `user.name` I think should be fine, but good point about additional floating notation.  I'll have it just try the cast and see how it goes.

bpkroth: used your suggestion and added explicit tests for it

mlos_bench/mlos_bench/tests/util_try_parse_test.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def test_try_parse_val() -> None:
+    """
+    Check that we can retrieve git info about the current repository correctly.
+    """
+    assert try_parse_val(None) is None
+    assert try_parse_val("1") == int(1)
+    assert try_parse_val("1.1") == float(1.1)
+    assert try_parse_val("1e6") == float(1e6)
```
motus, line 19:
```suggestion
    assert try_parse_val("1e6") == float(1e6)
    assert try_parse_val("NaN") == float("nan")
    assert try_parse_val("inf") == float("Inf")
```

motus (approved):
Looks great! Added a few more asserts

motus (approved):

PR #557, opened 2023-10-26
Add RemoteEnv tests using SshService testing infrastructure

bpkroth:
Adds tests for RemoteEnv `setup`, `run`, `teardown` using SshServices

Fixes some associated bugs:
- convert certain `SFTPError` exceptions to `FileNotFound` in `SshFileShareService`
  This is important for `LocalFileShareEnv` integration especially since it always calls `download` during calls to `status`, but only handles `FileNotFound` exceptions when `ignore_missing=True`
- Fixups to #517 for `download`, `upload` to connect via `self._params` (loaded from the `required_args`) instead of `self.config`

Also included:
- basic configs for SshServices and LocalExec for easy inclusion

Closes #521

PR #573, opened 2023-11-01
Do not pass all params as script env variables by default

### LINKED ISSUE #564: Don't expose all params as shell environment variables by default
_Originally posted by @bpkroth in https://github.com/microsoft/MLOS/pull/557#discussion_r1374921396_

motus:
Closes #564


bpkroth (commented):
Let's add tests for this after #557 is merged.

In particular, I think we can probably extend something in `test_ssh_env.jsonc` to make sure that nothing gets added, or else make a separate test.

bpkroth (approved):
Looks good, let's add tests though.

PR #574, opened 2023-11-01
Update json schema to the actual parameters of script_env

motus:
Remove unused parameters like `shell_env_params_match` and `shell_env_params_mapping`; add `shell_env_params_rename`


bpkroth (approved):
Real config test case we can load?

PR #579, opened 2023-11-06
Clean-up temp files created by mlos_core optimizers

motus:


mlos_bench/mlos_bench/config/optimizers/mlos_core_smac.jsonc
```diff
@@ -8,7 +8,6 @@
         "optimization_target": "score",
         "optimization_direction": "min",
         "max_iterations": 100,
-        "optimizer_type": "SMAC",
-        "output_directory": "smac_output"
```
bpkroth, line 12: I think this was important to have for default setups since that's where it stores history and things like that that smac needs.  We might need to remove it only for testing environments, or place those executions inside a tempdir context.

bpkroth: Looks like you mostly did that.

bpkroth: Might just need to verify that we can reprime the optimizer without that directory around.  Not sure if we have a test for that yet ...

bpkroth (approved):
Looks pretty good.  I think there's some lint issues.
I'll file an issue for the test case I mentioned and we can take it up in a future PR.

mlos_bench/mlos_bench/config/optimizers/mlos_core_smac.jsonc
motus: makes sense! I'll put an explicit `null` there so it can be overridden by the value from the global config. I don't want to have the explicit `"smac_output"` value in it because then our `launcher_run_test.py` would create that directory on every run.

PR #586, opened 2023-11-07
allow using all parameters in paths witout declaring them in `shell_env_params`

### LINKED ISSUE #585: Make all parameters visible as `$vars` in paths and other applications that does not involve the OS shell
When constructing paths for file share upload and download, we require that all `$vars` in such paths should be explicitly exposed via `shell_env_params`. This makes sense for local and remote shell commands, as it prevents the leakage of (potentially sensitive) parameters to the shell but is redundant and too restrictive for the path names. We should allow using all parameters in such paths without any explicit declarations.

motus:
Closes #585


mlos_bench/mlos_bench/environments/local/local_fileshare_env.py
```diff
@@ -136,7 +136,7 @@ def _download_files(self, ignore_missing: bool = False) -> None:
             If False, proceed with downloading other files and log a warning.
         """
         assert self._temp_dir is not None
-        params = self._get_env_params()
+        params = self._get_env_params(restrict=False)
```
bpkroth, line 139: This seems weird to me.  `upload` and `download` are not shell commands.
```suggestion
        params = self._params.copy()
```
Why not change this to just `self._params`.  Won't that accomplish the same thing?

Or is the underscore replacement important?

mlos_bench/mlos_bench/environments/local/local_fileshare_env.py
bpkroth, line 1: Tests?

mlos_bench/mlos_bench/environments/local/local_fileshare_env.py
motus: I agree that upload/download are not shell commands, but I think that the behavior of all dollar variables should be consistent across all uses - hence the `$PWD`, for example.

Besides, we might want to use at least the `shell_env_params_rename` feature and convert all parameter values to strings, otherwise `string.Template` will break on things like `trial_id`.

mlos_bench/mlos_bench/environments/local/local_fileshare_env.py
motus: working on tests now - will send a PR soon!

bpkroth (approved):
tests ...

mlos_bench/mlos_bench/environments/local/local_fileshare_env.py
bpkroth: Hmm, I've not had issues with substitutions for `trial_id` or other underscore variables before.
Dashed variables I know don't work though.

mlos_bench/mlos_bench/environments/script_env.py
```diff
@@ -78,17 +78,25 @@ def __init__(self,
...
+
         Returns
         -------
         env_params : Dict[str, str]
             Parameters to pass as *shell* environment variables into the script.
             This is usually a subset of `_params` with some possible conversions.
         """
-        rename = {self._RE_INVALID.sub("_", key): key for key in self._shell_env_params}
+        input_params = self._shell_env_params if restrict else self._params.keys()
+        rename = {self._RE_INVALID.sub("_", key): key for key in input_params}
```
bpkroth, line 99: A random thought on the `trial_id` vs `trial-id` issue mentioned above.
In this current code, we'd replace `trial-id` with `trial_id` and might have a conflict with the original `trial_id`.
I don't know how often that'd show up, but might be a good idea to issue a message about it at least.

Issue #568, opened 2023-10-27
Provide convenience method to inherit required_args from parent?


bpkroth:
_Originally posted by @bpkroth in https://github.com/microsoft/MLOS/pull/557#discussion_r1374921892_
            
Right now in CompositeEnvs we can `include_services` that are used by several `children` `Environments` but each of those must specify the same set of `required_args` in order to get the `params` necessary for the `Service` in each `Environment`

This is repetitive and doesn't make it very reusable.

We should try and fix this.

PR #593, opened 2023-11-14
Make FIO benchmark a configurable parameter

### LINKED ISSUE #500: Enhancement: make fio config accept and upload an alternative fio job config file.
- [ ] Future version could also make the fio config file a variable and allow uploading it.
Let's make that a future PR though.

_Originally posted by @bpkroth in https://github.com/microsoft/MLOS/pull/494#discussion_r1309030465_

- [ ] Similarly, minor renaming of the results to something `fio.results.csv` (mostly cosmetic)

motus:
Also, fix the shell env variables declarations and rename the output files for readability

Closes #500

### DIFF
--- a/mlos_bench/mlos_bench/config/environments/apps/fio/fio.jsonc
+++ b/mlos_bench/mlos_bench/config/environments/apps/fio/fio.jsonc
@@ -5,7 +5,8 @@
     "config": {
 
         "const_args": {
-            "fio_prefix": "fio"
+            "fio_prefix": "fio",
+            "fio_benchmark": "readwrite.fio.ini"
         },
 
         "children": [
@@ -14,6 +15,10 @@
                 "class": "mlos_bench.environments.local.LocalFileShareEnv",
                 "config": {
                     "required_args": [
+                        "experiment_id",
+                        "trial_id"
+                    ],
+                    "shell_env_params": [
                         "experiment_id",
                         "trial_id"
                     ],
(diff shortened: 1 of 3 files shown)


mlos_bench/mlos_bench/config/environments/apps/fio/fio.jsonc
```diff
@@ -17,6 +18,10 @@
                         "experiment_id",
                         "trial_id"
                     ],
+                    "shell_env_params": [
+                        "experiment_id",
+                        "trial_id"
```
bpkroth, line 23: Should we just expose these by default?  They're very often necessary/useful.

bpkroth: Separate PR I think ...

mlos_bench/mlos_bench/config/environments/apps/fio/fio.jsonc
```diff
@@ -60,16 +74,21 @@
                         "trial_id",
                         "fio_prefix"
                     ],
+                    "shell_env_params": [
+                        "experiment_id",
+                        "trial_id",
+                        "fio_prefix"
```
bpkroth, line 80: As mentioned in #568, this gets a little cumbersome I think at times to have to declare a parameter as necessary or exposed throughout the full stack of the configs.
Also a conversation for a separate PR, but would be nice to have a way to say, all the children in this CompositeEnv (maybe just this one and not recursively, though I'm not sure about that), should have this arg required or this shell_env exposed.

bpkroth: Then it could happen at the top of this stack and not have to be repeated after that.

mlos_bench/mlos_bench/config/environments/apps/fio/fio.jsonc
motus: I don't know - in my opinion, "explicit is better than implicit". I don't mind an extra "shell_env_params" clause, and it kind of makes it obvious that all other dollar variables should be defined there as well

bpkroth: I can see that side of the argument too.  It just gets very verbose after a while, which I don't love either.

bpkroth: The other proposal I made below could be another way of solving this, btw.

Basically, we declare trial_id as a required_arg or shell_env at the top of the CompositeEnv, and let it flow through to all others after that somehow.
Doesn't work well with the way that schema checks work atm, I think though.

bpkroth (approved):
LGTM.  Side discussions about future ergonomic improvements can be taken up later.

mlos_bench/mlos_bench/config/environments/apps/fio/fio.jsonc
motus: agree. I'll think of an elegant way to implement it

PR #601, opened 2023-11-17
experiment_id documentation improvements

### LINKED ISSUE #599: Document the importance of `experiment_id` parameter.
We need to add docs on why this parameter and what it controls (e.g., reloading trial data, etc.) and how changes to scripts, tunable params, configs, etc. can affect reusability of trial data in ways that we can't always detect (e.g., code, version, etc. lineage).

bpkroth:
Closes #599

### DIFF
--- a/mlos_bench/mlos_bench/launcher.py
+++ b/mlos_bench/mlos_bench/launcher.py
@@ -201,7 +201,7 @@ class Launcher:
             help='Disable teardown of the environment after the benchmark.')
 
         parser.add_argument(
-            '--experiment_id', '--experiment_id', required=False, default=None,
+            '--experiment_id', '--experiment-id', required=False, default=None,
             help=textwrap.dedent("""
                 Experiment ID to use for the benchmark.
                 If omitted, the value from the --cli config or --globals is used.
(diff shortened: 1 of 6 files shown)


mlos_bench/mlos_bench/launcher.py
```diff
@@ -50,7 +51,15 @@ class Launcher:
 
     def __init__(self, description: str, long_text: str = "", argv: Optional[List[str]] = None):
         _LOG.info("Launch: %s", description)
-        parser = argparse.ArgumentParser(description=f"{description} : {long_text}")
+        epilog = textwrap.dedent("""
```
motus, line 54:
```suggestion
        epilog = """
```
I don't think we ever need to call `textwrap.dedent()` for argparser help strings. it should automatically strip and dedent the text

motus (approved):
Looks good - but we can probably do without using `textwrap`

mlos_bench/mlos_bench/launcher.py
bpkroth: Ahh, I didn't realize that `argparser` did that too.  That might be why the newlines aren't preserved here.

Anyways, removed for now.  Can try and reprettify that text output some other time.

PR #603, opened 2023-11-20
Explicitly add the project root to pytest arguments

motus:
My VSCode adds `"."` to pytest arguments in config every time I rebuild the container - and deletes the `--log-level=DEBUG` argument at teh same time. With this small fix, vscode can accept the project config as we have it in git without modifying it


bpkroth (approved):
Hmm, we used to have this.  Must have lost it in some of my previous rash of fixups to pytest.

PR #608, opened 2023-11-22
Allow specifying services in CLI args or top level cli config.

### LINKED ISSUE #607: Provide a way to setup top-level services in configs for easier reuse
Right now environments specify which services they need.

That means that if you want to change how an environment is created (e.g., on AWS (if that existed) vs. Azure), you'd need to edit the environment config to link to the new service provider.

That limits reuse of environment configs.

Alternatively, we could specify the services in the cli using something like `--service some/service-list.mlos.jsonc`, and include that in the `--config cli/config.mlos.jsonc`.

Note: this method still has limitations, especially if you're mixing and matching remote_exec methods (e.g., Azure VM remote_exec for one VM and SSH for another), but I think that use case is probably less common and can handle it in the current way - copying/editing the environment configs.

bpkroth:
Closes #607

### DIFF
--- a/mlos_bench/mlos_bench/config/schemas/cli/cli-schema.json
+++ b/mlos_bench/mlos_bench/config/schemas/cli/cli-schema.json
@@ -47,6 +47,16 @@
             "uniqueItems": true
         },
 
+        "services": {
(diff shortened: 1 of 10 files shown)


motus (approved):
Looks good!

PR #606, opened 2023-11-20
Add NetworkEnv and AzureNetworkService

bpkroth:
Adds support for modeling virtual networks as an Environment in order to deploy virtual networks separately from VMs.

This can be useful if we want to, for instance, leave and reuse virtual networks accessible across different experiments.

To do this, we also move common code from `azure_vm_services` to a base class in `azure_services`.

- [x] `NetworkEnv` implementation
- [x] `NetworkEnv` schema
- [ ] `NetworkEnv` tests (kinda hard to do with the current test infra, may skip for now)
- [x] `AzureNetworkService` implementation
- [x] `AzureNetworkService` schema
- [x] `AzureNetworkService` tests
- [x] Add example config

PR #614, opened 2023-11-29
Add support for recursive dictionary templating to more places

bpkroth:
Fixes extracted from #606 

This expands support for doing secondary `$var` string template expansion to two additional places beyond `global_config`:

- `const_args`
- `deploymentTemplateParameters` in `AzureVmService` (and soon `AzureNetworkService`)

Expands a few test cases to check for this and relies on one prior existing one to still pass (the `launch_parse_args` tests which already checked for this behavior in `global_config` expansion).

### DIFF
--- a/mlos_bench/mlos_bench/dict_templater.py
+++ b/mlos_bench/mlos_bench/dict_templater.py
@@ -51,12 +51,11 @@ class DictTemplater:    # pylint: disable=too-few-public-methods
             The expanded dictionary.
         """
         self._dict = deepcopy(self._template_dict)
-        extra_source_dict = {} if extra_source_dict is None else extra_source_dict
         self._dict = self._expand_vars(self._dict, extra_source_dict, use_os_env)
         assert isinstance(self._dict, dict)
         return self._dict
 
-    def _expand_vars(self, value: Any, extra_source_dict: Dict[str, Any], use_os_env: bool) -> Any:
+    def _expand_vars(self, value: Any, extra_source_dict: Optional[Dict[str, Any]], use_os_env: bool) -> Any:
         """
         Recursively expand $var strings in the currently operating dictionary.
         """
@@ -69,12 +68,13 @@ class DictTemplater:    # pylint: disable=too-few-public-methods
             # Finally, fallback to the os environment.
             if use_os_env:
                 value = Template(value).safe_substitute(dict(os.environ))
-            return value
-        if isinstance(value, dict):
+        elif isinstance(value, dict):
             # Note: we use a loop instead of dict comprehension in order to
             # allow secondary expansion of subsequent values immediately.
             for (key, val) in value.items():
                 value[key] = self._expand_vars(val, extra_source_dict, use_os_env)
-        if isinstance(value, list):
-            return [self._expand_vars(val, extra_source_dict, use_os_env) for val in value]
+        elif isinstance(value, list):
+            value = [self._expand_vars(val, extra_source_dict, use_os_env) for val in value]
+        else:
+            raise ValueError(f"Unexpected type {type(value)} for value {value}")
         return value
--- a/mlos_bench/mlos_bench/dict_templater.py
+++ b/mlos_bench/mlos_bench/dict_templater.py
@@ -30,8 +30,6 @@ class DictTemplater:    # pylint: disable=too-few-public-methods
         self._template_dict = source_dict.copy()
         # The source/target dictionary to expand.
         self._dict: Dict[str, Any] = {}
-        # An additional source dictionary to use for expansion.
-        self._extra_source_dict: Dict[str, Any] = {}
 
     def expand_vars(self, *,
                     extra_source_dict: Optional[Dict[str, Any]] = None,
--- a/mlos_bench/mlos_bench/dict_templater.py
+++ b/mlos_bench/mlos_bench/dict_templater.py
@@ -6,6 +6,7 @@
 Simple class to help with nested dictionary $var templating.
 """
 
+from copy import deepcopy
 from string import Template
 from typing import Any, Dict, Optional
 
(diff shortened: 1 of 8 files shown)


mlos_bench/mlos_bench/dict_templater.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        Initialize the templater.
+
+        Parameters
+        ----------
+        source_dict : Dict[str, Any]
+            The template dict to use for source variables.
+        """
+        # A copy of the initial data structure we were given with templates intact.
+        self._template_dict = source_dict.copy()
```
bpkroth, line 30: Probably need deepcopies

mlos_bench/mlos_bench/dict_templater.py
bpkroth, line 1: I somewhat lazily didn't add explicit tests for this, and instead have tests just for the use cases of them.

Can add explicit tests for this as well if desired.

mlos_bench/mlos_bench/dict_templater.py
```diff
@@ -0,0 +1,80 @@
...
+            Whether to use the os environment variables a final fallback for expansion.
+
+        Returns
+        -------
+        Dict[str, Any]
+            The expanded dictionary.
+        """
+        self._dict = deepcopy(self._template_dict)
+        extra_source_dict = {} if extra_source_dict is None else extra_source_dict
+        self._dict = self._expand_vars(self._dict, extra_source_dict, use_os_env)
```
motus, line 55: a haskeller in me suggests
```suggestion
        self._dict = self._expand_vars(deepcopy(self._template_dict), extra_source_dict, use_os_env)
```
and make `extra_source_dict` parameter `Optional` in `_expand_vars()`

mlos_bench/mlos_bench/environments/base_environment.py
```diff
@@ -201,10 +201,7 @@ def _expand_vars(params: Dict[str, TunableValue], global_config: Dict[str, Tunab
         """
         Expand `$var` into actual values of the variables.
         """
-        return {
-            key: Template(val).safe_substitute(global_config) if isinstance(val, str) else val
-            for key, val in params.items()
-        }
+        return DictTemplater(params).expand_vars(extra_source_dict=global_config)
```
bpkroth, line 204: Left in a simple staticmethod due to existing test infra usage.

mlos_bench/mlos_bench/dict_templater.py
```diff
@@ -0,0 +1,80 @@
...
+            if use_os_env:
+                value = Template(value).safe_substitute(dict(os.environ))
+            return value
+        if isinstance(value, dict):
+            # Note: we use a loop instead of dict comprehension in order to
+            # allow secondary expansion of subsequent values immediately.
+            for (key, val) in value.items():
+                value[key] = self._expand_vars(val, extra_source_dict, use_os_env)
+        if isinstance(value, list):
+            return [self._expand_vars(val, extra_source_dict, use_os_env) for val in value]
```
motus, line 79:
```suggestion
        elif isinstance(value, dict):
            # Note: we use a loop instead of dict comprehension in order to
            # allow secondary expansion of subsequent values immediately.
            for (key, val) in value.items():
                value[key] = self._expand_vars(val, extra_source_dict, use_os_env)
        elif isinstance(value, list):
            value = [self._expand_vars(val, extra_source_dict, use_os_env) for val in value]
        else:
            raise ValueError(f"Unexpected type: {type(value)}")
```
so that we have a single `return` at the end (good got logging/debugging/etc.)

mlos_bench/mlos_bench/dict_templater.py
```diff
@@ -0,0 +1,80 @@
...
+        ----------
+        source_dict : Dict[str, Any]
+            The template dict to use for source variables.
+        """
+        # A copy of the initial data structure we were given with templates intact.
+        self._template_dict = deepcopy(source_dict)
+        # The source/target dictionary to expand.
+        self._dict: Dict[str, Any] = {}
+
+    def expand_vars(self, *,
```
motus, line 35: we have only two parameters of different types, so maybe there is no need to enforce explicit kwargs
```suggestion
    def expand_vars(self,
```
I'm ambivalent on this one, though

mlos_bench/mlos_bench/services/remote/azure/azure_vm_services.py
```diff
@@ -191,8 +192,11 @@ def __init__(self,
         assert template is not None and isinstance(template, dict)
         self._deploy_template = template
 
-        self._deploy_params = merge_parameters(
-            dest=self.config['deploymentTemplateParameters'].copy(), source=global_config)
+        # Allow for recursive variable expansion as we do with global params and const_args.
+        deploy_params = DictTemplater(self.config['deploymentTemplateParameters'].copy()).expand_vars(
+            extra_source_dict=global_config or {})
```
motus, line 197: I don't think there is a need to copy, so
```suggestion
        deploy_params = DictTemplater(self.config['deploymentTemplateParameters']).expand_vars(
            extra_source_dict=global_config)
```

mlos_bench/mlos_bench/dict_templater.py
bpkroth: Leaving that one as is for my own readability for now.

mlos_bench/mlos_bench/environments/base_environment.py
motus: makes perfect sense

mlos_bench/mlos_bench/dict_templater.py
motus: I am fine with this PR as it is. We can always add more tests later

motus (approved):

mlos_bench/mlos_bench/dict_templater.py
bpkroth: Added them.  They caught a bug with one of the suggestions :)

PR #617, opened 2023-12-01
Support for special tunable values outside of the range

motus:
Enables special values outside of the range (e.g., `-1` with a range of `[0, 100]`).

To do we make use of "conditionals" in ConfigSpace to constrain the space.  This has a number of implementation implications, addressed below:

* [x] Add support for special values to the `Tunable` class
* [x] Add unit tests for assigning special values outside of the range to the `Tunable` objects
* [x] Add special values outside of the range to the unit tests for `ConfigSpace` conversion
* [x] Implement proper `TunableGroups` to `ConfigSpace` conversion for tunables with special values
* [x] Update `mlos_core` optimizers to support conditionals and special values in `ConfigSpace`
* [x] Add more unit tests to check the conversion
* [x] Make LlamaTune adapter support conditionals in `ConfigSpace`

### DIFF
--- a/mlos_bench/mlos_bench/optimizers/convert_configspace.py
+++ b/mlos_bench/mlos_bench/optimizers/convert_configspace.py
@@ -148,3 +148,11 @@ def tunable_values_to_configuration(tunables: TunableGroups) -> Configuration:
             values[tunable.name] = tunable.value
     configspace = tunable_groups_to_configspace(tunables)
     return Configuration(configspace, values=values)
+
+
+def configspace_data_to_tunable_values(data: dict) -> dict:
+    """
+    Remove the fields that correspond to special values in ConfigSpace.
+    In particular, remove `__type:*` keys and trim `special:` prefixes.
+    """
+    return {k.split(":", 1)[0]: v for (k, v) in data.items() if not k.startswith("__")}
--- a/mlos_bench/mlos_bench/optimizers/convert_configspace.py
+++ b/mlos_bench/mlos_bench/optimizers/convert_configspace.py
@@ -24,6 +24,16 @@ from mlos_bench.tunables.tunable_groups import TunableGroups
 _LOG = logging.getLogger(__name__)
 
 
+class TunableValueKind:
+    """
+    Enum for the kind of the tunable value (special or not).
+    It is not a true enum because ConfigSpace wants string values.
+    """
+
+    SPECIAL = "special"
+    RANGE = "range"
+
+
 def _tunable_to_configspace(
         tunable: Tunable, group_name: Optional[str] = None, cost: int = 0) -> ConfigurationSpace:
     """
@@ -82,11 +92,13 @@ def _tunable_to_configspace(
             default_value=tunable.default if tunable.default in tunable.special else None,
             meta=meta),
         type_name: CategoricalHyperparameter(
-            name=type_name, choices=["special", "range"], default_value="special",
+            name=type_name,
+            choices=[TunableValueKind.SPECIAL, TunableValueKind.RANGE],
+            default_value=TunableValueKind.SPECIAL,
             weights=[0.5, 0.5]),  # TODO: Make weights configurable; FLAML requires uniform weights.
     })
-    cs.add_condition(EqualsCondition(cs[special_name], cs[type_name], "special"))
-    cs.add_condition(EqualsCondition(cs[tunable.name], cs[type_name], "range"))
+    cs.add_condition(EqualsCondition(cs[special_name], cs[type_name], TunableValueKind.SPECIAL))
+    cs.add_condition(EqualsCondition(cs[tunable.name], cs[type_name], TunableValueKind.RANGE))
 
     return cs
 
@@ -136,10 +148,10 @@ def tunable_values_to_configuration(tunables: TunableGroups) -> Configuration:
         if tunable.special:
             (special_name, type_name) = special_param_names(tunable.name)
             if tunable.value in tunable.special:
-                values[type_name] = "special"
+                values[type_name] = TunableValueKind.SPECIAL
                 values[special_name] = tunable.value
             else:
-                values[type_name] = "range"
+                values[type_name] = TunableValueKind.RANGE
                 values[tunable.name] = tunable.value
         else:
             values[tunable.name] = tunable.value
@@ -159,7 +171,7 @@ def configspace_data_to_tunable_values(data: dict) -> dict:
     ]
     for k in specials:
         (special_name, type_name) = special_param_names(k)
-        if data[type_name] == "special":
+        if data[type_name] == TunableValueKind.SPECIAL:
             data[k] = data[special_name]
         if special_name in data:
             del data[special_name]
--- a/mlos_bench/mlos_bench/optimizers/convert_configspace.py
+++ b/mlos_bench/mlos_bench/optimizers/convert_configspace.py
@@ -8,24 +8,27 @@ Functions to convert TunableGroups to ConfigSpace for use with the mlos_core opt
 
 import logging
 
-from typing import Optional
+from typing import Dict, Optional
 
-from ConfigSpace.hyperparameters import Hyperparameter
-from ConfigSpace import UniformIntegerHyperparameter
-from ConfigSpace import UniformFloatHyperparameter
-from ConfigSpace import CategoricalHyperparameter
-from ConfigSpace import ConfigurationSpace, Configuration
-
-from mlos_bench.tunables.tunable import Tunable
+from ConfigSpace import (
+    Configuration,
+    ConfigurationSpace,
+    CategoricalHyperparameter,
+    UniformIntegerHyperparameter,
+    UniformFloatHyperparameter,
+    EqualsCondition,
+)
+from mlos_bench.tunables.tunable import Tunable, TunableValue
 from mlos_bench.tunables.tunable_groups import TunableGroups
 
 _LOG = logging.getLogger(__name__)
 
 
-def _tunable_to_hyperparameter(
-        tunable: Tunable, group_name: Optional[str] = None, cost: int = 0) -> Hyperparameter:
+def _tunable_to_configspace(
+        tunable: Tunable, group_name: Optional[str] = None, cost: int = 0) -> ConfigurationSpace:
     """
-    Convert a single Tunable to an equivalent ConfigSpace Hyperparameter object.
+    Convert a single Tunable to an equivalent ConfigSpace Hyperparameter objects,
+    wrapped in a ConfigurationSpace for composability.
 
     Parameters
     ----------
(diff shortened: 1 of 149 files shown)


mlos_bench/mlos_bench/optimizers/convert_configspace.py
```diff
@@ -8,24 +8,27 @@
...
-from mlos_bench.tunables.tunable import Tunable
+from typing import Dict, Optional, Tuple
+
+from ConfigSpace import (
+    Configuration,
+    ConfigurationSpace,
+    CategoricalHyperparameter,
+    UniformIntegerHyperparameter,
+    UniformFloatHyperparameter,
+    EqualsCondition,
```
bpkroth, line 19: nit: sort these?

mlos_bench/mlos_bench/optimizers/convert_configspace.py
```diff
@@ -8,24 +8,27 @@
...
 _LOG = logging.getLogger(__name__)
 
 
-def _tunable_to_hyperparameter(
-        tunable: Tunable, group_name: Optional[str] = None, cost: int = 0) -> Hyperparameter:
+def _tunable_to_configspace(
+        tunable: Tunable, group_name: Optional[str] = None, cost: int = 0) -> ConfigurationSpace:
     """
-    Convert a single Tunable to an equivalent ConfigSpace Hyperparameter object.
+    Convert a single Tunable to an equivalent ConfigSpace Hyperparameter objects,
```
bpkroth, line 30:
```suggestion
    Convert a single Tunable to an equivalent set of ConfigSpace Hyperparameter objects,
```

mlos_bench/mlos_bench/optimizers/convert_configspace.py
```diff
@@ -38,25 +41,54 @@ def _tunable_to_hyperparameter(
...
+            name=tunable.name, lower=tunable.range[0], upper=tunable.range[1],
+            default_value=tunable.default if tunable.in_range(tunable.default) else None,
+            meta=meta),
+        special_name: CategoricalHyperparameter(
+            name=special_name, choices=tunable.special,
+            default_value=tunable.default if tunable.default in tunable.special else None,
+            meta=meta),
+        type_name: CategoricalHyperparameter(
+            name=type_name, choices=["special", "range"], default_value="special",
+            weights=[0.5, 0.5]),  # TODO: Make weights configurable; FLAML requires uniform weights.
```
bpkroth, line 85: Make an Issue to track this?

mlos_bench/mlos_bench/optimizers/convert_configspace.py
```diff
@@ -97,5 +130,91 @@ def tunable_values_to_configuration(tunables: TunableGroups) -> Configuration:
     Configuration
         A ConfigSpace Configuration.
     """
+    values: Dict[str, TunableValue] = {}
+    for (tunable, _group) in tunables:
+        if tunable.special:
+            (special_name, type_name) = special_param_names(tunable.name)
+            if tunable.value in tunable.special:
+                values[type_name] = "special"
```
bpkroth, line 138: Maybe make an `Enum` class for "special" and "range" static strings for better checking enforcement at dev time.

mlos_bench/mlos_bench/optimizers/convert_configspace.py
```diff
@@ -97,5 +130,91 @@ def tunable_values_to_configuration(tunables: TunableGroups) -> Configuration:
...
+            values[tunable.name] = tunable.value
     configspace = tunable_groups_to_configspace(tunables)
-    return Configuration(configspace, values={tunable.name: tunable.value for (tunable, _group) in tunables})
+    return Configuration(configspace, values=values)
+
+
+def configspace_data_to_tunable_values(data: dict) -> dict:
+    """
+    Remove the fields that correspond to special values in ConfigSpace.
+    In particular, remove `!type__` keys and trim `!special` suffixes.
```
bpkroth, line 152:
```suggestion
    In particular, remove and keys suffixes added by `special_param_names`.
```

Reword to avoid including the specific text for now? (in case it changes)
Maybe just reference the other routine where it was added.

mlos_bench/mlos_bench/optimizers/convert_configspace.py
```diff
@@ -97,5 +130,91 @@ def tunable_values_to_configuration(tunables: TunableGroups) -> Configuration:
...
+        The name of the tunable parameter.
+
+    Returns
+    -------
+    special_name : str
+        The name of the hyperparameter that corresponds to the special value.
+    type_name : str
+        The name of the hyperparameter that chooses between the regular and the special values.
+    """
+    return (name + "!special", name + "!type__")
```
bpkroth, line 186: Why the `__` suffix for `!type__` and not `!special`?

mlos_bench/mlos_bench/optimizers/convert_configspace.py
```diff
@@ -97,5 +130,91 @@ def tunable_values_to_configuration(tunables: TunableGroups) -> Configuration:
...
+    ----------
+    name : str
+        The name of the hyperparameter.
+
+    Returns
+    -------
+    stripped_name : str
+        The name of the hyperparameter without the temporary suffix.
+    """
+    return name.split("!", 1)[0]
```
bpkroth, line 220: Does it make sense to use a regex to be more precise about this?  (e.g., in case some user param includes a `!` in the future).

Or should we disallow that character in the `Tunable` constructor `name` check?

mlos_bench/mlos_bench/tests/tunables/tunable_to_configspace_test.py
```diff
@@ -8,16 +8,21 @@
...
-from ConfigSpace import UniformIntegerHyperparameter
-from ConfigSpace import UniformFloatHyperparameter
-from ConfigSpace import CategoricalHyperparameter
-from ConfigSpace import ConfigurationSpace
+from ConfigSpace import (
+    ConfigurationSpace,
+    CategoricalHyperparameter,
+    UniformIntegerHyperparameter,
+    UniformFloatHyperparameter,
+    EqualsCondition,
```
bpkroth, line 16: nit: sort imports

mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
```diff
@@ -128,13 +129,28 @@ def _to_df(self, configs: Sequence[dict]) -> pd.DataFrame:
...
+            df_configs[tunable.name] = df_configs[tunable.name].astype(tunable.dtype)
+            # Add columns for tunables with special values.
+            if tunable.special:
+                (special_name, type_name) = special_param_names(tunable.name)
+                tunables_names += [special_name, type_name]
+                is_special = df_configs[tunable.name].apply(tunable.special.__contains__)
+                df_configs[type_name] = "range"
+                df_configs.loc[is_special, type_name] = "special"
+                if tunable.type == "int":
+                    # Make int column NULLABLE:
```
bpkroth, line 149: Are there new tests for this?  Or did we already have one?

mlos_bench/mlos_bench/tests/tunables/tunables_assign_test.py
```diff
@@ -162,3 +162,37 @@ def test_tunable_assign_null_to_float(tunable_float: Tunable) -> None:
...
     with pytest.raises(TypeError):
         tunable_float.numerical_value = None    # type: ignore[assignment]
+
+
+def test_tunable_assign_special(tunable_int: Tunable) -> None:
+    """
+    Check the assignment of a special value outside of the range (but declared `special`).
+    """
+    tunable_int.numerical_value = -1
+    assert tunable_int.numerical_value == -1
```
bpkroth, line 172: need an assert for `is_special` too?

mlos_bench/mlos_bench/tunables/tunable.py
```diff
@@ -32,7 +32,7 @@ class TunableDict(TypedDict, total=False):
     default: TunableValue
     values: Optional[List[Optional[str]]]
     range: Optional[Union[Sequence[int], Sequence[float]]]
-    special: Optional[Union[List[int], List[str]]]
```
bpkroth, line 35: Hmm, I can't remember why I did it that way.  Wonder if that was just a mistake, or we really did have a good reason for it to be a `str` as well ...

bpkroth: Wouldn't surprise me if there's some settings that are things like `[0, 1024]` or `AUTO`

bpkroth: Yeah, sqlite has one of those for instance:
> The "PRAGMA cache_spill=N" form of this pragma sets a minimum cache size threshold required for spilling to occur. The number of pages in cache must exceed both the cache_spill threshold and the maximum cache size set by the [PRAGMA cache_size](https://www.sqlite.org/pragma.html#pragma_cache_size) statement in order for spilling to occur.
> The "PRAGMA cache_spill=boolean" form of this pragma applies across all databases attached to the database connection. But the "PRAGMA cache_spill=N" form of this statement only applies to the "main" schema or whatever other schema is specified as part of the statement.

https://www.sqlite.org/pragma.html#pragma_cache_spill

bpkroth: Whereas, I think it would be unlikely for a `float` to be used for a special value (though maybe we shouldn't preclude that either).

mlos_bench/mlos_bench/tunables/tunable.py
```diff
@@ -92,8 +92,8 @@ def _sanity_check(self) -> None:
                 raise ValueError(f"Range must be None for the categorical type tunable {self}")
             if len(set(self._values)) != len(self._values):
                 raise ValueError(f"Values must be unique for the categorical type tunable {self}")
-            if self._special is not None:
-                raise ValueError(f"Special values must be None for the categorical type tunable {self}")
+            if self._special:
```
bpkroth, line 95: I think this may fail for `False` special values.
Not sure why we'd have those but better to support and avoid weird hard to debug conditions for users I think.
Or throw a warning in that case?

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -132,11 +132,11 @@ def suggest(self, context: Optional[pd.DataFrame] = None, defaults: bool = False
             configuration = self._suggest(context)
             assert len(configuration) == 1, \
                 "Suggest must return a single configuration."
-            assert len(configuration.columns) == len(self.optimizer_parameter_space.values()), \
+            assert set(configuration.columns).issubset(set(self.optimizer_parameter_space)), \
```
bpkroth, line 135: Is this too lenient now?

bpkroth (changes requested):
A few nits here and there, but I think the main thing we should revisit is whether to support mixed numerical and string types for special values.

mlos_core/mlos_core/__init__.py
bpkroth, line 1: Completely unrelated to this file, but wanted a way to track this TODO: can you please run the following in your PR?
```sh
./scripts/update-version.sh --no-tag patch
```
That will save us a separate PR to bump the version.

mlos_bench/mlos_bench/optimizers/convert_configspace.py
motus: Done. #635

mlos_core/mlos_core/__init__.py
motus: done

mlos_bench/mlos_bench/optimizers/convert_configspace.py
motus: Done. #635

mlos_bench/mlos_bench/tunables/tunable.py
motus: agree. Let's leave it like this for now, and I'll see how we can improve it later. Most likely, I'll revisit teh code when I address #635

mlos_core/mlos_core/optimizers/optimizer.py
motus: I think it's good enough as a sanity check - in fact, it's better than checking just the number of hyperparameters. I don't know an easy way to make sure we have a consistent set of hyperparameters that satisfy all ConfigSpace conditionals while not omitting any required values. I'll do some research and maybe send another PR later

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -132,11 +132,11 @@ def suggest(self, context: Optional[pd.DataFrame] = None, defaults: bool = False
...
                 "Suggest returned a configuration with the wrong number of parameters."
```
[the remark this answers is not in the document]
motus: done

mlos_bench/mlos_bench/tunables/tunable.py
motus: nope - `_special` is always a list, and if there are no special values, it is equal to `[]`. If `False` is a special value, `self._special` will contain `[False]` (i.e., a list of 1 value)

mlos_bench/mlos_bench/tests/tunables/tunables_assign_test.py
motus: `Tunable` does not have an explicit `is_special` property, but I can make one

mlos_bench/mlos_bench/optimizers/convert_configspace.py
motus: uh, that's some artifact from the earlier commits, where I was using `__` suffix to tell apart the conditionals and regular hyperparameters. I'll update it now

mlos_bench/mlos_bench/optimizers/convert_configspace.py
motus: Let's just disallow `!` in the tunables' names for now. It is likely we will have other special hyperparameters and conditionals in the config space later and `*!SUFFIX` will come in handy

mlos_bench/mlos_bench/optimizers/convert_configspace.py
motus: done

mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
motus: we do have tests for the optimzier, but not specifically for the `_to_df()` functionality. I think it is a great idea to add some now. I've opened an issue #636 and will submit a PR soon after we merge this one

mlos_bench/mlos_bench/optimizers/convert_configspace.py
```diff
@@ -8,24 +8,38 @@
...
+    SPECIAL = "special"
+    RANGE = "range"
+
+
+def _tunable_to_configspace(
+        tunable: Tunable, group_name: Optional[str] = None, cost: int = 0) -> ConfigurationSpace:
+    """
+    Convert a single Tunable to an equivalent set of ConfigSpace Hyperparameter objects,
+    wrapped in a ConfigurationSpace for composability.
+    Note: this may more than one Hyperparameter in the case of special value handling.
```
bpkroth, line 42:
```suggestion
    Note: this may be more than one Hyperparameter in the case of special value handling.
```
correcting my own correction 🤦

mlos_bench/mlos_bench/tests/tunables/tunable_definition_test.py
```diff
@@ -12,6 +12,14 @@
 from mlos_bench.tunables.tunable import Tunable
 
 
+def test_tunable_name() -> None:
+    """
+    Check that tunable name is valid.
+    """
+    with pytest.raises(ValueError):
+        Tunable(name='test!tunable', config={"type": "float", "range": [0, 1], "default": 0})
```
bpkroth, line 20:
```suggestion
        # ! characters are currently disallowed in tunable names
        Tunable(name='test!tunable', config={"type": "float", "range": [0, 1], "default": 0})
```

PR #549, opened 2023-10-19
Accept non-string categoricals

bpkroth:
Closes #450

Just the simple version to start with: convert all categorical inputs to strings.

PR #623, opened 2024-01-03
Tunable fixups

bpkroth:
Improvements on #549 

- Handles the case where the default for a tunable is `False` (else it wouldn't be converted to a `str`).
- Adds more details during debug and error output to make troubleshooting easier


bpkroth:
@motus please have a look at this when you get a sec.  It's needed for the sqlite-autotuning demo to work

motus (approved):
Looks great! Thanks for fixing it!

PR #624, opened 2024-01-05
Introduce basic experiment visualization module mlos_viz via dabl

bpkroth:

PR #628, opened 2024-01-10
Add support for storing the optimization targets and direction of an experiment

bpkroth:
This PR is useful for mlos-viz and dabl wrapper (#624) to be able to automatically graph the results for a given optimization target, for instance via something like the following:

```python
for opt_target in exp.objectives:
    dabl.plot(exp.results, opt_target)
```

Since the prior efforts on capturing this data in the Trial metadata are somewhat problematic (allow conflicting changes between runs of an experiment, don't support multi-objective), we extend them to also store values directly as a part of the Experiment, which is a somewhat more appropriate location.  Upon retrieval, an attempt is also made to merge the two data sources for backwards compatibility.

This PR does not enforce strictness on that metadata, but future versions could (e.g., disallow resuming an Experiment if it looks like the objective targets have changed.  In that case the prior Trial results can potentially still be used to prewarm a new Experiment's optimizer).

### DIFF
--- a/mlos_bench/mlos_bench/storage/base_experiment_data.py
+++ b/mlos_bench/mlos_bench/storage/base_experiment_data.py
@@ -79,6 +79,6 @@ class ExperimentData(metaclass=ABCMeta):
         results : pandas.DataFrame
             A DataFrame with configurations and results from all trials of the experiment.
             Has columns [trial_id, config_id, ts_start, ts_end, status]
-            followed by config parameters and trial results. The latter can be NULLs
+            followed by tunable config parameters and trial results. The latter can be NULLs
             if the trial was not successful.
         """
--- a/mlos_bench/mlos_bench/storage/base_experiment_data.py
+++ b/mlos_bench/mlos_bench/storage/base_experiment_data.py
@@ -64,7 +64,7 @@ class ExperimentData(metaclass=ABCMeta):
 
         Returns
         -------
-        objectives : Dict[str, objectives]
+        objectives : Dict[str, objective]
             A dictionary of the experiment's objective names (optimization_targets)
             and their directions (e.g., min or max).
         """
--- a/mlos_bench/mlos_bench/storage/base_experiment_data.py
+++ b/mlos_bench/mlos_bench/storage/base_experiment_data.py
@@ -58,6 +58,19 @@ class ExperimentData(metaclass=ABCMeta):
 
     @property
     @abstractmethod
+    def objectives(self) -> Dict[str, str]:
+        """
+        Retrieve the experiment's objectives data from the storage.
+
+        Returns
+        -------
+        objectives : Dict[str, objectives]
+            A dictionary of the experiment's objective names (optimization_targets)
+            and their directions (e.g., min or max).
+        """
+
+    @property
+    @abstractmethod
     def trials(self) -> Dict[int, TrialData]:
         """
         Retrieve the trials' data from the storage.
--- a/mlos_bench/mlos_bench/storage/base_storage.py
+++ b/mlos_bench/mlos_bench/storage/base_storage.py
@@ -259,6 +259,7 @@ class Storage(metaclass=ABCMeta):
             self._trial_id = trial_id
             self._config_id = config_id
             self._opt_target = opt_target
+            assert opt_direction in {"min", "max"}
             self._opt_direction = opt_direction
             self._config = config or {}
 
--- a/mlos_bench/mlos_bench/storage/base_storage.py
+++ b/mlos_bench/mlos_bench/storage/base_storage.py
@@ -188,7 +188,7 @@ class Storage(metaclass=ABCMeta):
             """
 
         @abstractmethod
-        def load_config(self, config_id: int) -> Dict[str, Any]:
+        def load_tunable_config(self, config_id: int) -> Dict[str, Any]:
             """
             Load tunable values for a given config ID.
             """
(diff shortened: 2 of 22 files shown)


motus:
wait, we already store that data in the `trial_param` table - see https://github.com/microsoft/MLOS/blob/2032d7908fd3dfd748cf30c331013cd4d0f66541/mlos_bench/mlos_bench/run.py#L120

you can add more key/value pairs to store with each trial there

bpkroth:
> wait, we already store that data in the `trial_param` table - see
> 
> https://github.com/microsoft/MLOS/blob/2032d7908fd3dfd748cf30c331013cd4d0f66541/mlos_bench/mlos_bench/run.py#L120-L124
> 
> 
> you can add more key/value pairs to store with each trial there

We had a conversation about this offline.  I think the conclusion was that:
1. The optimization configuration is really a property of the Experiment and not the Trial, so if something about the Experiment's optimization goal's configuration changes, we should really require a new Experiment, though previously we didn't store or enforce this.
2. That doesn't preclude reusing *some* TrialData from a previous Experiment to pre-warm a new one.
3. However, storing metadata about the Optimization goals used during a given Trial, which potentially redundant, may still be useful.
    Though, with that in mind, we may also want to store other parameter data used too ...

I think with that in mind, I will still pursue this PR and a future one can enforce the strictness aspects of not changing things, perhaps after we make it easier to merge other "related" TrialData (though capturing which those are might itself be tricky).

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -112,15 +115,28 @@ def experiment(self, *,
...
-        def __init__(self, tunables: TunableGroups, experiment_id: str, root_env_config: str):
+        def __init__(self,
+                     *,
+                     tunables: TunableGroups,
+                     experiment_id: str,
+                     trial_id: int,
+                     root_env_config: str,
+                     description: str,
+                     opt_target: str,
+                     opt_direction: str):
```
bpkroth, line 132: Moving some of these args from the sql class to the base class since I think they make sense for any implementation.

mlos_bench/mlos_bench/tests/storage/trial_data_test.py
```diff
@@ -0,0 +1,9 @@
+#
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+#
+"""
+Unit tests for loading the trial metadata.
+"""
+
+# TODO
```
bpkroth, line 9: This is a stub for now.  There were no tests for TrialData loading previously.  I didn't really change that code yet, so didn't add tests for it yet, but wanted to note that we need some while I was adding some for the ExperimentData class.

mlos_bench/mlos_bench/storage/sql/experiment_data.py
```diff
@@ -34,6 +38,55 @@ def __init__(self, *, engine: Engine, schema: DbSchema, exp_id: str,
...
+    def objectives(self) -> Dict[str, str]:
+        objectives: Dict[str, str] = {}
+        # First try to lookup the objectives from the experiment metadata in the storage layer.
+        if hasattr(self._schema, "objectives"):
+            with self._engine.connect() as conn:
+                objectives_db_data = conn.execute(
+                    self._schema.objectives.select().where(
+                        self._schema.objectives.c.exp_id == self._exp_id,
+                    ).order_by(
+                        self._schema.objectives.c.optimization_target.asc(),
```
motus, line 51: I wonder if the order of optimization targets is important? Maybe we need an extra column like "order" or "priority" in the `objectives` table?

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -83,7 +83,8 @@ def experiment(self, *,
                    trial_id: int,
                    root_env_config: str,
                    description: str,
-                   opt_target: str) -> 'Storage.Experiment':
+                   opt_target: str,
+                   opt_direction: str) -> 'Storage.Experiment':
```
motus, line 87: I think that optimization direction can be nullable - e.g. when we run a standalone benchmark. (we should still have the optimization target, though, because even in a standalone benchmark we still measure something)

mlos_bench/mlos_bench/storage/sql/experiment_data.py
bpkroth: In true multi-objective, no, they should be equally explored by the optimizer, but it's not a bad thing to include for future support.
How about just an optional floating point "weight"?

bpkroth: I added the column for now, but am leaving it not fully supported in the rest of the APIs just yet as we don't fully support multi-objective yet anyways and adding weights would change a lot of these APIs more that I need for the initial dabl one sooner.

mlos_bench/mlos_bench/storage/base_experiment_data.py
```diff
@@ -56,6 +56,19 @@ def root_env_config(self) -> Tuple[str, str, str]:
...
 
+    @property
+    @abstractmethod
+    def objectives(self) -> Dict[str, str]:
+        """
+        Retrieve the experiment's objectives data from the storage.
+
+        Returns
+        -------
+        objectives : Dict[str, objectives]
```
motus, line 67:
```suggestion
        objectives : Dict[str, objective]
```

motus (approved):
Looks good! I leave it up to you to keep the optimization direction nullable or roll back to the non-nullable variant

PR #629, opened 2024-01-11
Rename and improve tunable vs config API and documentation

bpkroth:
Minor tweaks to make it easier to understand the difference between a (tunable) config (for an Environment) and a config (part of the framework) when interacting with the storage layer (who's schema switches terminology slightly than the code).


motus (approved):
looks good!

PR #630, opened 2024-01-11
Add more info on const_args vs. tunable_params; start the DEVNOTES

motus:


mlos_bench/mlos_bench/config/README.md
```diff
@@ -8,6 +8,13 @@ In general the `config` directory layout follows that of the `mlos_bench` module
 
 Full end-to-end examples are provided in the [`cli`](./cli/) directory, and typically and make use of the root [`CompositeEnvironments`](./environments/root/) to combine multiple [`Environments`](./environments/), also referencing [`Services`](./services/), [`Storage`](./storage/), and [`Optimizer`](./optimizers/) configs, into a single [`mlos_bench`](../run.py) run.
 
+## Config parameters
+
+An `Environment` configuration can have two sections, `const_args` and `tunable_params`.
```
bpkroth, line 13: Also `required_args`

motus: ooh that's right. I will add a section on those in connection with the globals and composite envs

mlos_bench/mlos_bench/config/README.md
```diff
@@ -8,6 +8,13 @@ In general the `config` directory layout follows that of the `mlos_bench` module
...
+At runtime, at each trial the data from both sections will be merged and passed to the environment as a single dictionary of key/value pairs.
+The difference between `const_args` and `tunable_params` is that the former values stay constant across all trials, whereas `tunable_params` values can change from one trial to the next (e.g., when the optimizer proposes a new configuration).
```
bpkroth, line 15: Not sure this is entirely true either.
`const_args` can read from environment variables or from a value further up in the stack of configs, so they're not exactly "constant", but more so "defaults" for `required_args`

mlos_bench/mlos_bench/config/README.md
```diff
@@ -8,6 +8,13 @@ In general the `config` directory layout follows that of the `mlos_bench` module
...
+The storage system will only save the values of the `tunable_params` for each trial; we rely on git to preserve the config files along with the `const_args` values, and save the commit hash of the configs for each experiment.
```
bpkroth, line 16: The storage system maybe should save the const_args too ...

mlos_bench/mlos_bench/config/README.md
bpkroth, line 15: Maybe another note to add: when you would use `const_args` (e.g., for adjusting the behavior of reusable scripts).

mlos_bench/mlos_bench/DEVNOTES.md
```diff
@@ -0,0 +1,56 @@
...
+}
+```
+
+### `const_args` and `tunable_params`
+
+Note that in the config above we have three groups of parameters.
+`tunable_params` are the configuration parameters that are passed to the `.setup()` call on each trial.
+They are external to the environment, and are usually either suggested by the optimizer, or specified explicitly by the user (e.g., when benchmarking a certain configuration).
+`const_args` is a loose collection of key/value pairs that complement the `tunable_params` values.
+These values do not change from one trial to the next, but they also appear as input parameters for each `Environment.setup()` call.
```
bpkroth, line 49:
```suggestion
These values do not change from one trial to the next (though could change from one `mlos_bench` `run.py` to another via environment variable consumption), but they also appear as input parameters for each `Environment.setup()` call to use in their `setup` and `run` scripts, for instance.
```

mlos_bench/mlos_bench/config/README.md
motus: that's a slippery slope - some const_args values are sensitive, and most of them never change during the experiment. There is also a possibility of making some const_arg a tunable, some const_args get different values in siblings environments, etc. etc. I'd say let's keep it in git for now, and later we can think of something. In general, I agree that it would be nice to be able to retrieve all such parameters as a table or dict without parsing the configs or instantiating the environments' stack

bpkroth (approved):
Good start.  I may do some tweaks on it more later.

PR #626, opened 2024-01-09
Pypi packaging

bpkroth:

PR #631, opened 2024-01-16
CI tweak: use "release" event instead of "tag push" for pypi publishing

bpkroth:
Minor tweak on #626: instead of acting on tag push events, we only act on "release" events.  This gives us a chance to fix bad tag pushes (which otherwise don't have much gated checking on them) prior to attempting to publish a version on pypi, which cannot be overwritten (only superseded).

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.2.5
+current_version = 0.2.6
 commit = True
 tag = True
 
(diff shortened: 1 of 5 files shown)


bpkroth:
*NOTE*: The version bump currently in this PR was use for testing on my fork, but should be replaced with the one from #624 after it's merged.  I don't think there's any point in publishing this particular patch version release - there's no functional change from the last one.

motus (approved):
Looks good! I wonder if there is a way to store version in a single place?

bpkroth:
> Looks good! I wonder if there is a way to store version in a single place?

It'd be nice, but no not really.  We publish several packages out of the same repo, which is somewhat less common.  That's in part what the `bumpversion` script is for to help us with managing that.

PR #639, opened 2024-01-18
Tell pylint to ignore FIXME comments in our code.

bpkroth:
We can find them other ways.


motus (approved):
Thank you! Let's maybe merge it after my PR so that my fixes get into the new release?

PR #633, opened 2024-01-16
Improve mlos-viz for multiple repeats of a config and add tests

bpkroth:
- Mark `mlos_viz` as `typed` for `mypy`
- Bump version
- Mock calls to matplotlib/dabl for testing
- Add plotting of top-N configs
- Improve plots for handling repeat config trials via variance error bars

PR #644, opened 2024-01-19
Remove TunableGroups from Storage classes and add TrialData tests

bpkroth:
Work split out from #633

- Reorgs the unit test fixtures for reuse by mlos_viz (future PR)
- Removes TunablesGroups from Storage classes. Currently unnecessary, belongs in Experiment, and causes inconsistencies with ExperimentData interactions.
- Adjusts the initialization code to match.


mlos_bench/mlos_bench/storage/base_trial_data.py
```diff
@@ -46,38 +48,53 @@ def exp_id(self) -> str:
...
+    def _df_to_dict(self, dataframe: pandas.DataFrame) -> Dict[str, Optional[TunableValue]]:
+        """Utility function to convert certain key-value dataframe formats to a dict."""
+        if dataframe.columns.tolist() == ['metric', 'value']:
+            dataframe = dataframe.copy()
+            dataframe.rename(columns={'metric': 'parameter'}, inplace=True)
+        assert dataframe.columns.tolist() == ['parameter', 'value']
+        data = {}
+        for _, row in dataframe.iterrows():
+            assert isinstance(row['parameter'], str)
+            assert row['value'] is None or isinstance(row['value'], (str, int, float))
```
motus, line 92:
```python
from types import NoneType
```
and
```suggestion
            assert isinstance(row['value'], (NoneType, str, int, float))
```

mlos_bench/mlos_bench/tests/tunable_groups.py
```diff
@@ -0,0 +1,103 @@
...
+
+import json5 as json
+
+from mlos_bench.config.schemas import ConfigSchema
+from mlos_bench.tunables.covariant_group import CovariantTunableGroup
+from mlos_bench.tunables.tunable_groups import TunableGroups
+
+# pylint: disable=redefined-outer-name
+
+TUNABLE_GROUPS_JSON = """
```
motus, line 21: Don't we have it already in `conftest.py` on the same level? Maybe let's keep it there and reduce the duplication? (Also, I have changes coming to that JSON in my PR (adding weights, etc))

mlos_bench/mlos_bench/storage/base_trial_data.py
bpkroth: We've had issues with NoneType support across python versions in the past, so I skipped this for now.

mlos_bench/mlos_bench/tests/tunable_groups.py
bpkroth: In #644 I'm reusing these definitions for mlos-viz tests.  We did something similar in `mlos_bench/mlos_bench/tests/services/remote/ssh/fixtures.py`.

I can separate those changes out again if you want, but it was easiest to apply those patches all at once here.

bpkroth: Actually, might make sense to move those to a `fixtures.py` as well.

bpkroth: And actually, in my original splitting of the work I'd forgotten to completely cleanup the old `conftest.py`.  Should be fixed now.  Please have a look and let me know what you think.  Thanks!

mlos_bench/mlos_bench/tests/environments/__init__.py
```diff
@@ -19,7 +19,7 @@
 
 def check_env_success(env: Environment,
                       tunable_groups: TunableGroups,
-                      expected_results: Dict[str, Union[TunableValue]],
```
bpkroth, line 22: pylint complaint

motus (approved):

PR #646, opened 2024-01-19
Add weights for tunables with categorical or special numerical values

### LINKED ISSUE #635: Allow user to specify custom probabilities for special values of the tunables.
We also need customizable probabilities for picking special vs. regular values by the optimizer.
These values are currently hardcoded in `convert_configspace.py`.

motus:
Closes #635


mlos_bench/mlos_bench/config/schemas/tunables/tunable-params-schema.json
```diff
@@ -35,6 +35,12 @@
                     },
                     "minItems": 1,
                     "uniqueItems": true
+                },
+                "weights": {
```
bpkroth, line 39: Proposal:
Change
```json
"specials": ["flat", "list", "of", "special", "values"],
"weights": ["list", "must", "align", "gives", "no", "context"]
```
to
```json
"specials": ["flat", "list", "implicitly", "uniform", "as", "before"],
"specials": {
  "val1": {
    "description": "docs",
    "meta": {
       "extra": "info",
    },
    "weight": 1  // the default if omitted
  },
},
```

The latter seems more powerful and is easily converted into other forms if need be.

bpkroth: Same should be possible for categoricals btw.

mlos_bench/mlos_bench/tests/tunables/tunable_definition_test.py
```diff
@@ -174,6 +225,103 @@ def test_numerical_tunable_reversed_range(tunable_type: str) -> None:
...
+        "range": [0, 100],
+        "special": [-1, 0],
+        "weights": [0, 10, 90],
+        "default": 0
+    }}
+    """
+    config = json.loads(json_config)
+    tunable = Tunable(name='test', config=config)
+    assert tunable.special == [-1, 0]
+    assert tunable.weights == [0, 10, 90]  # Zero weights are ok
```
bpkroth, line 284: Why are zero weights ok?  That should fail, right?

Also, there are 3 weights and 2 specials?

mlos_bench/mlos_bench/config/schemas/tunables/tunable-params-schema.json
bpkroth: One challenge I see with this version would be weights of the special values relative to the range, but that could be handled via a separate field: for instance
```jsonc
"range_bias": 0.9,
```

Which gives 90% chance of choosing a range value instead of a special value.

bpkroth: Also, there are several other prioritzation policies we could explore here:
- always sample special values first (there are likely to be few of them), then switch over to the range
- have some sort of decay function where we sample specials with high probability initially, and decrease over time
- reverse that

mlos_bench/mlos_bench/tests/tunables/tunable_definition_test.py
motus: Setting some weight to 0 is a very convenient way of (temporarily) disabling the special value.

Last weight is for the values in range. i.e. in this example, we pick `-1` with probability `0.0`, special value of `0` with probability `0.1`, and uniform values in range `[0, 100]` with probability '0.9'

mlos_bench/mlos_bench/config/schemas/tunables/tunable-params-schema.json
motus: That's all good, but I don't want to make this PR too big. My plan is the following:
1. Add weights to special and categorical values and pass this data all the way to the ConfigSpace and the optimizer (this PR)
2. Run some experiments and see how it works in practice, e.g., for SQL Server QO
3. Add a way to specify distributions (beta, log uniform, etc) in another PR and see what else is missing. This can be an opportunity to refactor this PR and see if we need a different syntax for weights and other stuff (e.g., grouping each special value with weight and metadata)

Who knows, maybe we will want to keep weights and distributions separately from the basic tunables' data?

mlos_bench/mlos_bench/tests/tunables/tunable_definition_test.py
bpkroth: Disabling a certain choice makes some sense, but there's a more generic approach there that I think we should consider longer term.  Prior attempts called these "focus" regions.

Anyways, that part I'm okay with for now, however unaligned arrays like this are really hard to read and interpret without extra context.  Some more self-describing data structure I think would be much better.

bpkroth: Here's another option: leave `special` values as is, but make the (new) `weights` property more self describing:

```json
"range": [0, 100],
"specials": [-1, "auto"],
"weights": {
  "!range": .09,
  "0": 0,  // temporarily remove 0 from the range
  "-1": .9, // temporarily really focus on "-1",
  "auto": .01,
},
```

This seems easier to read than an unaligned array and gives us more flexibility later on to extend it to support decays, distributions, etc.

motus: I like that - let me look at other distributions available in ConfigSpace, and see what other parameters we might need to pass. I'll push the update shortly

motus: After some thought, I've decided to keep weights as an array and add a separate `range_weight` parameter, i.e.
```javascript
"type": "int",
"range": [0, 100],
"special": [-1, 0],
"weights": [0.1, 0],
"range_weight": 0.9,
// etc.
```
I don't expect that we will have that many special values, so array looks nicer and more concise to me. Also, it's easier to typecheck comparing to the `"!range"` special key, and it is closer to the internal representation in Python.

bpkroth: I don't think the last point is super important.
Configs, in my opinion, should be generally human readable.  It's up to the tool to handle converting it into something useful to itself.

bpkroth: I'm still not a super huge fan of the two arrays that need to be aligned pattern, but I guess it's ok enough for now to move on.

One request:
- instead of "weights" and "range_weights", let's name it "range_weights" and "special_weights" so that it's slightly more self documenting.

bpkroth: Yeah, I was thinking the second version of the `weights` as a dict that we discussed in this version of the thread, instead of the first one that was about reformatting `specials`, but that's OK.  I think it should be good enough for now.

mlos_bench/mlos_bench/config/schemas/tunables/tunable-params-schema.json
```diff
@@ -73,11 +79,20 @@
...
+                    }
+                },
+                "range_weight": {
+                    "type": "number"
                 }
             },
             "required": ["type", "default", "range"],
             "not": {
-                "required": ["values"]
+                "required": ["values", "weights"]
```
bpkroth, line 95: What's this about?

bpkroth: Ah, I see, `weights` is for `categorical` `values` and `special_weight` and `range_weight` are for "numeric" `ranges` and `specials` 👍

bpkroth: For naming, does it make sense to have `values_weights` instead of just `weights` then to be consistent with the other naming conventions?

motus: done!

mlos_bench/mlos_bench/optimizers/convert_configspace.py
```diff
@@ -75,28 +86,44 @@ def _tunable_to_configspace(
...
+                name=tunable.name,
+                lower=tunable.range[0],
+                upper=tunable.range[1],
                 default_value=tunable.default if tunable.in_range(tunable.default) else None,
                 meta=meta)
         })
 
+    # Compute the probabilities of switching between regular and special values.
+    special_weights: Optional[List[float]] = None
+    switch_weights = [0.5, 0.5]  # FLAML requires uniform weights.
```
bpkroth, line 98: Do we need to issue a `LOG.info()` message for this?

bpkroth (approved):
LGTM

mlos_bench/mlos_bench/optimizers/convert_configspace.py
bpkroth: Resolving this for now, but can come back to it later.

PR #650, opened 2024-01-26
Add `default_tunable_config_id` property to `ExperimentData`

bpkroth:
Splitting work out from #633.

Having an easy way to identify what is the "default" config group for an Experiment makes analysis of improvements easier.

One challenge with this is that we don't currently strictly enforce non-changes between resumptions of an Experiment, so it's currently possible for the user to change the tunables config and adjust what the "default".

Moreover, in the past we didn't store what the default config was in the DB.

By convention we take the first one for now as a fallback.

### DIFF
--- a/mlos_bench/mlos_bench/storage/sql/experiment_data.py
+++ b/mlos_bench/mlos_bench/storage/sql/experiment_data.py
@@ -10,7 +10,7 @@ from typing import Dict
 import logging
 
 import pandas
-from sqlalchemy import Engine, Integer, func
+from sqlalchemy import Engine, Integer, String, func
 
 from mlos_bench.storage.base_experiment_data import ExperimentData
 from mlos_bench.storage.base_trial_data import TrialData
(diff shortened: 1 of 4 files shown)


mlos_bench/mlos_bench/storage/sql/experiment_data.py
```diff
@@ -113,7 +113,8 @@ def tunable_config_trial_groups(self) -> Dict[int, TunableConfigTrialGroupData]:
             tunable_config_trial_groups = conn.execute(
                 self._schema.trial.select().with_only_columns(
                     self._schema.trial.c.config_id,
-                    func.min(self._schema.trial.c.trial_id).cast(Integer).label('tunable_config_trial_group_id'),
+                    func.min(self._schema.trial.c.trial_id).cast(Integer).label(    # pylint: disable=not-callable
```
bpkroth, line 116: false positive

mlos_bench/mlos_bench/storage/sql/experiment_data.py
```diff
@@ -154,6 +155,62 @@ def tunable_configs(self) -> Dict[int, TunableConfigData]:
...
+        """
+        with self._engine.connect() as conn:
+            query_results = conn.execute(
+                self._schema.trial.select().with_only_columns(
+                    self._schema.trial.c.config_id.cast(Integer).label('config_id'),
+                ).where(
+                    self._schema.trial.c.exp_id == self._experiment_id,
+                    self._schema.trial.c.trial_id.in_(
+                        self._schema.trial_param.select().with_only_columns(
+                            func.min(self._schema.trial_param.c.trial_id).cast(Integer).label(  # pylint: disable=not-callable
```
bpkroth, line 179: false positive

motus (approved):

Issue #649, opened 2024-01-23
Schema versioning for Sql Storage


bpkroth:
- [ ] Add a `schema_version` table.
- [ ] If it doesn't exist, infer version 0
- [ ] Maintain code that adjusts the tables step by step from version 0 to version N (the latest)
  - (e.g., column renames, table additions, etc.)

bpkroth:
Better yet, use this:
https://alembic.sqlalchemy.org/en/latest/

bpkroth:
Note:
- table additions are already natively supported via sqlalchemy's `create_all()` call.
- it's only *column* changes that needs additional work.

PR #652, opened 2024-01-26
Remove mistakenly added columns

bpkroth:
#628 mistakenly included an early attempt at adding `optimization_target` and `optimization_direction` to the `experiment` table in the `mlos_bench.storage.sql` backend.

In that PR we later moved it to its own `objectives` table to eventually support multi-objectives.

Nothing accesses those columns now, however including them in the metadata makes it impossible to load storage backends previously created with the old schema since adjusting columns with sqlalchemy's `create_all()` API only considers table existence.

On the contrary, the latter means that we will automatically support old storage backends with the new code for the `objectives` table.

Removing these two columns in the `metadata` schema description simply allows that to proceed without error.

See Also: #649


bpkroth:
See also: #649

motus (approved):

PR #659, opened 2024-02-01
Reorg doc deploy step for new GITHUB_TOKEN permissions requirements

bpkroth:
Fixes our documentation deploy step in the CI pipeline, which is currently failing due to a change to the default permissions of the `GITHUB_TOKEN` which is now read-only unless otherwise specified.

### DIFF
--- a/.github/workflows/devcontainer.yml
+++ b/.github/workflows/devcontainer.yml
@@ -176,6 +176,11 @@ jobs:
         # Make sure we can publish the coverage report.
         rm -f doc/build/html/htmlcov/.gitignore
 
+    - uses: actions/upload-artifact@v4
+      with:
+        name: docs
+        path: doc/build/html
+
     - name: Publish package to Test PyPi
       if: github.event_name == 'release' && github.ref_type == 'tag'
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/devcontainer.yml
```diff
@@ -241,3 +239,40 @@ jobs:
...
         docker push ${{ secrets.ACR_LOGINURL }}/mlos-devcontainer:$image_tag
+
+  DeployDocs:
+    if: github.ref == 'refs/heads/main'
+    needs: DevContainer
+    runs-on: ubuntu-latest
+
+    # Required for github-pages-deploy-action to push to the gh-pages branch.
+    permissions:
+      contents: write
```
bpkroth, line 250: This is the key part of the change, but out of security caution, I wanted to push it to its own separate dependent job so not all other steps also have access to push to that branch of the repo.

bpkroth:
Original announcement:
> ‌GITHUB_TOKEN permissions changes  
‌On February 1 the default permissions for the GITHUB_TOKEN from Read/Write to Read-only. This is a breaking change for workflows that haven't added a perms block to their workflows. It is on you and your team to prepare and take steps to avoid breaking your GitHub Actions build workflow.

motus (approved):

PR #662, opened 2024-02-02
Add log scale and quantization to the tunables

### LINKED ISSUE #664: Support log scale and quantization of numeric range of the Tunables

motus:
Closes #664

### DIFF
--- a/mlos_bench/mlos_bench/config/schemas/tunables/tunable-params-schema.json
+++ b/mlos_bench/mlos_bench/config/schemas/tunables/tunable-params-schema.json
@@ -24,8 +24,7 @@
         },
         "quantization": {
             "description": "The number of buckets to quantize the range into.\nSee Also:\nhttps://automl.github.io/ConfigSpace/main/api/hyperparameters.html#module-ConfigSpace.api.types.float,\nhttps://automl.github.io/ConfigSpace/main/api/hyperparameters.html#module-ConfigSpace.api.types.integer",
-            "$comment": "type left unspecified here",
-            "minimum": 1
+            "$comment": "type left unspecified here"
         },
         "log_scale": {
             "description": "Whether to use log instead of linear scale for the range search.",
@@ -115,7 +114,8 @@
                 },
                 "quantization": {
                     "$ref": "#/$defs/quantization",
-                    "type": "integer"
+                    "type": "integer",
+                    "exclusiveMinimum": 1
                 },
                 "log": {
                     "$ref": "#/$defs/log_scale"
@@ -161,7 +161,8 @@
                 },
                 "quantization": {
                     "$ref": "#/$defs/quantization",
-                    "type": "number"
+                    "type": "number",
+                    "exclusiveMinimum": 0
                 },
                 "log": {
                     "$ref": "#/$defs/log_scale"
--- a/mlos_bench/mlos_bench/tunables/tunable.py
+++ b/mlos_bench/mlos_bench/tunables/tunable.py
@@ -32,7 +32,7 @@ class TunableDict(TypedDict, total=False):
     default: TunableValue
     values: Optional[List[Optional[str]]]
     range: Optional[Union[Sequence[int], Sequence[float]]]
-    quantization: Optional[int]
+    quantization: Optional[Union[int, float]]
     log: Optional[bool]
     special: Optional[Union[List[int], List[float]]]
     values_weights: Optional[List[float]]
@@ -78,7 +78,7 @@ class Tunable:  # pylint: disable=too-many-instance-attributes
             self._values = [str(v) if v is not None else v for v in self._values]
         self._meta: Dict[str, Any] = config.get("meta", {})
         self._range: Optional[Union[Tuple[int, int], Tuple[float, float]]] = None
-        self._quantization: Optional[int] = config.get("quantization")
+        self._quantization: Optional[Union[int, float]] = config.get("quantization")
         self._log: Optional[bool] = config.get("log")
         config_range = config.get("range")
         if config_range is not None:
@@ -95,6 +95,7 @@ class Tunable:  # pylint: disable=too-many-instance-attributes
         self.value = self._default
 
     def _sanity_check(self) -> None:
+        # pylint: disable=too-complex,too-many-branches
         """
         Check if the status of the Tunable is valid, and throw ValueError if it is not.
         """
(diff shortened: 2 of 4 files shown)


mlos_bench/mlos_bench/config/schemas/tunables/tunable-params-schema.json
```diff
@@ -73,6 +73,12 @@
                     "minItems": 2,
                     "maxItems": 2
                 },
+                "quantization": {
+                    "type": "number"
```
bpkroth, line 77:
```suggestion
                    "description": "The number of buckets to divide the range into.  See Also: https://automl.github.io/ConfigSpace/main/api/hyperparameters.html#module-ConfigSpace.api.types.float"
                    "type": "number"
```

Add a descriptive text that can appear during editing with supported IDEs.

Similar for the "log" parameter types.

bpkroth: BTW, I think it should be possible to consolidate the "integer" and "float" param types in the schema a bit to reduce a little duplication.  I'll send a PR along for that.

bpkroth, line 77:
```suggestion
                    "type": "number",
                    "minimum": 1,
```

Also, is it a "number", or an "integer"?

mlos_bench/mlos_bench/config/schemas/tunables/tunable-params-schema.json
```diff
@@ -120,6 +126,12 @@
...
```
bpkroth, line 130: I guess it's only an int for "integer" types:
https://automl.github.io/ConfigSpace/main/api/hyperparameters.html#module-ConfigSpace.api.types.integer

mlos_bench/mlos_bench/tunables/tunable.py
```diff
@@ -32,6 +32,8 @@ class TunableDict(TypedDict, total=False):
     default: TunableValue
     values: Optional[List[Optional[str]]]
     range: Optional[Union[Sequence[int], Sequence[float]]]
+    quantization: Optional[int]
```
bpkroth, line 35: Per discussion above:
- floats accept quantization as a float > 0, but
- ints accept quantization as an int

mlos_bench/mlos_bench/config/schemas/tunables/tunable-params-schema.json
motus: it is `int` for `ConfigSpace`, but JSON schema does not support ints and floats - the only valid type is `number` 😒

mlos_bench/mlos_bench/config/schemas/tunables/tunable-params-schema.json
bpkroth: https://github.com/motus/MLOS/pull/8

mlos_bench/mlos_bench/config/schemas/tunables/tunable-params-schema.json
bpkroth: json has "integer" too:
https://json-schema.org/understanding-json-schema/reference/numeric#integer

mlos_bench/mlos_bench/tests/tunables/tunable_definition_test.py
```diff
@@ -384,6 +421,24 @@ def test_numerical_weights_wrong_values(tunable_type: str) -> None:
...
+    json_config = f"""
+    {{
+        "type": "{tunable_type}",
+        "range": [0, 100],
+        "quantization": 0,
+        "default": 0
+    }}
+    """
+    config = json.loads(json_config)
+    with pytest.raises(ValueError):
```
bpkroth, line 438: This might change to fail with a "SchemaValidation" error now with fixes from https://github.com/motus/MLOS/pull/8

mlos_bench/mlos_bench/tunables/tunable.py
```diff
@@ -115,6 +123,8 @@ def _sanity_check(self) -> None:
                 raise ValueError(f"Values must be None for the numerical type tunable {self}")
             if not self._range or len(self._range) != 2 or self._range[0] >= self._range[1]:
                 raise ValueError(f"Invalid range for tunable {self}: {self._range}")
+            if self._quantization is not None and self._quantization <= 1:
+                raise ValueError(f"Number of quantization points is <= 1: {self}")
```
bpkroth, line 127: I think I only enforce `minimum: 1` in the schema.  Could adjust that up to 2.

bpkroth: But, not sure how that works for a float param with range [0, 4.5] and with quantization factor 1.5 for instance.

bpkroth: For that matter, this is also valid:

```python
import ConfigurationSpace as CS
cs = CS.ConfigurationSpace()
cs.add_hyperparameter(CS.Float(name="float", bounds=(0, 1), q=0.1))
cs.sample_configuration()
```
```python
cs.sample_configuration()
Configuration(values={
  'float': 0.30000000000000004,
})
```

mlos_bench/mlos_bench/config/schemas/tunables/tunable-params-schema.json
motus: merged, thank you!

mlos_bench/mlos_bench/tunables/tunable.py
motus: mm, I cannot see a `float` anywhere. Parameter `q` is an integer number of the intervals to split the hyperparameter's range into. `ConfigSpace` API requires it to be either `int` or `None`
Here's a factory method I invoke for float hyperparameters: https://automl.github.io/ConfigSpace/main/_modules/ConfigSpace/api/types/float.html

motus: **UPDATE:** I've found where `q` can be `float` - it's only when we pass it directly to the constructor of a specific distribution-specific float hyperparameter. I never do that. I use the `ConfigSpace.Float()` factory method instead, and it only takes `q` as `int` - even though all three classes that it can build can take a `float`. I would suggest to keep it as `int` for simplicity - otherwise our code will get bloated by duplicating the `ConfigSpace.Float()` functionality

mlos_bench/mlos_bench/tunables/tunable.py
motus: that code runs, but `mypy` will issue a warning, because `CS.Float` type signature does *not* allow `float q`

mlos_bench/mlos_bench/tunables/tunable.py
bpkroth: Added a PR I think addresses this here:
https://github.com/motus/MLOS/pull/10

mlos_bench/mlos_bench/tunables/tunable.py
bpkroth: I submitted an [Issue](https://github.com/automl/ConfigSpace/issues/349) and [PR](https://github.com/automl/ConfigSpace/pull/350) to fix that upstream.
Meantime, I'd say we just add `type: ignore[arg]` or whatever to ignore that part since it's definitely supported behind the scenes by ConfigSpace - just a type hints bug.

bpkroth: Weird that this didn't fire for me though.  I wonder what's different about mypy in Windows.

mlos_bench/mlos_bench/tunables/tunable.py
motus: merged, thank you!

mlos_bench/mlos_bench/tunables/tunable.py
motus: agree. mypy does not complain on my Windows PC, so I think we can merge it

mlos_bench/mlos_bench/optimizers/convert_configspace.py
```diff
@@ -108,6 +110,8 @@ def _tunable_to_configspace(
             name=tunable.name,
             lower=tunable.range[0],
             upper=tunable.range[1],
+            log=tunable.is_log,
+            q=tunable.quantization,
```
bpkroth, line 114:
```suggestion
            q=tunable.quantization,  # type: ignore[arg-type,unused-ignore]
```

bpkroth: if necessary

bpkroth (approved):
LGTM

PR #663, opened 2024-02-02
Add support for ConfigSpace distributions

### LINKED ISSUE #665: Specify prior distributions for the Tunable parameters
Currently we assume that all numeric tunables have uniform distribution of the values within the specified range.
We have to give the user an ability to specify other prior distributions and their parameters. We need to support at least Beta and Normal distributions.

motus:
To be merged after #662 

Closes #665


mlos_bench/mlos_bench/config/schemas/tunables/tunable-params-schema.json
```diff
@@ -13,6 +13,23 @@
...
+                "type": {
+                    "type": "string",
+                    "enum": ["uniform", "normal", "beta"]
+                },
+                "params": {
+                    "type": "object",
+                    "items": {
+                        "type": "number"
+                    },
+                    "minItems": 1
```
bpkroth, line 28: For auto complete purposes it might be good to include the types of parameters we expect here.
Let it be free form (e.g., allow others), but also include the known properties as of this time (e.g., "mu" and "sigma" for type == "normal", "alpha" and "beta" for "type" == "beta").

Since we know it's currently a fixed set of supported distributions I think this shouldn't be too hard, and also makes it easier for people to fill things in without having to refer to ConfigSpace source code (which also isn't obvious, so may want to additionally include a reference to https://automl.github.io/ConfigSpace/main/api/hyperparameters.html#distributions in the "description" field).

If you want I can cook up a PR for this.

mlos_bench/mlos_bench/config/schemas/tunables/tunable-params-schema.json
```diff
@@ -73,6 +90,15 @@
                     "minItems": 2,
                     "maxItems": 2
                 },
+                "quantization": {
+                    "type": "number"
```
bpkroth, line 94: I guess it's only an "int" for "integer" types:
https://automl.github.io/ConfigSpace/main/api/hyperparameters.html#module-ConfigSpace.api.types.integer

mlos_bench/mlos_bench/config/schemas/tunables/tunable-params-schema.json
bpkroth: https://github.com/motus/MLOS/pull/9

motus: Since there were conflicts in this PR due to the previous merge, I've cherry picked the functionality from it into this PR. Thank you!

mlos_bench/mlos_bench/optimizers/convert_configspace.py
```diff
@@ -76,24 +79,51 @@ def _tunable_to_configspace(
                 meta=meta)
         })
 
+    distribution: Union[Uniform, Normal, Beta, None] = None
+    if tunable.distribution == "uniform":
```
bpkroth, line 83:
```suggestion
    if tunable.distribution == "uniform":
```

Enums instead of static strings?

bpkroth (approved):
LGTM!  Thanks :)

mlos_bench/mlos_bench/optimizers/convert_configspace.py
```diff
@@ -76,24 +79,51 @@ def _tunable_to_configspace(
...
+            meta=meta
+        )
     elif tunable.type == "float":
-        hp_type = UniformFloatHyperparameter
+        range_hp = Float(
+            name=tunable.name,
+            bounds=tunable.range,
+            log=bool(tunable.is_log),
+            q=tunable.quantization,     # type: ignore[arg-type]
+            distribution=distribution,  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
motus: on both! I guess the reason is that the return type is implicitly parameterized on the hyperparameter type and hence the return type must match the type of q, distribution, and bounds

mlos_bench/mlos_bench/optimizers/convert_configspace.py
motus: the type is `Literal["uniform", "normal", "beta"]` so I guess we're ok

Issue #666, opened 2024-02-05
SMAC optimizer does not support mixed input space


ephoris:
Using the SMAC optimizer in conjunction with a mix of different parameter space types will throw an error. Minimum reproducible example is attached.

```python
import pandas as pd
import mlos_core.optimizers
import ConfigSpace as CS


def objective(x: pd.DataFrame):
    out = x.values
    out = out.sum()

    return out


def run_optimization(optimizer: mlos_core.optimizers.BaseOptimizer):
    suggested_value = optimizer.suggest()
    target_value = objective(suggested_value)
    print(f"{suggested_value}")
    print(f"{target_value=}")
    optimizer.register(suggested_value, pd.Series([target_value]))

    return


if __name__ == "__main__":
    parameter_space = CS.ConfigurationSpace(seed=0)
    dummy_val_one = CS.Float("dummy_one", (1, 10.0), default=5.0)
    dummy_val_two = CS.Integer("dummy_two", (1, 10), default=5)
    parameter_space.add_hyperparameters([dummy_val_one, dummy_val_two])

    optimizer = mlos_core.optimizers.SmacOptimizer(parameter_space=parameter_space)

    n_iterations = 10
    for _ in range(n_iterations):
        run_optimization(optimizer)
```

Results in an error 
```
Traceback (most recent call last):
  File "*/tmp/mlos_tmp.py", line 33, in <module>
    run_optimization(optimizer)
  File "*/tmp/mlos_tmp.py", line 18, in run_optimization
    optimizer.register(suggested_value, pd.Series([target_value]))
  File "*/anaconda3/lib/python3.9/site-packages/mlos_core/optimizers/optimizer.py", line 91, in register
    return self._register(configurations, scores, context)
  File "*/anaconda3/lib/python3.9/site-packages/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py", line 247
, in _register
    for config, score in zip(self._to_configspace_configs(configurations), scores.tolist()):
  File "*/anaconda3/lib/python3.9/site-packages/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py", line 337
, in _to_configspace_configs
    return [
  File "*/anaconda3/lib/python3.9/site-packages/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py", line 338
, in <listcomp>
    ConfigSpace.Configuration(self.optimizer_parameter_space, values=config.to_dict())
[...]

PR #667, opened 2024-02-05
Fix mixed numeric datatypes for optimizers

ephoris:
Addressing issue discussed in #666

### DIFF
--- a/mlos_bench/mlos_bench/storage/util.py
+++ b/mlos_bench/mlos_bench/storage/util.py
@@ -30,7 +30,7 @@ def kv_df_to_dict(dataframe: pandas.DataFrame) -> Dict[str, Optional[TunableValu
         dataframe.rename(columns={'metric': 'parameter'}, inplace=True)
     assert dataframe.columns.tolist() == ['parameter', 'value']
     data = {}
-    for _, row in dataframe.iterrows():
+    for _, row in dataframe.astype('O').iterrows():
         assert isinstance(row['parameter'], str)
         assert row['value'] is None or isinstance(row['value'], (str, int, float))
         if row['parameter'] in data:
--- a/mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
+++ b/mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
@@ -335,6 +335,6 @@ class SmacOptimizer(BaseBayesianOptimizer):
             List of ConfigSpace configurations.
         """
         return [
-            ConfigSpace.Configuration(self.optimizer_parameter_space, values=config.to_dict())
-            for (_, config) in configurations.iterrows()
+            ConfigSpace.Configuration(self.optimizer_parameter_space, values=config._asdict())
+            for config in configurations.itertuples(index=False)
         ]
--- a/mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
+++ b/mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
@@ -335,6 +335,6 @@ class SmacOptimizer(BaseBayesianOptimizer):
             List of ConfigSpace configurations.
         """
         return [
-            ConfigSpace.Configuration(self.optimizer_parameter_space, values=config._asdict())
-            for config in configurations.itertuples(index=False)
+            ConfigSpace.Configuration(self.optimizer_parameter_space, values=config.to_dict())
+            for (_, config) in configurations.astype('O').iterrows()
         ]
--- a/mlos_core/mlos_core/spaces/adapters/llamatune.py
+++ b/mlos_core/mlos_core/spaces/adapters/llamatune.py
@@ -90,8 +90,8 @@ class LlamaTuneAdapter(BaseSpaceAdapter):   # pylint: disable=too-many-instance-
     def inverse_transform(self, configurations: pd.DataFrame) -> pd.DataFrame:
         target_configurations = []
         for config in configurations.itertuples(index=False):
-            config = config._asdict()
-            configuration = ConfigSpace.Configuration(self.orig_parameter_space, values=config)
+            configuration = ConfigSpace.Configuration(
+                self.orig_parameter_space, values=config._asdict())
 
             target_config = self._suggested_configs.get(configuration, None)
             # NOTE: HeSBO is a non-linear projection method, and does not inherently support inverse projection
--- a/mlos_core/mlos_core/spaces/adapters/llamatune.py
+++ b/mlos_core/mlos_core/spaces/adapters/llamatune.py
@@ -89,9 +89,9 @@ class LlamaTuneAdapter(BaseSpaceAdapter):   # pylint: disable=too-many-instance-
 
     def inverse_transform(self, configurations: pd.DataFrame) -> pd.DataFrame:
         target_configurations = []
-        for config in configurations.itertuples(index=False):
+        for (_, config) in configurations.astype('O').iterrows():
             configuration = ConfigSpace.Configuration(
-                self.orig_parameter_space, values=config._asdict())
+                self.orig_parameter_space, values=config.to_dict())
 
             target_config = self._suggested_configs.get(configuration, None)
             # NOTE: HeSBO is a non-linear projection method, and does not inherently support inverse projection
--- a/mlos_core/mlos_core/spaces/adapters/llamatune.py
+++ b/mlos_core/mlos_core/spaces/adapters/llamatune.py
@@ -89,8 +89,9 @@ class LlamaTuneAdapter(BaseSpaceAdapter):   # pylint: disable=too-many-instance-
 
     def inverse_transform(self, configurations: pd.DataFrame) -> pd.DataFrame:
         target_configurations = []
-        for (_, config) in configurations.iterrows():
-            configuration = ConfigSpace.Configuration(self.orig_parameter_space, values=config.to_dict())
+        for config in configurations.itertuples(index=False):
+            config = config._asdict()
+            configuration = ConfigSpace.Configuration(self.orig_parameter_space, values=config)
 
             target_config = self._suggested_configs.get(configuration, None)
             # NOTE: HeSBO is a non-linear projection method, and does not inherently support inverse projection
--- a/mlos_core/mlos_core/tests/optimizers/optimizer_test.py
+++ b/mlos_core/mlos_core/tests/optimizers/optimizer_test.py
@@ -23,7 +23,7 @@ from mlos_core.optimizers import (
 from mlos_core.optimizers.bayesian_optimizers import BaseBayesianOptimizer, SmacOptimizer
 from mlos_core.spaces.adapters import SpaceAdapterType
 
-from mlos_core.tests import get_all_concrete_subclasses
+from mlos_core.tests import get_all_concrete_subclasses, SEED
 
 
 _LOG = logging.getLogger(__name__)
@@ -76,7 +76,7 @@ def test_basic_interface_toy_problem(configuration_space: CS.ConfigurationSpace,
         ret: npt.ArrayLike = (6 * x - 2)**2 * np.sin(12 * x - 4)
         return ret
     # Emukit doesn't allow specifying a random state, so we set the global seed.
-    np.random.seed(42)
+    np.random.seed(SEED)
     optimizer = optimizer_class(parameter_space=configuration_space, **kwargs)
 
     with pytest.raises(ValueError, match="No observations"):
@@ -320,7 +320,7 @@ def test_mixed_numeric_type_input_space_types(optimizer_type: Optional[Optimizer
         ret: pd.Series = point["x"] + point["y"]
         return ret
 
-    input_space = CS.ConfigurationSpace(seed=2169)
+    input_space = CS.ConfigurationSpace(seed=SEED)
     # add a mix of numeric datatypes
     input_space.add_hyperparameter(CS.UniformIntegerHyperparameter(name='x', lower=0, upper=5))
     input_space.add_hyperparameter(CS.UniformFloatHyperparameter(name='y', lower=0.0, upper=5.0))
@@ -347,14 +347,13 @@ def test_mixed_numeric_type_input_space_types(optimizer_type: Optional[Optimizer
         suggestion = optimizer.suggest()
         assert isinstance(suggestion, pd.DataFrame)
         assert (suggestion.columns == ['x', 'y']).all()
-        # Build suggestion mapping to cooperate with Configuration, note that
-        # doing a .iloc[0].to_dict() will cause pandas convert all numeric types
-        # to float64
-        tmp_suggest = {'x': suggestion['x'].values[0], 'y': suggestion['y'].values[0]}
-        # check that suggestion is in the space
-        configuration = CS.Configuration(optimizer.parameter_space, tmp_suggest)
+        # Check suggestion values are the expected dtype
+        assert isinstance(suggestion['x'].iloc[0], np.integer)
+        assert isinstance(suggestion['y'].iloc[0], np.floating)
+        # Check that suggestion is in the space
+        test_configuration = CS.Configuration(optimizer.parameter_space, suggestion.astype('O').iloc[0].to_dict())
         # Raises an error if outside of configuration space
-        configuration.is_valid_configuration()
+        test_configuration.is_valid_configuration()
         observation = objective(suggestion)
         assert isinstance(observation, pd.Series)
         optimizer.register(suggestion, observation)
--- a/mlos_core/mlos_core/tests/optimizers/optimizer_test.py
+++ b/mlos_core/mlos_core/tests/optimizers/optimizer_test.py
@@ -354,6 +354,7 @@ def test_mixed_numeric_type_input_space_types(optimizer_type: Optional[Optimizer
         test_configuration = CS.Configuration(optimizer.parameter_space, suggestion.astype('O').iloc[0].to_dict())
         # Raises an error if outside of configuration space
         test_configuration.is_valid_configuration()
+        # Test registering the suggested configuration with a score.
         observation = objective(suggestion)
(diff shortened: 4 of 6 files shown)


ephoris:
@microsoft-github-policy-service agree

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -335,6 +335,6 @@ def _to_configspace_configs(self, configurations: pd.DataFrame) -> List[ConfigSp
             List of ConfigSpace configurations.
         """
         return [
-            ConfigSpace.Configuration(self.optimizer_parameter_space, values=config.to_dict())
-            for (_, config) in configurations.iterrows()
+            ConfigSpace.Configuration(self.optimizer_parameter_space, values=config._asdict())
+            for config in configurations.itertuples(index=False)
```
motus, line 339: That's neat. We should use it in other similar situations. I did not know that `namedtuple._asdict()` is not private.

P.S. could you please check that pylint is not complaining about accessing a private method - and if it does, add a corresponding `# pylint disable:...` comment?

mlos_core/mlos_core/spaces/adapters/llamatune.py
```diff
@@ -89,8 +89,9 @@ def target_parameter_space(self) -> ConfigSpace.ConfigurationSpace:
 
     def inverse_transform(self, configurations: pd.DataFrame) -> pd.DataFrame:
         target_configurations = []
-        for (_, config) in configurations.iterrows():
-            configuration = ConfigSpace.Configuration(self.orig_parameter_space, values=config.to_dict())
+        for config in configurations.itertuples(index=False):
+            config = config._asdict()
+            configuration = ConfigSpace.Configuration(self.orig_parameter_space, values=config)
```
motus, line 94: `config` is used only once, so maybe
```suggestion
            configuration = ConfigSpace.Configuration(
                self.orig_parameter_space, values=config._asdict())
```

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
```diff
@@ -1,5 +1,3 @@
-#
-# Copyright (c) Microsoft Corporation.
```
motus, line 2: is it not copyrighted by Microsoft anymore? @bpkroth do you know what's our policy?

motus (approved):
Looks nice! Next step, we should make sure LlamaTune does not choke on ConfigSpace instances that have conditionals (like, with the tunables that have special values). For starters, LlamaTune should be able to support that input: [`tunable_to_configspace_test.py:32`](https://github.com/microsoft/MLOS/blob/100d56772e3458e0fd9573593b207704c7f8d1d2/mlos_bench/mlos_bench/tests/tunables/tunable_to_configspace_test.py#L32)

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
ephoris: Whoops that's on me. That's embarrassing I did not intend to remove those lines.

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
ephoris: I was also surprised it is not a private method.

pylint looks good on my end

ephoris:
Hm okay I have noticed that the [`llamatune_opt_test.py:49`](https://github.com/microsoft/MLOS/blob/main/mlos_bench/mlos_bench/tests/optimizers/llamatune_opt_test.py#L49-L65) failed. Looking into it, I see that some test in [`mlos_core_opt_df_test.py:45`](https://github.com/microsoft/MLOS/blob/100d56772e3458e0fd9573593b207704c7f8d1d2/mlos_bench/mlos_bench/tests/optimizers/mlos_core_opt_df_test.py#L45-L46) define parameters with special characters. For example `kernel_sched_migration_cost_ns!type`. Normally this column name is fine if we're converting from `pandas.DataFrame` to `dict` converting, but `NamedTuples` will parse this with undefined behavior, which is why I am getting a test failure. I think we might have to use a different solution instead of `pandas.DataFrame.itertuples()`.

ephoris:
Since I am assuming MLOS wants to support generic strings for hyper parameters, we can revert back to iterrows(). Then by typecasting the dataframe to the `object` type, we can preserve numeric `dtypes` when calling `iterrows()` (see [`smac_optimizers.py:339`](https://github.com/microsoft/MLOS/pull/667/files#diff-5c515d52f65ac919621de24fc8ed1b0fbaccd732deed766dac3c6a137268bc5aR339)). It's not as elegant unfortunately, but this prevents pandas from upcasting integers to floats.

lmnotran:
@ephoris It looks like the DevContainer check is failing. Could you take a look, please?

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
```diff
@@ -298,3 +298,69 @@ def test_optimizer_type_defs(optimizer_class: Type[BaseOptimizer]) -> None:
...
+        optimizer = OptimizerFactory.create(
+            parameter_space=input_space,
+            optimizer_kwargs=kwargs,
+        )
+    else:
+        optimizer = OptimizerFactory.create(
+            parameter_space=input_space,
+            optimizer_type=optimizer_type,
+            optimizer_kwargs=kwargs,
+        )
```
lmnotran, line 338: Does doing something like this instead make sense?
```suggestion

    optimizer_create_args = {
        "parameter_space": input_space,
        "optimizer_kwargs": kwargs,
    }

    if optimizer_type:
        optimizer_create_args['optimizer_type'] = optimizer_type

    optimizer = OptimizerFactory.create(**optimizer_create_args )
```

ephoris:
Linter failed on [983763f](https://github.com/microsoft/MLOS/pull/667/commits/983763f96eaa4841926b59f2107be8f389e789b5). Fixed then reran linter locally, sorry for the spam. I think workflows should pass this time around.

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
```diff
@@ -298,3 +298,69 @@ def test_optimizer_type_defs(optimizer_class: Type[BaseOptimizer]) -> None:
...
+    max_iterations = 10
+    if kwargs is None:
+        kwargs = {}
+
+    def objective(point: pd.DataFrame) -> pd.Series:
+        # mix of hyperparameters, optimal is to select the highest possible
+        ret: pd.Series = point["x"] + point["y"]
+        return ret
+
+    input_space = CS.ConfigurationSpace(seed=2169)
```
bpkroth, line 323: is that particular seed number important?

bpkroth: Elsewhere I think we have a `SEED` that you can reference.

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
```diff
@@ -298,3 +298,69 @@ def test_optimizer_type_defs(optimizer_class: Type[BaseOptimizer]) -> None:
...
+
+
+@pytest.mark.parametrize(('optimizer_type', 'kwargs'), [
+    # Default optimizer
+    (None, {}),
+    # Enumerate all supported Optimizers
+    *[(member, {}) for member in OptimizerType],
+    # Optimizer with non-empty kwargs argument
+])
+def test_mixed_input_space_types(optimizer_type: Optional[OptimizerType], kwargs: Optional[dict]) -> None:
```
bpkroth, line 310:
```suggestion
def test_mixed_numeric_type_input_space_types(optimizer_type: Optional[OptimizerType], kwargs: Optional[dict]) -> None:
```

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
```diff
@@ -298,3 +298,69 @@ def test_optimizer_type_defs(optimizer_class: Type[BaseOptimizer]) -> None:
...
+    """
+    Toy problem to test the optimizers.
```
bpkroth, line 312:
```suggestion
    Toy problem to test the optimizers with mixed numeric types to ensure that original dtypes are retained.
```

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
```diff
@@ -298,3 +298,69 @@ def test_optimizer_type_defs(optimizer_class: Type[BaseOptimizer]) -> None:
...
+        optimizer.get_observations()
+
+    for _ in range(max_iterations):
+        suggestion = optimizer.suggest()
+        assert isinstance(suggestion, pd.DataFrame)
+        assert (suggestion.columns == ['x', 'y']).all()
+        # Build suggestion mapping to cooperate with Configuration, note that
+        # doing a .iloc[0].to_dict() will cause pandas convert all numeric types
+        # to float64
+        tmp_suggest = {'x': suggestion['x'].values[0], 'y': suggestion['y'].values[0]}
```
bpkroth, line 353: A little concerned about this part.

Is the intent behind `tmp_suggest` for the test to avoid particular code paths that might do type conversions we don't want so that we can check that the types were correct when not doing that?  If so, can you amend the comment to document that a bit more?

I guess what I'm worried about is for callers to have to be concerned about these types of things.  We should probably try and handle it internally as much as possible.

In principle, I think it's probably an error if we can't immediately do `register` with the value returned from `suggest`.

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
```diff
@@ -298,3 +298,69 @@ def test_optimizer_type_defs(optimizer_class: Type[BaseOptimizer]) -> None:
...
+        # check that suggestion is in the space
+        configuration = CS.Configuration(optimizer.parameter_space, tmp_suggest)
```
bpkroth, line 355:
```suggestion
        test_configuration = CS.Configuration(optimizer.parameter_space, tmp_suggest)
```

bpkroth: Since `configuration` is not used in the call to `register` it might help explain what's going on a little better by naming it as `test_configuration` or `check_configuration` or `tmp_configuration` so that we can help be clear that we're just checking that the type conversions are being kept consistent.

bpkroth:
> Hm okay I have noticed that the [`llamatune_opt_test.py:49`](https://github.com/microsoft/MLOS/blob/main/mlos_bench/mlos_bench/tests/optimizers/llamatune_opt_test.py#L49-L65) failed. Looking into it, I see that some test in [`mlos_core_opt_df_test.py:45`](https://github.com/microsoft/MLOS/blob/100d56772e3458e0fd9573593b207704c7f8d1d2/mlos_bench/mlos_bench/tests/optimizers/mlos_core_opt_df_test.py#L45-L46) define parameters with special characters. For example `kernel_sched_migration_cost_ns!type`. Normally this column name is fine if we're converting from `pandas.DataFrame` to `dict` converting, but using a `NamedTuples` received from `itertuples()` will parse these columns strings with undefined behavior, which is why I am getting a test failure. I think we might have to use a different solution instead of `pandas.DataFrame.itertuples()`.

You could also try using `itertuples(name=None)` to get non-NamedTuples back.

bpkroth:
> Linter failed on [983763f](https://github.com/microsoft/MLOS/pull/667/commits/983763f96eaa4841926b59f2107be8f389e789b5). Fixed then reran linter locally, sorry for the spam. I think workflows should pass this time around.

No worries.  I generally suggest using the devcontainer and just locally running `make check` and `make test` when you're doing things.  It should run the same tests we did in CI then.

bpkroth:
Final thought: I think there's a few places in `mlos_bench` where we do a `to_numeric` that might need re-examine with this.

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
ephoris: No reason, saw that previous tests were manually defining seeds. `mlos_core` does not have a SEED defined anywhere, but I see [`mlos_bench/tests/__init__.py:40`](https://github.com/microsoft/MLOS/blob/main/mlos_bench/mlos_bench/tests/__init__.py#L40-L42) does. I see this repo likes `SEED = 42` as the default.

For the sake of isolation, I'll add `SEED = 42` to `mlos_core/tests/__init__.py` rather than importing SEED from `mlos_bench` to be used to reference in tests.

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
ephoris: Yes so if we take a look a previous test at [`optimizer_tests.py:93`](https://github.com/microsoft/MLOS/blob/main/mlos_core/mlos_core/tests/optimizers/optimizer_test.py#L93), this code path will fail if we utilize a mix of input types (namely floats and integers). This is due to the fact that `pd.DataFrame.iloc` will always promote data types. We can fix this by calling `.astype('O')` before any `iloc` call, but again that would mean the caller would have to be worried about this type casting.

For now I can remove `tmp_suggest` and instead replace the input with `suggestion.astype('O').iloc[0].to_dict()`

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
```diff
@@ -298,3 +298,68 @@ def test_optimizer_type_defs(optimizer_class: Type[BaseOptimizer]) -> None:
...
+        assert isinstance(suggestion, pd.DataFrame)
+        assert (suggestion.columns == ['x', 'y']).all()
+        # Check suggestion values are the expected dtype
+        assert isinstance(suggestion['x'].iloc[0], np.integer)
+        assert isinstance(suggestion['y'].iloc[0], np.floating)
+        # Check that suggestion is in the space
+        test_configuration = CS.Configuration(optimizer.parameter_space, suggestion.astype('O').iloc[0].to_dict())
+        # Raises an error if outside of configuration space
+        test_configuration.is_valid_configuration()
+        observation = objective(suggestion)
```
bpkroth, line 357:
```suggestion
        # Test registering the suggested configuration with a score.
        observation = objective(suggestion)
```

bpkroth (approved):
LGTM.  Thanks!

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
bpkroth: I like this new version a little better.  Consistent with the code used elsewhere as well.  Thanks!

bpkroth: As an aside, because of these weird behaviors of pandas, I'm not certain it was necessarily the right data structure to use as our APIs.  The original thought was that it was better from a data science user perspective, but mlos_bench later took a different tact of using dicts more frequently so we end up converting for the user quite a bit anyways.  Automatic type conversions behind the scenes based on your access method may be convenient in some circumstances, but not others.

I think at least in the common case, mlos_bench is the main user of mlos_core, so we can potentially mask some of it.

Anyways, just some backstory.  Very much appreciate you catching this.  Thanks!

bpkroth:
> Final thought: I think there's a few places in `mlos_bench` where we do a `to_numeric` that might need re-examine with this.

We can take this up elsewhere I think.

mlos_bench/mlos_bench/storage/util.py
```diff
@@ -30,7 +30,7 @@ def kv_df_to_dict(dataframe: pandas.DataFrame) -> Dict[str, Optional[TunableValu
         dataframe.rename(columns={'metric': 'parameter'}, inplace=True)
     assert dataframe.columns.tolist() == ['parameter', 'value']
     data = {}
-    for _, row in dataframe.iterrows():
+    for _, row in dataframe.astype('O').iterrows():
```
bpkroth, line 33: Hmm, may need to add a test for this in mlos_bench as well.

ephoris:
> > Hm okay I have noticed that the [`llamatune_opt_test.py:49`](https://github.com/microsoft/MLOS/blob/main/mlos_bench/mlos_bench/tests/optimizers/llamatune_opt_test.py#L49-L65) failed. Looking into it, I see that some test in [`mlos_core_opt_df_test.py:45`](https://github.com/microsoft/MLOS/blob/100d56772e3458e0fd9573593b207704c7f8d1d2/mlos_bench/mlos_bench/tests/optimizers/mlos_core_opt_df_test.py#L45-L46) define parameters with special characters. For example `kernel_sched_migration_cost_ns!type`. Normally this column name is fine if we're converting from `pandas.DataFrame` to `dict` converting, but using a `NamedTuples` received from `itertuples()` will parse these columns strings with undefined behavior, which is why I am getting a test failure. I think we might have to use a different solution instead of `pandas.DataFrame.itertuples()`.
> 
> You could also try using `itertuples(name=None)` to get non-NamedTuples back.

Unfortunately, itertuples preserves dtypes, but it also preserves `pandas.NA` values. Normally I would say this is good, however, `pandas.NA` value is not treated the same as `None` values. This results in situations where the `ConfigSpace` package will throw errors as it only handles `None` values. For now I guess `pandas.DataFrame.astype('O')` might be the most elegant solution for the time being as `iterrows()` will implicitly convert `pandas.NA` to `None` values which most of the code paths were set up to deal with.

mlos_bench/mlos_bench/storage/util.py
ephoris: Looks like this util function is mostly used in displaying meta-data in the `mlos_bench/storage` layer. I can sift through and see if there is a reasonable location for a `dtype` test here.

bpkroth: Eh, that's a little more complicated now that I look at it.  Let's take it as a separate item.  I think we can merge this one as is and circle back around to a proper test.  Thanks!

bpkroth: https://github.com/microsoft/MLOS/issues/669

PR #670, opened 2024-02-08
Add tests for mixed numerical type conversions of tunable config data from storage

### LINKED ISSUE #669: Add tests for kv_df_to_dict and mixed integer types
Per discussion in #667 we need a test for this line:

https://github.com/microsoft/MLOS/commit/9175d1885634d4a711a00918b617bbe351757e87#diff-8e173777b75fcdc7bdc22f9d4f8a60f2be2cc7265e2a2076add3c090cd25ce48R33

The overall codepath should be something like this:

1. Create Experiment with mixed numerical type Tunables.
2. suggest/register some data
3. Load the data in an ExperimentData object
4. Evaluate the types on the configs returned (e.g., are they still int/float as expected).

bpkroth:
Closes #669

### DIFF
--- a/doc/nginx-docker.sh
+++ b/doc/nginx-docker.sh
@@ -29,9 +29,9 @@ if [ "$cmd" == 'start' ]; then
     set -x
     tmpdir=$(mktemp -d)
     docker build --progress=plain -t mlos-doc-nginx \
-        --build-arg http_proxy=$http_proxy \
-        --build-arg https_proxy=$https_proxy \
-        --build-arg no_proxy=$no_proxy \
+        --build-arg http_proxy=${http_proxy:-} \
+        --build-arg https_proxy=${https_proxy:-} \
+        --build-arg no_proxy=${no_proxy:-} \
         --build-arg NGINX_PORT=$NGINX_PORT \
         -f Dockerfile "$tmpdir"
     rmdir "$tmpdir"
--- a/mlos_bench/mlos_bench/tests/storage/sql/fixtures.py
+++ b/mlos_bench/mlos_bench/tests/storage/sql/fixtures.py
@@ -78,7 +78,6 @@ def mixed_numerics_exp_storage(storage: SqlStorage, mixed_numerics_tunable_group
 
 
 def _dummy_run_exp(exp: SqlStorage.Experiment, tunable_name: str) -> SqlStorage.Experiment:
-    # pylint: disable=too-many-locals
     """
     Generates data by doing a simulated run of the given experiment.
     """
--- a/mlos_bench/mlos_bench/tests/storage/sql/fixtures.py
+++ b/mlos_bench/mlos_bench/tests/storage/sql/fixtures.py
@@ -17,6 +17,7 @@ from mlos_bench.storage.sql.storage import SqlStorage
 from mlos_bench.optimizers.mock_optimizer import MockOptimizer
 from mlos_bench.tunables.tunable_groups import TunableGroups
 
+from mlos_bench.tests import SEED
 from mlos_bench.tests.storage import CONFIG_COUNT, CONFIG_TRIAL_REPEAT_COUNT
 
 # pylint: disable=redefined-outer-name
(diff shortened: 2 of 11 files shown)


doc/nginx-docker.sh
```diff
@@ -29,9 +29,9 @@ if [ "$cmd" == 'start' ]; then
     set -x
     tmpdir=$(mktemp -d)
     docker build --progress=plain -t mlos-doc-nginx \
-        --build-arg http_proxy=$http_proxy \
-        --build-arg https_proxy=$https_proxy \
-        --build-arg no_proxy=$no_proxy \
+        --build-arg http_proxy=${http_proxy:-} \
+        --build-arg https_proxy=${https_proxy:-} \
+        --build-arg no_proxy=${no_proxy:-} \
```
bpkroth, line 34: Unrelated minor fixup to prevent errors on unbound variables when those haven't been set (or have been unset for testing).

mlos_bench/mlos_bench/tests/storage/sql/fixtures.py
```diff
@@ -99,7 +118,26 @@ def exp_storage_with_trials(exp_storage: SqlStorage.Experiment) -> SqlStorage.Ex
                 # And some influence from the tunable value.
                 "score": tunable_value_norm + random() / 100
             })
-    return exp_storage
+    return exp
+
+
+@pytest.fixture
+def exp_storage_with_trials(exp_storage: SqlStorage.Experiment) -> SqlStorage.Experiment:
+    # pylint: disable=too-many-locals
```
motus, line 126:
```suggestion
```
definitely don't need this one

mlos_bench/mlos_bench/tests/storage/sql/fixtures.py
```diff
@@ -57,24 +58,42 @@ def exp_storage(storage: SqlStorage, tunable_groups: TunableGroups) -> SqlStorag
...
+        description="pytest experiment",
+        tunables=mixed_numerics_tunable_groups,
+        opt_target=opt_target,
+        opt_direction=opt_direction,
+    ) as exp:
+        return exp
+
+
+def _dummy_run_exp(exp: SqlStorage.Experiment, tunable_name: str) -> SqlStorage.Experiment:
+    # pylint: disable=too-many-locals
```
motus, line 81:
```suggestion
```
can you check if we still need this pylint exception?

mlos_bench/mlos_bench/tests/storage/sql/fixtures.py
```diff
@@ -99,7 +118,26 @@ def exp_storage_with_trials(exp_storage: SqlStorage.Experiment) -> SqlStorage.Ex
...
+    """
+    Test fixture for Experiment using in-memory SQLite3 storage.
+    """
+    return _dummy_run_exp(exp_storage, tunable_name="kernel_sched_latency_ns")
+
+
+@pytest.fixture
+def mixed_numerics_exp_storage_with_trials(mixed_numerics_exp_storage: SqlStorage.Experiment) -> SqlStorage.Experiment:
+    # pylint: disable=too-many-locals
```
motus, line 135:
```suggestion
```
this one, too

motus (approved):
Looks good + a few nitpicks. 

P.S. I don't quite like that the `mixed_numerics_exp_storage` fixture returns an experiment detached from its context (i.e., after `Experiment._teardown()` has been called internally), but I guess that's ok for tests. I'll think about it later.

mlos_bench/mlos_bench/tests/storage/sql/fixtures.py
bpkroth: Was complaining for me locally in VSCode earlier, but yeah not on the CLI.

bpkroth:
> Looks good + a few nitpicks.
> 
> P.S. I don't quite like that the `mixed_numerics_exp_storage` fixture returns an experiment detached from its context (i.e., after `Experiment._teardown()` has been called internally), but I guess that's ok for tests. I'll think about it later.

I actually copied that one from your original `exp_storage` example.
I think the way that `pytest` handles that is that the context doesn't actually exit until the fixture is no longer in scope.
That pattern is also used for our docker based tests.

bpkroth:
> > Looks good + a few nitpicks.
> > P.S. I don't quite like that the `mixed_numerics_exp_storage` fixture returns an experiment detached from its context (i.e., after `Experiment._teardown()` has been called internally), but I guess that's ok for tests. I'll think about it later.
> 
> I actually copied that one from your original `exp_storage` example. I think the way that `pytest` handles that is that the context doesn't actually exit until the fixture is no longer in scope. That pattern is also used for our docker based tests.

Actually, we might need to use `yield` for that instead.  Let's take that up in the next PR.

motus (approved):

PR #671, opened 2024-02-09
Return status and actual timestamp from all environments

motus:
This is the first step to making trials asynchronous. In this PR we just make sure every Environment method that returns status also returns the timestamp of that status. In the next PR we will get these timestamps and status values from the remote environment.

### DIFF
--- a/mlos_bench/mlos_bench/environments/composite_env.py
+++ b/mlos_bench/mlos_bench/environments/composite_env.py
@@ -179,7 +179,7 @@ class CompositeEnv(Environment):
             env_context.teardown()
         super().teardown()
 
-    def run(self) -> Tuple[Status, Optional[Dict[str, TunableValue]]]:
+    def run(self) -> Tuple[Status, datetime, Optional[Dict[str, TunableValue]]]:
         """
         Submit a new experiment to the environment.
         Return the result of the *last* child environment if successful,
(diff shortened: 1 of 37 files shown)


mlos_bench/mlos_bench/environments/composite_env.py
```diff
@@ -179,61 +179,64 @@ def teardown(self) -> None:
...
+            (status, timestamp, telemetry) = env_context.status()
             _LOG.debug("Child env. status: %s :: %s", env_context, status)
             joint_telemetry.extend(telemetry)
             if not status.is_good() and final_status is None:
                 final_status = status
 
         final_status = final_status or status
         _LOG.info("Final status: %s :: %s", self, final_status)
-        return (final_status, joint_telemetry)
+        # Return the status and the timestamp of the last child environment.
```
bpkroth, line 241:
```suggestion
        # Return the status and the timestamp of the last child environment or the first failed child environment.
```

This seems to more closely match the logic of what's being returned, but I'm not sure it's always the most helpful.

Could be that we miss tracking the status of other failed environments except in the logs.

mlos_bench/mlos_bench/storage/sql/schema.py
```diff
@@ -144,6 +144,19 @@ def __init__(self, engine: Engine):
                                  [self.trial.c.exp_id, self.trial.c.trial_id]),
         )
 
+        self.trial_status = Table(
+            "trial_status",
+            self._meta,
+            Column("exp_id", String(255), nullable=False),
+            Column("trial_id", Integer, nullable=False),
+            Column("ts", DateTime, nullable=False, default="now"),
+            Column("status", String(16), nullable=False),
```
bpkroth, line 153: Maybe also `enviroment_name` or `environment_path` (e.g., include the parent's name as well to account for disambiguation since we don't enforce uniqueness)?

mlos_bench/mlos_bench/storage/sql/trial.py
```diff
@@ -80,21 +81,41 @@ def update(self, status: Status, timestamp: datetime,
...
+                    _LOG.warning("Record already exists: %s :: %s",
+                                 (metric_ts, metric_id, metric_value), ex)
+
+    def _update_status(self, conn: Connection, status: Status, timestamp: datetime) -> None:
+        """
+        Insert a new status record into the database.
+        This call is idempotent.
+        """
+        try:
+            conn.execute(self._schema.trial_status.insert().values(
```
bpkroth, line 113: Not seeing any tests for this new data source.
Also, what about retrieval?

bpkroth:
> Also in this PR, we store the statuses and timestamps in a new table that will later serve as a scheduling queue.

Not sure I agree with this, but maybe I just am not following where you're heading with it.

I think we need to store two separate things:
1. in progress status events of the trial (e.g., how different environments status change at different times).  I think that's valuable for debugging and other stateful reasons (e.g., which environments do we still need to call `teardown` or `cleanup` on).
2. the overall status of a trial (e.g., pending, in progress, aborting, succeeded, failed, etc.).  This could be used to implement a scheduling queue.

However, if implement this with a single table, then the latter requires messy aggregates over the former to find out the overall status of each current trial.

I think that should probably just be stored as a single column in the `trial` table and updated explicitly with logic inside `mlos_bench`.

Thoughts?

mlos_bench/mlos_bench/storage/sql/schema.py
motus: The plan is to use that table to store the status updates and also use it as a scheduling queue. That is, we'll have records with a status `SCHEDULED` + some future timestamp, and also add a record to this table each time when we poll the environment for status. I would rather keep this table's schema very minimal for that reason

motus: I've removed this change from the PR

mlos_bench/mlos_bench/storage/sql/trial.py
motus: this will be in the next PR - removing it from this one

motus:
> Thoughts?

I think we are on the same page. This is exactly how I want to implement this. I will remove all DB-related code from this PR and move it to the next one that actually implements the queue - i.e., with the corresponding read method, unit tests, etc.

bpkroth (approved):
LGTM

PR #676, opened 2024-02-14
Scheduling queue: Allow submitting trials with future timestamps and use watermarks to retrieve configs and results

motus:
* [x] `pending_trials` can now take timestamp and `running` flag to check for trials scheduled to run
* [x] `load` can now take the last trial id watermark to load only recent updates into the optimizer
* [x] `new_trial` now takes a (future) timestamp when scheduling a new trial
* [x] add a new table `trial_status` to track status polls and updates. (we use the existing `trial` table as a queue)
* [x] add more unit tests for the scheduling queue and status updates

### DIFF
--- a/mlos_bench/mlos_bench/run.py
+++ b/mlos_bench/mlos_bench/run.py
@@ -178,17 +178,14 @@ def _run(env: Environment, opt: Optimizer, trial: Storage.Trial, global_config: 
         opt.register(trial.tunables, Status.FAILED)
         return
 
-    (status, results) = env.run()  # Block and wait for the final result.
+    (status, timestamp, results) = env.run()  # Block and wait for the final result.
     _LOG.info("Results: %s :: %s\n%s", trial.tunables, status, results)
 
     # In async mode (TODO), poll the environment for status and telemetry
     # and update the storage with the intermediate results.
-    (_, telemetry) = env.status()
+    (_status, _timestamp, telemetry) = env.status()
 
-    # FIXME: Use the actual timestamp from the trial.
-    timestamp = datetime.utcnow()
-
-    # Use the status from `.run()` as it is the final status of the experiment.
+    # Use the status and timestamp from `.run()` as it is the final status of the experiment.
     # TODO: Use the `.status()` output in async mode.
     trial.update_telemetry(status, timestamp, telemetry)
 
--- a/mlos_bench/mlos_bench/run.py
+++ b/mlos_bench/mlos_bench/run.py
@@ -184,12 +184,15 @@ def _run(env: Environment, opt: Optimizer, trial: Storage.Trial, global_config: 
     # In async mode (TODO), poll the environment for status and telemetry
     # and update the storage with the intermediate results.
     (_, telemetry) = env.status()
+
+    # FIXME: Use the actual timestamp from the trial.
+    timestamp = datetime.utcnow()
+
     # Use the status from `.run()` as it is the final status of the experiment.
     # TODO: Use the `.status()` output in async mode.
-    trial.update_telemetry(status, telemetry)
+    trial.update_telemetry(status, timestamp, telemetry)
 
-    # FIXME: Use the actual timestamp from the benchmark.
-    trial.update(status, datetime.utcnow(), results)
+    trial.update(status, timestamp, results)
     # Filter out non-numeric scores from the optimizer.
     scores = results if not isinstance(results, dict) \
         else {k: float(v) for (k, v) in results.items() if isinstance(v, (int, float))}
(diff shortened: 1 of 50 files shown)


mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -256,21 +256,50 @@ def load_telemetry(self, trial_id: int) -> List[Tuple[datetime, str, Any]]:
...
+                Tunable values, benchmark scores, and status of the trials.
             """
 
         @abstractmethod
-        def pending_trials(self) -> Iterator['Storage.Trial']:
+        def pending_trials(self, timestamp: datetime, *, running: bool) -> Iterator['Storage.Trial']:
             """
-            Return an iterator over the pending trial runs for this experiment.
+            Return an iterator over the pending trials that are scheduled to run
+            on or before the specified timestamp.
```
bpkroth, line 285: ❤️

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -279,6 +308,8 @@ def new_trial(self, tunables: TunableGroups,
             ----------
             tunables : TunableGroups
                 Tunable parameters to use for the trial.
+            ts_start : Optional[datetime]
+                Timestamp of the trial start (can be in the future).
```
bpkroth, line 312: Yes!

mlos_bench/mlos_bench/storage/sql/experiment.py
```diff
@@ -122,7 +122,8 @@ def load_telemetry(self, trial_id: int) -> List[Tuple[datetime, str, Any]]:
             return [(row.ts, row.metric_id, row.metric_value)
                     for row in cur_telemetry.fetchall()]
 
-    def load(self, opt_target: Optional[str] = None) -> Tuple[List[dict], List[Optional[float]], List[Status]]:
+    def load(self, last_trial_id: int = -1,
```
bpkroth, line 125:
```suggestion
    def load(self,
             last_trial_id: int = -1,
```
nit: when breaking multiple params across lines, I find it easier to read when `self` or maybe `self, *,` are by themselves, and all others are on their own line.

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -256,21 +256,50 @@ def load_telemetry(self, trial_id: int) -> List[Tuple[datetime, str, Any]]:
             """
 
         @abstractmethod
-        def load(self, opt_target: Optional[str] = None) -> Tuple[List[dict], List[Optional[float]], List[Status]]:
+        def load(self, last_trial_id: int = -1,
```
bpkroth, line 259:
```suggestion
        def load(self,
                 last_trial_id: int = -1,
```

mlos_bench/mlos_bench/tests/storage/exp_load_test.py
```diff
@@ -127,7 +130,7 @@ def test_exp_trial_pending_3(exp_storage: Storage.Experiment,
     trial_fail.update(Status.FAILED, datetime.utcnow())
     trial_succ.update(Status.SUCCEEDED, datetime.utcnow(), score)
```
bpkroth, line 131: Do we have a test where we start one trial from the pending set, update it's status midway, and then check that in the main table it's still registered as "RUNNING"?

bpkroth (approved):
Few tests I think we could expand on, but LGTM.  Thank you!  I'm excited to finally have this started :)

bpkroth:
Next request: add a storage API to allow a user to enqueue a new trial for an experiment from a notebook :)

mlos_bench/mlos_bench/tests/storage/exp_load_test.py
motus: not yet. I am adding more unit tests now for cases like this one.

mlos_bench/mlos_bench/run.py
```diff
@@ -108,7 +108,7 @@ def _optimize(*,
             (configs, scores, status) = exp.load()
             opt_context.bulk_register(configs, scores, status)
             # Complete any pending trials.
-            for trial in exp.pending_trials():
+            for trial in exp.pending_trials(datetime.utcnow(), running=True):
```
bpkroth, line 111: Future work: instead of simple loop, also support a class that implements a schedule (order) on the available pending trials.

That class should be swappable/configurable, to allow us to explore different scheduling policies.

motus: yes - for starters, I was thinking about an "ordering" functor that would look simply like this:
```python
for trial in scheduling_policy(exp.pending_trials(datetime.utcnow(), running=True)):
    # ...
```

PR #678, opened 2024-02-14
use named constants to define columns containing IDs and generic values

### LINKED ISSUE #677: `objectives.optimization_target` column is too long to be in the primary key for MySQL
From @eujing :
> during CREATE TABLE that sqlalchemy issues, it cant allocate the column VARCHAR(1024) because primary keys internally cannot be more than 3072 bytes

motus:
Also, make `objectives.optimization_target` use the new `_LEN_ID` constant to make sure it is not too long to be used as a primary key in MySQL (issue #677)

Closes #677

### DIFF
--- a/mlos_bench/mlos_bench/storage/sql/schema.py
+++ b/mlos_bench/mlos_bench/storage/sql/schema.py
@@ -17,10 +17,6 @@ from sqlalchemy import (
 
 _LOG = logging.getLogger(__name__)
 
-# This class is internal to SqlStorage and is mostly a struct
-# for all DB tables, so it's ok to disable the warnings.
-# pylint: disable=too-many-instance-attributes
-
 
 class _DDL:
     """
@@ -45,11 +41,15 @@ class DbSchema:
     """
     A class to define and create the DB schema.
     """
+    # This class is internal to SqlStorage and is mostly a struct
+    # for all DB tables, so it's ok to disable the warnings.
+    # pylint: disable=too-many-instance-attributes
 
-    # Default string column sizes.
+    # Common string column sizes.
     _ID_LEN = 512
     _PARAM_VALUE_LEN = 1024
     _METRIC_VALUE_LEN = 255
+    _STATUS_LEN = 16
 
     def __init__(self, engine: Engine):
         """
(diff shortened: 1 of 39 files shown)


mlos_bench/mlos_bench/storage/sql/schema.py
```diff
@@ -46,6 +46,9 @@ class DbSchema:
     A class to define and create the DB schema.
     """
 
+    _LEN_ID = 255
```
bpkroth, line 49: Would prefer we revert back to the original lengths for now and have an override specifically for when MySQL or MariaDB are specified as the engine in `__init__` so we don't mess up existing ones or unnecessarily limit the length on certain items.

mlos_bench/mlos_bench/storage/sql/schema.py
```diff
@@ -46,6 +46,9 @@ class DbSchema:
...
+    _LEN_VALUE = 255
```
bpkroth, line 50: Additionally, does this need to be so short even for non-indexed value columns?  Or do we need to change the type on some of them to `TEXT`?

mlos_bench/mlos_bench/storage/sql/schema.py
motus: the original length *was* 255 - except for the `optimization_target`, which was 1024. I'd just set it to 512 and call it a day rather than creating exceptions for certain DBs

mlos_bench/mlos_bench/storage/sql/schema.py
motus: 255 is what we have now for both IDs and values. Let's have 512/1024 instead and move on. We can always review it later. Right now I just want to unblock @eujing

motus:
bumped up the ID and value lengths to 512 and 1024 correspondingly. That's an increase from our current 255/255 limit while still allowing us to use optimization_target in a primary key on MySQL

mlos_bench/mlos_bench/storage/sql/schema.py
bpkroth: Gave an alternative method suggestion here:
https://github.com/motus/MLOS/pull/11

motus: merged, thank you! I like that we have separate constants for parameters and metrics. I've added another one for status - we'll have another table with status soon, so it will come in handy.

P.S. We don't need to override the values for MySQL, though - as long as the IDs' lengths are 512 or less, we are fine.

bpkroth: Do we want to keep that logic in there for now anyways.  Or maybe just a comment in the __init__ so that we know how to override it in the future?  I can imagine this being a thing we run into again with different storage backends.

motus: I'd much prefer to keep the logic DB-agnostic and have `_ID_LEN` at some reasonable value that can work with any DB engine. Overriding logic that you propose is a bit more flexible, but it would prevent the users from migrating their existing data from, e.g., SQLite to MySQL

bpkroth: I suppose that's true, but we also haven't tested every DB that sqlalchemy supports.
If some new one came along and required even shorter column lengths, would you propose we alter existing schemas to accommodate that (#649)? That'd potentially lose data silently but be hard to special case which schema modifications we automate and which we don't.
Flexible length seems like the more sustainable future to me. It means we can start with a large column length. If it's too large for a new DB we encounter, it's immediately known and we can fix it with an override without affecting any existing installs. If someone is migrating from one backend to another that is an explicitly manual operation (we don't manage that for them automatically), hence they're on the hook for any data loss due to column length differences.

bpkroth: We discussed offline and decided to move forward as is.
In case it's necessary, we can pull this commit up again in the future:
https://github.com/motus/MLOS/pull/11/commits/2589bf2c24e067a72b91ad9bdb73f5e360a7ad8c

bpkroth (approved):

PR #683, opened 2024-02-19
Address pandas timestamp warnings

bpkroth:
Addressing some warnings from pandas about datetime conversions and deprecated and future strict behavior.

One issue we currently have is that a timestamp produced by a client may be in a different timezone or lack timezone info from what that receiver parses before injecting it into the DB.

This starts to at least attempt to work on those conversions for consistency.

Currently also includes changes from #682 and #680

### DIFF
--- a/mlos_bench/mlos_bench/environments/local/local_env.py
+++ b/mlos_bench/mlos_bench/environments/local/local_env.py
@@ -217,7 +217,7 @@ class LocalEnv(ScriptEnv):
         return data
 
 
-    _MIN_TS = datetime(2000, 1, 1, 0, 0, 0, tzinfo=pytz.UTC)
+    _MIN_TS = datetime(2024, 1, 1, 0, 0, 0, tzinfo=pytz.UTC)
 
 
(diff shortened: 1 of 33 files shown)


mlos_bench/mlos_bench/environments/local/local_env.py
```diff
@@ -215,6 +216,37 @@ def _normalize_columns(data: pandas.DataFrame) -> pandas.DataFrame:
             data.rename(str.rstrip, axis='columns', inplace=True)
         return data
 
+    # All timestamps in the telemetry data must be greater than this date
+    # (a very rough approximation for the start of this feature).
+    _MIN_TS = datetime(2024, 1, 1, 0, 0, 0, tzinfo=pytz.UTC)
+
+    @staticmethod
+    def _datetime_parser(datetime_col: pandas.Series) -> pandas.Series:
```
bpkroth, line 224: Could potentially move this and `_normalize_columns` to `mlos_bench.util` in the future.

Issue #679, opened 2024-02-19
Expand scheduler to support additional policies


bpkroth:
# 676 adds initial support for storing pending trials in the database and retrieving them.

Right now, no scheduling policy is enforced (other than perhaps shortest deadline first, based on backend storage retrieval ordering).

We should expand on that support to allow pluggable/configurable alternative scheduling (ordering) policies (e.g., gradually assign more workers to validate config variance in what looks like a feasible region).
 
>               Future work: instead of simple loop, also support a class that implements a schedule (order) on the available pending trials.
> 
> That class should be swappable/configurable, to allow us to explore different scheduling policies.

_Originally posted by @bpkroth in https://github.com/microsoft/MLOS/pull/676#discussion_r1494834676_

Issue #427, opened 2023-06-29
Implement more flexible "Stopping Conditions"


bpkroth:
Options:
- max_cost
- max_time
- max_iterations (currently available)
- "converaged"
  - e.g., max_change over M < max_iterations steps

PR #684, opened 2024-02-23
Initial work to separate the scheduler and the optimizer (no async yet)

motus:
Initial work to support pluggable Schedulers (#679).
This starts by adding the necessary support to auxiliary classes and splits the current simple run/optimize loop into two stages.
Future PRs will expand on this to separate abstract and configurable classes for more flexible scheduling policies.

### DIFF
--- a/mlos_bench/mlos_bench/optimizers/base_optimizer.py
+++ b/mlos_bench/mlos_bench/optimizers/base_optimizer.py
@@ -195,7 +195,7 @@ class Optimizer(metaclass=ABCMeta):     # pylint: disable=too-many-instance-attr
 
     @abstractmethod
     def bulk_register(self, configs: Sequence[dict], scores: Sequence[Optional[float]],
-                      status: Optional[Sequence[Status]] = None, is_warm_up: bool = True) -> bool:
+                      status: Optional[Sequence[Status]] = None, is_warm_up: bool = False) -> bool:
         """
         Pre-load the optimizer with the bulk data from previous experiments.
 
--- a/mlos_bench/mlos_bench/optimizers/base_optimizer.py
+++ b/mlos_bench/mlos_bench/optimizers/base_optimizer.py
@@ -195,7 +195,7 @@ class Optimizer(metaclass=ABCMeta):     # pylint: disable=too-many-instance-attr
 
     @abstractmethod
     def bulk_register(self, configs: Sequence[dict], scores: Sequence[Optional[float]],
-                      status: Optional[Sequence[Status]] = None) -> bool:
+                      status: Optional[Sequence[Status]] = None, is_warm_up: bool = True) -> bool:
         """
         Pre-load the optimizer with the bulk data from previous experiments.
 
@@ -207,13 +207,16 @@ class Optimizer(metaclass=ABCMeta):     # pylint: disable=too-many-instance-attr
             Benchmark results from experiments that correspond to `configs`.
         status : Optional[Sequence[float]]
             Status of the experiments that correspond to `configs`.
+        is_warm_up : bool
+            True for the initial load, False for subsequent calls.
 
         Returns
         -------
         is_not_empty : bool
             True if there is data to register, false otherwise.
         """
-        _LOG.info("Warm-up the optimizer with: %d configs, %d scores, %d status values",
+        _LOG.info("%s the optimizer with: %d configs, %d scores, %d status values",
+                  "Warm-up" if is_warm_up else "Load",
                   len(configs or []), len(scores or []), len(status or []))
         if len(configs or []) != len(scores or []):
             raise ValueError("Numbers of configs and scores do not match.")
--- a/mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
+++ b/mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
@@ -99,7 +99,7 @@ class MlosCoreOptimizer(Optimizer):
         return f"{self.__class__.__name__}:{self._opt.__class__.__name__}"
 
     def bulk_register(self, configs: Sequence[dict], scores: Sequence[Optional[float]],
-                      status: Optional[Sequence[Status]] = None, is_warm_up: bool = True) -> bool:
+                      status: Optional[Sequence[Status]] = None, is_warm_up: bool = False) -> bool:
         if not super().bulk_register(configs, scores, status, is_warm_up):
             return False
         df_configs = self._to_df(configs)  # Impute missing values, if necessary
--- a/mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
+++ b/mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
@@ -111,6 +111,8 @@ class MlosCoreOptimizer(Optimizer):
             df_configs = df_configs[df_status_completed]
             df_scores = df_scores[df_status_completed]
         self._opt.register(df_configs, df_scores)
+        if not is_warm_up:
+            self._iter += len(df_scores)
         if _LOG.isEnabledFor(logging.DEBUG):
             (score, _) = self.get_best_observation()
             _LOG.debug("Warm-up end: %s = %s", self.target, score)
--- a/mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
+++ b/mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
@@ -99,8 +99,8 @@ class MlosCoreOptimizer(Optimizer):
         return f"{self.__class__.__name__}:{self._opt.__class__.__name__}"
 
     def bulk_register(self, configs: Sequence[dict], scores: Sequence[Optional[float]],
-                      status: Optional[Sequence[Status]] = None) -> bool:
-        if not super().bulk_register(configs, scores, status):
+                      status: Optional[Sequence[Status]] = None, is_warm_up: bool = True) -> bool:
+        if not super().bulk_register(configs, scores, status, is_warm_up):
             return False
         df_configs = self._to_df(configs)  # Impute missing values, if necessary
         df_scores = pd.Series(scores, dtype=float) * self._opt_sign
--- a/mlos_bench/mlos_bench/run.py
+++ b/mlos_bench/mlos_bench/run.py
@@ -99,7 +99,7 @@ def _optimization_loop(*,
             # Complete trials that are pending or in-progress.
             _scheduler(exp, env_context, global_config, running=True)
             # Load past trials data into the optimizer
-            _optimizer(exp, opt_context, is_warm_up=True)
+            last_trial_id = _optimizer(exp, opt_context, is_warm_up=True)
         else:
             _LOG.warning("Skip pending trials and warm-up: %s", opt)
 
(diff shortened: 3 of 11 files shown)


mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
```diff
@@ -111,6 +111,8 @@ def bulk_register(self, configs: Sequence[dict], scores: Sequence[Optional[float
             df_configs = df_configs[df_status_completed]
             df_scores = df_scores[df_status_completed]
         self._opt.register(df_configs, df_scores)
+        if not is_warm_up:
+            self._iter += len(df_scores)
```
bpkroth, line 115: Does `bulk_register()` get called outside of the warmup procedure?  Or is that for resuming an experiment?

mlos_bench/mlos_bench/run.py
```diff
@@ -101,50 +94,26 @@ def _optimize(*,
...
-            # to impute the missing tunable values.
-            (configs, scores, status) = exp.load()
-            opt_context.bulk_register(configs, scores, status)
-            # Complete any pending trials.
-            for trial in exp.pending_trials(datetime.utcnow(), running=True):
-                _run(env_context, opt_context, trial, global_config)
+            # Complete trials that are pending or in-progress.
+            _scheduler(exp, env_context, global_config, running=True)
+            # Load past trials data into the optimizer
+            last_trial_id = _optimizer(exp, opt_context, is_warm_up=True)
```
bpkroth, line 102: According to the `is_warm_up` flag, this won't increment the `_iter` counter, right?

mlos_bench/mlos_bench/run.py
```diff
@@ -154,29 +123,96 @@ def _optimize(*,
...
+    tunable_values = exp.load_tunable_config(config_id)
+    tunables = env_context.tunable_params.assign(tunable_values)
+    _LOG.info("Load config from storage: %d", config_id)
+    if _LOG.isEnabledFor(logging.DEBUG):
+        _LOG.debug("Config %d ::\n%s",
+                   config_id, json.dumps(tunable_values, indent=2))
+    return tunables
+
+
+def _scheduler(exp: Storage.Experiment, env_context: Environment,
```
bpkroth, line 140: Let's make a new abstract and simple concrete class for this.
That will allow us to swap out different scheduling policies in the future.

bpkroth, line 140:
```suggestion
def _run_schedule(exp: Storage.Experiment, env_context: Environment,
```

Though, I think this should probably be put in the "Scheduler" class I mentioned below.

mlos_bench/mlos_bench/run.py
```diff
@@ -154,29 +123,96 @@ def _optimize(*,
...
+               global_config: Dict[str, Any], running: bool = False) -> None:
+    """
+    Scheduler part of the loop. Check for pending trials in the queue and run them.
+    """
+    for trial in exp.pending_trials(datetime.utcnow(), running=running):
+        _run_trial(env_context, trial, global_config)
+
+
+def _optimizer(exp: Storage.Experiment, opt_context: Optimizer,
```
bpkroth, line 149:
```suggestion
def _get_optimizer_suggestions(exp: Storage.Experiment, opt_context: Optimizer,
```

Am thinking that actual scheduling, should be left up to the Scheduler class, and the Optimizer should only be in charge of suggesting new configs.

Now, especially when it comes to multi-objective and/or parallel trials, it might be worthwhile for there to be some sort of feedback between the two that allows, for instance, the Scheduler to accept additional information like cost/budget/uncertainty/priority/weight in from the Optimizer, but I think actually determining which of the subset of options the Optimizer has presented should be run next, should be part of the Scheduler.

bpkroth: Also, if the Scheduler is swappable, it's possible that the config changes mid run, and a new Scheduler policy gets used after some other ones were already enqueued to the pending trials queue.
Not entirely certain how those might interact.  Maybe scheduling only happens on entry to the queue, and on popping items off the queue at any given iteration, we can reorder with the current policy if there are any ties (e.g., equal weights and start times or some such), but otherwise not have much smart logic for that.  i.e., it should mostly be admission control policies.

Thoughts?

mlos_bench/mlos_bench/run.py
```diff
@@ -154,29 +123,96 @@ def _optimize(*,
...
+            # between resuming the experiment (since that is not currently
+            # prevented).
+            # TODO: Improve for supporting multi-objective
+            # (e.g., opt_target_1, opt_target_2, ... and opt_direction_1, opt_direction_2, ...)
+            "optimizer": opt.name,
+            "opt_target": opt.target,
+            "opt_direction": opt.direction,
+            "repeat_i": repeat_i,
+            "is_defaults": tunables.is_defaults,
+        })
```
bpkroth, line 187: Here's a great example.  In this case, a smart scheduling policy might want to consider spreading these repeats out over time.

mlos_bench/mlos_bench/run.py
```diff
@@ -154,29 +123,96 @@ def _optimize(*,
...
+def _optimizer(exp: Storage.Experiment, opt_context: Optimizer,
+               last_trial_id: int = -1, trial_config_repeat_count: int = 1,
+               is_warm_up: bool = False) -> int:
+    """
+    Optimizer part of the loop. Load the results of the executed trials
+    into the optimizer, suggest new configurations, and add them to the queue.
+    Return the last trial ID processed by the optimizer.
+    """
+    (trial_ids, configs, scores, status) = exp.load(last_trial_id)
+    opt_context.bulk_register(configs, scores, status, is_warm_up)
```
bpkroth, line 158: Another possibility here:
Perhaps we don't just bulk register, but entirely retrain an optimizer with the latest data on each round.

For instance, not all the optimizers we have support an internal notion of "noise", so we can't always just feed the `[(config_a, score_1), (config_a, score_2,), ...]` to the optimizer, we might actually need to feed `[(config_a, aggregate_score)]` in, which we won't have until after the fact, which means *replacing* prior Optimizer models, not just augmenting them.

mlos_bench/mlos_bench/run.py
```diff
@@ -154,29 +123,96 @@ def _optimize(*,
...
+
+    tunables = opt_context.suggest()
+    _schedule_trial(exp, opt_context, tunables, trial_config_repeat_count)
+
+    return max(trial_ids, default=last_trial_id)
+
+
+def _schedule_trial(exp: Storage.Experiment, opt: Optimizer,
```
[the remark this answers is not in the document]
motus: makes sense. let me think about it. I'll push a new commit soon

mlos_bench/mlos_bench/run.py
motus: agree. Let me fix that

mlos_bench/mlos_bench/run.py
motus: good point, let me do that

mlos_bench/mlos_bench/run.py
motus: I think this is more of an optimizer policy than the queue one. Regardless of the optimizer, we warm it up with the prior data, and then feed more results as we get them. It is up to the optimizer to decide whether and how to retain this data and produce the results. In fact, BO uses *all* data points every time it produces a suggestion. Let's keep it simple in the optimization loop and add that retraining logic to the optimizer when we need it (and, again, not all optimizers will need it anyway)

mlos_bench/mlos_bench/run.py
motus: no, it shouldn't - this is a warm-up phase, and we don't want to increment the counter here. at least, this is the logic that we've had before.

mlos_bench/mlos_bench/run.py
motus: yes, I was not exactly happy with the naming of my functions, either 🙂 - thanks for the suggestions!

My plan is to implement the swappable Scheduler class in the next PR. The purpose of this one is to disentangle the trial execution and config recommendation parts and see what other components have to be updated to accommodate this change. This way, we keep the PRs small, and the next PR won't have updates to the storage, optimizer, etc.

mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
bpkroth: That will be pretty dependent on timing though, right?
We may want some notion of a group commit to wait for others to come in.
For instance, config_a score comes in a 10 seconds, config_b score comes in at 10.1 seconds, but we've already collected the first one and started retraining.  We could wait for a full batch to finish, or just some timeline (e.g., 500ms) for a collection to come in before retraining and starting a new one.

Anyways ... future work.

mlos_bench/mlos_bench/run.py
bpkroth: Yeah, just means that whenever you resume, the max_iterations is constantly moving forward.
For instance, then if you run 90 iterations, kill the script, and then start it again until completion, you will run for 190 total iterations, rather than just 10 more.
Not sure which is more correct atm.
Maybe we can take that up more formally in #427

mlos_bench/mlos_bench/run.py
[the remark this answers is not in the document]
bpkroth: We discussed pushing this to a future PR to keep things cleaner.

mlos_bench/mlos_bench/run.py
bpkroth: In principle I agree, but BO doesn't always do that - it depends on the implementation of the library.  If you look at the output for some of these libraries, they claim to *replace* the `(config_a, score_x)` value instead of augmenting the dataset.

That's what makes me think we might want to be able to manage that for ourselves on occasion, but I think you're right - it's not a Scheduler problem.  Rather it's an interaction between the Optimizer and the Storage class (assuming we don't want to keep all historical records in memory all the time).

mlos_bench/mlos_bench/tests/launcher_in_process_test.py
```diff
@@ -0,0 +1,43 @@
+#
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+#
+"""
+Unit tests to check the launcher and the main optimization loop in-process.
```
bpkroth, line 6: Was the intention for this to be able to attach to the process for debugging?
I think we might have an example of that already ...

bpkroth: Nm, I'd used something like this in a few cases where the Windows VSCode pytest integration wasn't allowing be to do the usual "debug test".

```python
# rest of the test file ...

if __name__ == '__main__':
    pytest.main([__file__, "-n0"])
```

Then you can just debug the whole file as it's own executable.  Not sure if that's relevant here though.

bpkroth (approved):
LGTM

Looking forward to the expanded Scheduler classes

mlos_bench/mlos_bench/optimizers/base_optimizer.py
```diff
@@ -195,7 +195,7 @@ def supports_preload(self) -> bool:
 
     @abstractmethod
     def bulk_register(self, configs: Sequence[dict], scores: Sequence[Optional[float]],
-                      status: Optional[Sequence[Status]] = None) -> bool:
+                      status: Optional[Sequence[Status]] = None, is_warm_up: bool = True) -> bool:
```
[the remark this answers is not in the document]
motus: agree! I just wanted to minimize the change to the existing code, and by default this method is used for warm-up in all our unit tests. I think I'll implement your suggestion in this PR, though, so we don't have to touch this class later

PR #686, opened 2024-02-23
Allow no tunable configs for benchmarking

### LINKED ISSUE #613: Storage schema doesn't allow an empty config
In some cases its useful to test deployment and other scripts without a config (i.e., vnet deployment with no tunable params).

Currently some combination of the storage schema and the way we serialize the config representation to hash it to create a config id prevents us from doing this.

It would be nice if we handled that as an "empty" config somehow as well.

bpkroth:
Closes #613

- [ ] add tests for data retrieval

### DIFF
--- a/mlos_bench/mlos_bench/tunables/tunable_groups.py
+++ b/mlos_bench/mlos_bench/tunables/tunable_groups.py
@@ -36,6 +36,9 @@ class TunableGroups:
         for (name, group_config) in config.items():
             self._add_group(CovariantTunableGroup(name, group_config))
 
+    def __bool__(self) -> bool:
+        return bool(self._index)
+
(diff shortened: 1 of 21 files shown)


mlos_bench/mlos_bench/tunables/tunable_groups.py
```diff
@@ -36,6 +36,9 @@ def __init__(self, config: Optional[dict] = None):
         for (name, group_config) in config.items():
             self._add_group(CovariantTunableGroup(name, group_config))
 
+    def __bool__(self) -> bool:
+        return bool(self._index)
```
bpkroth, line 40: I removed the code that was using this, but still might be useful.

motus (approved):

PR #690, opened 2024-02-26
Add GridSearch Optimizer support

### LINKED ISSUE #688: mlos_bench: grid search support
In benchmarking only mode, sometimes we actually do want to use grid search to evaluate all possibilities for comparison sake.

This can also be useful when comparing more efficient optimizer search strategy performance as well.

We should be able to implement this as an additional mlos_bench optimizer in combination with the storage backend to check which ones have already been evaluated.

bpkroth:
Closes #688

- Introduces `GridSearchOptimizer` to `mlos_bench`
  - Generates and stores a set of `tuple(dict.values())` from `ConfigSpace` to track elements of the config grid to search
  - If `max_iterations` > `len(grid)` can refill the grid if desired (e.g., by calling `suggest()` after `not_converged()` returns `False`.
  - If `max_iterations` < `len(grid)` (i.e., we don't have enough iterations to complete the grid) will issue a warning.


mlos_bench/mlos_bench/tunables/tunable.py
```diff
@@ -546,19 +562,73 @@ def range(self) -> Union[Tuple[int, int], Tuple[float, float]]:
         assert self._range is not None
         return self._range
 
+    @property
+    def span(self) -> Union[int, float]:
```
motus, line 566: I would rather remove the `.span`, `.quantized_values`, and `.values` properties from the `Tunable`. They are not used anywhere outside of the grid search, and the `Tunable` class has enough accessors to implement that functionality in the optimizer. In general, if something is not used in the scheduler/environment, it should not be part of the `Tunable`.

P.S. I understand that by that logic we should have implemented the distributions/quantization/log stuff outside of the `Tunable`, too - e.g., in some sort of constraints on the optimizer side. But in case of the distributions it was a question of the user convenience - i.e., having all info about the Tunable in one place in JSON. Here I would much rather keep the Tunable intact and move the required logic to teh optimzier

mlos_bench/mlos_bench/tunables/tunable.py
```diff
@@ -9,14 +9,29 @@
...
+TunableValueType = Union[Type[int], Type[float], Type[str]]
+
+"""
+Tunable value type tuple.
+For checking with isinstance()
+"""
+TunableValueTypeTuple = (int, float, str, type(None))
+
+"""The string name of a tunable value type."""
+TunableValueTypeName = Literal["int", "float", "categorical"]
```
motus, line 32: let's make these type-related changes a separate PR. I'll be happy to merge it before the main one

mlos_bench/mlos_bench/storage/util.py
```diff
@@ -32,7 +32,7 @@ def kv_df_to_dict(dataframe: pandas.DataFrame) -> Dict[str, Optional[TunableValu
     data = {}
     for _, row in dataframe.astype('O').iterrows():
         assert isinstance(row['parameter'], str)
-        assert row['value'] is None or isinstance(row['value'], (str, int, float))
+        assert isinstance(row['value'], TunableValueTypeTuple)
```
motus, line 35: ❤️ 
make it a separate PR, though

mlos_bench/mlos_bench/tunables/covariant_group.py
```diff
@@ -179,15 +180,15 @@ def get_names(self) -> Iterable[str]:
...
         """
         Get current values of all tunables in the group as a dict.
 
         Returns
         -------
-        tunables : Dict[str, TunableValue]
+        tunables : HashableTunableValuesDict
         """
-        return {name: tunable.value for (name, tunable) in self._tunables.items()}
+        return HashableTunableValuesDict({name: tunable.value for (name, tunable) in self._tunables.items()})
```
motus, line 191: I would rather not expose the data structure that exists exclusively for the convenience of the grid search optimizer here in the Tunables or anywhere on the scheduler side. I am not even sure we need that `HashableTunableValuesDict` structure at all - how about we just keep track of the tuples of tunable values in the optimizer? I'll comment on that in detail at `GridSearchOptimizer`
Alternatively, just make `TunableGroups` hashable

mlos_bench/mlos_bench/optimizers/grid_search_optimizer.py
```diff
@@ -0,0 +1,151 @@
...
+                 score: Optional[Union[float, dict]] = None) -> Optional[float]:
+        registered_score = super().register(tunables, status, score)
+        if status.is_succeeded() and (
+            self._best_score is None or (registered_score is not None and registered_score < self._best_score)
+        ):
+            self._best_score = registered_score
+            self._best_config = tunables.copy()
+        self._iter += 1
+        try:
+            self._suggested_configs.remove(tunables.get_param_values())
```
motus, line 142: How about we just keep track of the *tuples of values* instead of introducing a special data structure? I mean, something like
```suggestion
            self._suggested_configs.remove(tuple(tunables.get_param_values().values()))
```
(if you wish, we can also add some sanity check to make sure that the keys of the tunables match those we've passed to the optimizer in the constructor, but I think it is not necessary)

motus (commented):
Left a few comments. To summarize:

* let's have a separate PR with type-related fixes to the Tunable and related classes
* remove .span/.values/.quantized_values from Tunable and move that logic to the optimizer
* I am sure we can do without HashableTunableValuesDict class. I would suggest to track the tuples of tunable values in the optimizer instead of using the dict (should use less memory, too)

mlos_bench/mlos_bench/optimizers/grid_search_optimizer.py
bpkroth: Not a bad idea.  I'll poke at it.

mlos_bench/mlos_bench/storage/util.py
bpkroth: Yep, on the list.

mlos_bench/mlos_bench/tunables/covariant_group.py
bpkroth: Yeah, I was waffling on this too.

I think part of the issue is that we are somewhat inconsistent about dicts vs. TunableGroups vs. DataFrames.

I think both of the options you proposed are good though, so I'll play with it.

mlos_bench/mlos_bench/tunables/tunable.py
bpkroth: Not sure I follow the critique here.  These are basically just introspection properties on the Tunable, and feel pretty natural to me to keep here.

We could export them as utility functions that take a Tunable instead, but why?

mlos_bench/mlos_bench/environments/base_environment.py
```diff
@@ -289,9 +289,10 @@ def _combine_tunables(self, tunables: TunableGroups) -> Dict[str, TunableValue]:
         params : Dict[str, Union[int, float, str]]
             Free-format dictionary that contains the new environment configuration.
         """
-        return tunables.get_param_values(
-            group_names=list(self._tunable_params.get_covariant_group_names()),
-            into_params=self._const_args.copy())
+        params = self._const_args.copy()
+        params.update(tunables.get_param_values(
+            group_names=self._tunable_params.get_covariant_group_names()))
+        return params
```
bpkroth, line 295: TODO: revert this change - no longer necessary (though this was the only call for that method and might be worth dropping anyways)

mlos_bench/mlos_bench/tunables/covariant_group.py
bpkroth: An issue with making TunableGroups, CovariantGroups, Tunable hashable is that they are all mutable so one could do something like this:

```python
configs = set()
tunable_groups = TunableGroup(...)
configs.add(tunable_groups)
tunable_groups.assign(new_values)
assert len(configs) == 1
assert tunable_groups in configs  # would fail, since the hash now no longer matches
```

I think from a data structure perspective this is wrong.

The way I'd written `HashableTunableValuesDict` it was immutable, so this couldn't happen.

I can rework things internally to just use `tuples` of values, but it does make for some inefficiencies (e.g., having to reconstruct `dicts` from those repeatedly), but that might be less important.

Anyways, just leaving a note about this for now.

mlos_bench/mlos_bench/tunables/tunable.py
bpkroth: #696

mlos_bench/mlos_bench/tunables/tunable.py
bpkroth: Moved these to #698, but I think we should keep them.

mlos_bench/mlos_bench/storage/util.py
bpkroth: #696

mlos_bench/mlos_bench/tunables/covariant_group.py
bpkroth: FWIW, using `tuples` is also sensitive to the order they arrive into the `dict` in, which is different between `ConfigSpace` and `TunableGroups`

mlos_bench/mlos_bench/optimizers/grid_search_optimizer.py
bpkroth: Comments on this in the other thread below.

bpkroth: The short version is this is what I did, but you have to be careful about what the source of the dict is since the order of the values in the tuple matters.

mlos_bench/mlos_bench/optimizers/grid_search_optimizer.py
```diff
@@ -0,0 +1,190 @@
...
+            status = [Status.SUCCEEDED] * len(configs)
+        for (params, score, trial_status) in zip(configs, scores, status):
+            tunables = self._tunables.copy().assign(params)
+            self.register(tunables, trial_status, None if score is None else float(score))
+            if is_warm_up:
+                # Do not advance the iteration counter during warm-up.
+                self._iter -= 1
+        if _LOG.isEnabledFor(logging.DEBUG):
+            (score, _) = self.get_best_observation()
+            _LOG.debug("Warm-up end: %s = %s", self.target, score)
```
motus, line 127:
```suggestion
            _LOG.debug("%s end: %s = %s", "Warm-up" if is-warm_up else "Update", self.target, score)
```

mlos_bench/mlos_bench/optimizers/grid_search_optimizer.py
```diff
@@ -0,0 +1,190 @@
...
+        except KeyError:
+            _LOG.warning("Attempted to remove missing config (previously registered?) from suggested set: %s", tunables)
+        return registered_score
+
+    def not_converged(self) -> bool:
+        if self._iter > self._max_iter:
+            if bool(self._pending_configs):
+                _LOG.warning("Exceeded max iterations, but still have %d pending configs: %s",
+                             len(self._pending_configs), list(self._pending_configs.keys()))
+        return self._iter <= self._max_iter and bool(self._pending_configs)
```
motus, line 184: maybe
```suggestion
        if self._iter > self._max_iter:
            if bool(self._pending_configs):
                _LOG.warning("Exceeded max iterations, but still have %d pending configs: %s",
                             len(self._pending_configs), list(self._pending_configs.keys()))
            return False
        return bool(self._pending_configs)
```
say, if I want to run only 5 iterations of the larger grid search

BTW, why `bool()`? Is that something mypy wants?

mlos_bench/mlos_bench/tests/optimizers/grid_search_optimizer_test.py
```diff
@@ -0,0 +1,246 @@
...
+        "float": 0.75,
+    }
+    grid_search_opt_pending_configs = list(grid_search_opt.pending_configs)
+    assert expected_config_example in grid_search_tunables_grid
+    assert expected_config_example in grid_search_opt_pending_configs
+    # Check the rest of the contents.
+    # Note: ConfigSpace param name vs TunableGroup parameter name order is not
+    # consistent, so we need to full dict comparison.
+    assert len(grid_search_opt_pending_configs) == expected_grid_size
+    assert all(config in grid_search_tunables_grid for config in grid_search_opt_pending_configs)
```
motus, line 113: maybe just
```suggestion
    assert grid_search_opt_pending_configs.issubset(grid_search_tunables_grid)
```
(also applies to many other places below in this file)

mlos_bench/mlos_bench/tests/optimizers/grid_search_optimizer_test.py
```diff
@@ -0,0 +1,246 @@
...
+    expected_grid_size = math.prod(tunable.cardinality for tunable, _group in grid_search_tunables)
+    assert expected_grid_size > len(grid_search_tunables)
+    assert len(grid_search_tunables_grid) == expected_grid_size
+    # Check for specific example configs inclusion.
+    expected_config_example: Dict[str, TunableValue] = {
+        "cat": "a",
+        "int": 2,
+        "float": 0.75,
+    }
+    grid_search_opt_pending_configs = list(grid_search_opt.pending_configs)
```
motus, line 106:
```suggestion
    grid_search_opt_pending_configs = set(grid_search_opt.pending_configs)
```
if we want to check for subsets/supersets

mlos_bench/mlos_bench/tests/optimizers/grid_search_optimizer_test.py
```diff
@@ -0,0 +1,246 @@
...
+    Make sure that grid search optimizer works correctly when suggest and register
+    are called out of order.
+    """
+    score = 1.0
+    status = Status.SUCCEEDED
+    suggest_count = 10
+    suggested = [grid_search_opt.suggest() for _ in range(suggest_count)]
+    suggested_shuffled = suggested.copy()
+    random.shuffle(suggested_shuffled)
+    assert suggested != suggested_shuffled
```
motus, line 204:
```suggestion
```
there's > 0 probability this condition can be false 😃

motus (commented):
Left a few comments. overall, looks good, thank you!

mlos_bench/mlos_bench/optimizers/grid_search_optimizer.py
bpkroth: re `bool()` yeah I think it was a type thing, but I can double check.

The other logic I think covers your case as well btw - it's even checked in the tests, but it may not be as readable.

Should evaluate to `return False and bool(self._pending_configs) => False`

mlos_bench/mlos_bench/tests/optimizers/grid_search_optimizer_test.py
bpkroth: The aren't sets unfortunately, so this doesn't work.
This btw, was a little why I'd implemente the `HashableTunableValuesDict` class.

mlos_bench/mlos_bench/tests/optimizers/grid_search_optimizer_test.py
bpkroth: I want it to be true though.  I added a retry loop for now.

mlos_bench/mlos_bench/optimizers/grid_search_optimizer.py
```diff
@@ -0,0 +1,188 @@
...
+            del self._pending_configs[default_config_values]
+            self._suggested_configs.add(default_config_values)
+        else:
+            # Select the first item from the pending configs.
+            if not self._pending_configs and self._iter <= self._max_iter:
+                _LOG.info("No more pending configs to suggest. Restarting grid.")
+                self._config_keys, self._pending_configs = self._get_grid()
+            try:
+                next_config_values = next(iter(self._pending_configs.keys()))
+            except StopIteration as exc:
```
motus, line 150: maybe
```suggestion
                next_config_values = self._pending_configs.popitem()[0]
            except KeyError as exc:
```
and remove that `del` below?

motus (approved):
Looks good!

PR #691, opened 2024-02-28
Pluggable Scheduler implementation

motus:
Initial implementation of a separate `Scheduler` class from the core `run` loop.


bpkroth (approved):
LGTM, I can adapt what I had started for parallel workers pretty nicely for this.  Thanks!

Issue #380, opened 2023-06-01
parallel trial execution


bpkroth:
Would like to be able to execute multiple experiment trials in parallel

bpkroth:
Possible method would be to use the async + background event loop thread technique demonstrated in #510

bpkroth:
Related: #463

bpkroth:
A few thoughts on design possibilities for this:
- [ ] We need some sort of "pending trial configs" abstraction.
  - [ ] A simple way to implement this I think would be to provide a small class that interfaces with the storage backend and looks up pending trial/configs based on their status.
    - Note that that would support the ability to inject new trial/configs manually, repeats (e.g., for noisy handling confidence), or to mix ensembles of different "suggestion" modes (e.g., SMAC, FLAML, LLM, etc.) since it can separate the optimization method from the trial/config scheduling (see below).
- [ ] We need a notion of a "worker" (terminology to be determined).
  - [ ] At a high level, this could be a forked child process that manages running one trial to completion and returning the results to the orchestrator (i.e., the optimization loop in `run.py`)
  - [ ] top-level mlos_bench cli config will be needed to determine how many of these slots there should be active (e.g., max 10 parallel trials).
- [ ] Trial configs would need to be able to make use of a new built in variable (e.g., `$worker_instance_id`) in order to template out VM names, for instance, so there are no overlaps between parallel trial instances.
- [ ] We need to a "scheduler" component, that matches a pending trial/config to a worker.
  - This should support pluggable policies so that when a worker becomes idle (returns a status from a finished execution) we can determine which of the pending trials to assign to it vs. generate a new suggestion
    - This might need additional configs like max pending trials or max suggestions or some such.  See Also #427 for relation to improved stopping conditions specifications.
  - [ ] Additionally, for restart and error handling, we will need to store which worker_id was assigned which trial in the backend storage.
- [ ] The run.py optimization loop will need to be adjusted so that the main process now acts purely as an orchestrator.  Something like multiprocessing pool could work well for this.

bpkroth:
Some of this has been implemented in #720.

Still todo are the parallel execution scheduler using forked children.

PR #693, opened 2024-02-29
Split `Environment.run()` into two (async) parts - WIP

motus:
Everything is broken atm

### DIFF
--- a/mlos_bench/mlos_bench/environments/base_environment.py
+++ b/mlos_bench/mlos_bench/environments/base_environment.py
@@ -116,9 +116,11 @@ class Environment(metaclass=abc.ABCMeta):
         self.config = config
         self._service = service
         self._service_context: Optional[Service] = None
-        self._is_ready = False
         self._in_context = False
         self._const_args: Dict[str, TunableValue] = config.get("const_args", {})
+
+        self._status = Status.PENDING
+        self._timestamp = datetime.utcnow()  # Timestamp of the last status update
         self._results: Optional[Dict[str, TunableValue]] = None
 
         if _LOG.isEnabledFor(logging.DEBUG):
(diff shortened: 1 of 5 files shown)


bpkroth:
High level design though: why split run at all?

Why not keep the current idea: setup(), run(), teardown() are all synchronous.
That would make it easier to keep the semantics for CompositeEnvs as well, otherwise we need to start expressing which parts of those can be parallelized or not.

Instead, I propose making a new "Worker" concept.

Worker can be assigned an Environment to take through the entire setup, run, teardown phase.

Then, the Scheduler, can have a pool of Workers, that it runs asynchronously, e.g., in separate processes or threads.

See Also #380 for some design note details.

bpkroth:
> High level design though: why split run at all?
> 
> Why not keep the current idea: setup(), run(), teardown() are all synchronous. That would make it easier to keep the semantics for CompositeEnvs as well, otherwise we need to start expressing which parts of those can be parallelized or not.
> 
> Instead, I propose making a new "Worker" concept.
> 
> Worker can be assigned an Environment to take through the entire setup, run, teardown phase.
> 
> Then, the Scheduler, can have a pool of Workers, that it runs asynchronously, e.g., in separate processes or threads.
> 
> See Also #380 for some design note details.

FWIW, I have some async status polling work started that builds off of this idea:

TrialRunner (Worker) class walks an Environment thru the setup/run/teardown phases.
It also gets an EventContextLoop to run background tasks with.
The base `run` phase is augmented to also start a background async status poll using that EventContextLoop.

I'll try and post a PR when I have something working a little better end-to-end.

motus:
Abandoning in favor of multithreaded wrapper around a synchronous Environment class. Might resume that work later

PR #696, opened 2024-03-04
Typing tweaks

bpkroth:
Some type hint and alias tweaks for Tunables and a few others.

Split off from #690.

### DIFF
--- a/mlos_bench/mlos_bench/storage/util.py
+++ b/mlos_bench/mlos_bench/storage/util.py
@@ -10,7 +10,7 @@ from typing import Dict, Optional
 
 import pandas
 
-from mlos_bench.tunables.tunable import TunableValue
+from mlos_bench.tunables.tunable import TunableValue, TunableValueTypeTuple
 from mlos_bench.util import try_parse_val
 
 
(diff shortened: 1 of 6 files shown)


mlos_bench/mlos_bench/storage/util.py
```diff
@@ -32,7 +32,7 @@ def kv_df_to_dict(dataframe: pandas.DataFrame) -> Dict[str, Optional[TunableValu
     data = {}
     for _, row in dataframe.astype('O').iterrows():
         assert isinstance(row['parameter'], str)
-        assert row['value'] is None or isinstance(row['value'], (str, int, float))
+        assert isinstance(row['value'], TunableValueTypeTuple)
```
motus, line 35: Now when I look at it, I think we should throw an exception instead of using an `assert`. We should use the assertions for sanity checks of our own code; in this case, we check the validity of the external data (that can be produced by the user's script, i.e., not by our code)
```suggestion
        if not isinstance(row['value'], TunableValueTypeTuple):
            raise TypeError(f"Invalid column type: value:{type[row'value']}")
```

motus (approved):
Looks good! Left one small comment, but I'm ambivalent on this one

PR #697, opened 2024-03-04
Move config_space to base Optimizer property

bpkroth:
Also fix some missing conversion issues.

Split out from #690.


motus (approved):
Looks good!

PR #707, opened 2024-03-08
A function for a popular `None if value is None else func(value)` code pattern

motus:
basically, a poor man's Maybe monad


bpkroth (approved):
Nice!

PR #705, opened 2024-03-07
Conda Updates

bpkroth:
- Update the base image and make sure related dependencies are consistently from either pip or conda, but not both (follow on fixes to #699)
- Minor tweak to include the PATH in the local script exec environment variables
- Another fix to the CI issue in #703

PR #708, opened 2024-03-08
Roll back forceful assignment of PATH when invoking a local process

motus:
Roll back one of the updates from #705 
It breaks tests on both Windows and WSL DevContainer on my PC.
* On Windows, `environ` may not have `PATH` at all
* On my WSL DevContainer, passing non-null `env` to `subprocess.run` prevents it from finding the right python interpreter

### DIFF
--- a/mlos_bench/mlos_bench/services/local/local_exec.py
+++ b/mlos_bench/mlos_bench/services/local/local_exec.py
@@ -195,10 +195,8 @@ class LocalExecService(TempDirContextService, SupportsLocalExec):
         cmd = [token for subcmd in subcmds for token in subcmd]
 
         env: Dict[str, str] = {}
-        # Need to include at least some basic environment variables to run the script.
-        env["PATH"] = environ["PATH"]
         if env_params:
-            env.update({key: str(val) for (key, val) in env_params.items()})
+            env = {key: str(val) for (key, val) in env_params.items()}
 
         if sys.platform == 'win32':
             # A hack to run Python on Windows with env variables set:


mlos_bench/mlos_bench/services/local/local_exec.py
```diff
@@ -195,10 +195,8 @@ def _local_exec_script(self, script_line: str,
         cmd = [token for subcmd in subcmds for token in subcmd]
 
         env: Dict[str, str] = {}
-        # Need to include at least some basic environment variables to run the script.
-        env["PATH"] = environ["PATH"]
         if env_params:
-            env.update({key: str(val) for (key, val) in env_params.items()})
+            env = {key: str(val) for (key, val) in env_params.items()}
```
bpkroth, line 199: Can you give an example of where this breaks?

Without this, things like this can fail when it needs to reference a python environment (set via the `PATH`) with `pandas` installed, for instance.

```python
#!/usr/bin/env python
import pandas
print("Hello World!")
```

bpkroth (approved):
Approving to unblock, but I think we'll need to revisit (see comments in the PR).

mlos_bench/mlos_bench/services/local/local_exec.py
motus: currently it breaks in all my environments, i.e. Windows and WSL/DevContainer on both of my PCs. On Windows, `PATH` is not in `environ` at all, and all unit tests with local_exec crash with `KeyError`.

In my DevContainers, only `launcher_run_test.py` tests fail, and it looks like the local_exec gets incorrect path or other environment variables when we launch `run.py` from inside the unit tests.

Rolling back this change fixes both issues.

bpkroth:
> On Windows, environ may not have PATH at all

That seems like a bug.  `PATH` is used to resolve command locations on both platforms.

> On my WSL DevContainer, passing non-null env to subprocess.run prevents it from finding the right python interpreter

Oh that's interesting.  That's exactly the situation I was trying to fix, but in the opposite direction.  Do you have an example?

PR #710, opened 2024-03-09
Pluggable Scheduler

### LINKED ISSUE #700: Config support for pluggable Schedulers
Follow up work to #691

motus:
* [x] Make Scheduler class loadable from JSON configs
* [x] Update the Launcher and `run.py` to instantiate Scheduler from JSON
* [x] Create JSON schema for the Scheduler config
* [x] Add unit tests for the new Scheduler JSON configs

Closes #700

### DIFF
--- a/mlos_bench/mlos_bench/config/schemas/schedulers/scheduler-schema.json
+++ b/mlos_bench/mlos_bench/config/schemas/schedulers/scheduler-schema.json
@@ -13,13 +13,13 @@
             "type": "object",
             "properties": {
                 "experiment_id": {
-                    "$ref": "./common-defs-subschemas.json#/$defs/experiment_id"
+                    "$ref": "../cli/common-defs-subschemas.json#/$defs/experiment_id"
                 },
                 "trial_id": {
-                    "$ref": "./common-defs-subschemas.json#/$defs/trial_id"
+                    "$ref": "../cli/common-defs-subschemas.json#/$defs/trial_id"
                 },
                 "config_id": {
-                    "$ref": "./common-defs-subschemas.json#/$defs/config_id"
+                    "$ref": "../cli/common-defs-subschemas.json#/$defs/config_id"
                 },
                 "teardown": {
                     "description": "Whether to teardown the experiment after running it.",
--- a/mlos_bench/mlos_bench/config/schemas/schedulers/scheduler-schema.json
+++ b/mlos_bench/mlos_bench/config/schemas/schedulers/scheduler-schema.json
@@ -1,0 +1,100 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://raw.githubusercontent.com/microsoft/MLOS/main/mlos_bench/mlos_bench/config/schemas/schedulers/scheduler-schema.json",
+    "title": "mlos_bench Scheduler config",
+
+    "$defs": {
+        "comment": {
+            "$comment": "This section contains reusable partial schema bits (or just split out for readability)"
+        },
+
+        "config_base_scheduler": {
+            "$comment": "config properties common to all Scheduler types.",
+            "type": "object",
+            "properties": {
+                "experiment_id": {
+                    "$ref": "./common-defs-subschemas.json#/$defs/experiment_id"
+                },
+                "trial_id": {
+                    "$ref": "./common-defs-subschemas.json#/$defs/trial_id"
+                },
+                "config_id": {
+                    "$ref": "./common-defs-subschemas.json#/$defs/config_id"
+                },
+                "teardown": {
+                    "description": "Whether to teardown the experiment after running it.",
+                    "type": "boolean"
+                },
+                "trial_config_repeat_count": {
+                    "description": "Number of times to repeat a config.",
+                    "type": "integer",
+                    "minimum": 1,
+                    "examples": [3, 5]
+                }
+            }
+        }
+    },
+
+    "description": "config for the mlos_bench scheduler",
+    "$comment": "top level schema document rules",
+    "type": "object",
+    "properties": {
+        "$schema": {
+            "description": "The schema to use for validating the scheduler config (accepts both URLs and local paths).",
+            "type": "string",
+            "$comment": "This is optional, but if provided, should match the name of this file.",
+            "pattern": "/schemas/schedulers/scheduler-schema.json$"
+        },
+
+        "description": {
+            "description": "Optional description of the config.",
+            "type": "string"
+        },
+
+        "class": {
+            "description": "The name of the scheduler class to use.",
+            "$comment": "required",
+            "enum": [
+                "mlos_bench.schedulers.SyncScheduler",
+                "mlos_bench.schedulers.sync_scheduler.SyncScheduler"
+            ]
+        },
+
+        "config": {
+            "description": "The scheduler-specific config.",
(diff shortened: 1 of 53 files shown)


mlos_bench/mlos_bench/config/schemas/schedulers/scheduler-schema.json
```diff
@@ -0,0 +1,100 @@
...
+    "$defs": {
+        "comment": {
+            "$comment": "This section contains reusable partial schema bits (or just split out for readability)"
+        },
+
+        "config_base_scheduler": {
+            "$comment": "config properties common to all Scheduler types.",
+            "type": "object",
+            "properties": {
+                "experiment_id": {
```
bpkroth, line 15: Why is experiment_id being repeated in scheduler?
Isn't that an expected global or cli variable that's fed into other things?

Same for config_id, trial_id, etc.

Guess I think of those as controlling a run of a given instance of `mlos_bench` whereas configs for the scheduler have to do with how it schedules *new* config (e.g., how many repeats, which order, which workers, etc.).  And that in general we would have reusable scheduler configs that don't include any of those details.

Or is this just to support the possibility of one-off overrides to influence a single instance run to only consider a single pre-existing config or only invoke a single pre-existing trial (both of which are only relevant under an existing experiment_id which we would also need to reference)?

mlos_bench/mlos_bench/launcher.py
```diff
@@ -146,6 +146,8 @@ def __init__(self, description: str, long_text: str = "", argv: Optional[List[st
         _LOG.info("Init storage: %s", self.storage)
 
         self.teardown: bool = bool(args.teardown) if args.teardown is not None else bool(config.get("teardown", True))
```
bpkroth, line 148: Should `self.teardown` also become `global_config['teardown']`?

I'm a little wary about moving some of these things into `global_config`.

`trial_id` makes sense, but does `trial_config_repeat_config`?
When would a user script need to access that value (e.g., via `shell_env_params`)?

Also, we should document somewhere the set of "known" variables that go into `global_config` so that 
a) users don't conflict with them
b) they know what they can use in their scripts

mlos_bench/mlos_bench/launcher.py
```diff
@@ -203,9 +205,14 @@ def _parse_args(parser: argparse.ArgumentParser, argv: Optional[List[str]]) -> T
                  ' a single trial with default (or specified in --tunable_values).')
 
         parser.add_argument(
-            '--trial_config_repeat_count', '--trial-config-repeat-count', required=False, type=int, default=1,
```
bpkroth, line 206: Dropping `default=1` because now the default can be enforced in the chosen `Scheduler` class?

mlos_bench/mlos_bench/launcher.py
```diff
@@ -203,9 +205,14 @@ def _parse_args(parser: argparse.ArgumentParser, argv: Optional[List[str]]) -> T
...
+            '--trial_config_repeat_count', '--trial-config-repeat-count', required=False, type=int,
             help='Number of times to repeat each config. Default is 1 trial per config, though more may be advised.')
 
+        parser.add_argument(
+            '--scheduler', required=False,
+            help='Path to the scheduler configuration file. By default, use' +
+                 ' a single-threaded synchronous scheduler.')
```
bpkroth, line 214:
```suggestion
                 ' a single worker synchronous scheduler.')
```

bpkroth: Basically avoid threaded vs. process distinctions here - it's not relevant to them.

mlos_bench/mlos_bench/launcher.py
```diff
@@ -337,17 +344,13 @@ def _load_optimizer(self, args_optimizer: Optional[str]) -> Optimizer:
         in the --optimizer command line option. If config file not specified,
         create a one-shot optimizer to run a single benchmark trial.
         """
-        if 'max_iterations' in self.global_config:
-            self.global_config['max_iterations'] *= self.trial_config_repeat_count
```
bpkroth, line 341: Discussed offline:

- previously (#642) we artificially increased the iteration count so that when we repeated trials it wouldn't lessen the number of configs we sampled
- now that we're separating the optimizer (suggest config) and scheduler (repeat of configs), the we can move the tracking of number of configs suggested to the `suggest()` routine in the optimizer and `number of trials executed` to the scheduler
- that opens up the door for more flexible stopping conditions (#427)
- to do that, we should also cleanup our terminology and rename `max_iterations` to `num_config_suggestions` and add a new `max_trials` argument (separate PR though I think)

mlos_bench/mlos_bench/schedulers/base_scheduler.py
```diff
@@ -42,14 +42,34 @@ def __init__(self, *,
...
+            The optimizer to use.
+        storage : Storage
+            The storage to use.
+        root_env_config : str
+            Path to the root environment configuration.
         """
         self.global_config = global_config
-        config = merge_parameters(dest=config.copy(), source=global_config)
+        config = merge_parameters(dest=config.copy(), source=global_config,
+                                  required_keys=["experiment_id", "trial_id"])
```
bpkroth, line 63: Why is `trial_id` required?

mlos_bench/mlos_bench/launcher.py
motus: that, too - but the main reason is that since command-line options have priority over JSON config settings, having default value here means that we always override values in configs with 1 when command-line option is not specified :)

mlos_bench/mlos_bench/launcher.py
motus: technically, it should, but if you add `teardown` to global config, it will override the `teardown` property of all `ScriptEnv` environments :)

motus: but in all, yes, that's more of a backward compatibility fix so we can keep using e.g. '--trial-config-repeat-count` option instead of `--trial_config_repeat_count` (with underscores)

mlos_bench/mlos_bench/config/schemas/schedulers/scheduler-schema.json
motus: All these parameters (except of the `trial_config_repeat_count`, probably) are mostly placeholders for properties that we need, e.g., for storage initialization - but also, yes, to make sure they are present so that other components have them, too. I like the idea of one-off overrides - I guess it should work this way. I'll add another PR with that scenario + unit tests

bpkroth: > but also, yes, to make sure they are present so that other components have them, too

I don't think the need to be here for that to happen, but if we're going to implement the override option as well, it might make sense to keep them.

One general trend I see is that sometimes, because a parameter flows from, globals, overring the cli, and also appears in scheduler, and the launcher, and ...
it just gets hard to track what's *supposed* to be authoritative.

If we instead include it only in certain places and then flow thru only in the code where necessary it might be more manageable, from a config writing/debugging standpoint for users.

mlos_bench/mlos_bench/launcher.py
bpkroth: Ah, makes sense.  Thanks!

bpkroth (approved):
LGTM.  Thanks!

PR #713, opened 2024-03-15
Rename `max_iterations` to `max_suggestions` and track in Optimizer `.suggest()` instead of `.register()`

### LINKED ISSUE #711: Increment iteration counter in `.suggest()` instead of `.register()`
That fix simplifies warming-up the optimizer and allows running multiple trials per suggestion without adjusting `max_iterations` parameter

motus:
makes optimizers and schedulers a bit simpler. Part of issue #715 

Closes #711 

Note: the move from `--max_iterations` to `--max_suggestions` is a breaking change, so we will need to cut a new release for this.

### DIFF
--- a/mlos_bench/README.md
+++ b/mlos_bench/README.md
@@ -195,10 +195,12 @@ Searching for an optimal set of tunable parameters is very similar to running a 
 All we have to do is specifying the [`Optimizer`](./mlos_bench/optimizers/) in the top-level configuration, like in our [`azure-redis-opt.jsonc`](./mlos_bench/config/cli/azure-redis-opt.jsonc) example.
 
 ```sh
-mlos_bench --config "./mlos_bench/mlos_bench/config/cli/azure-redis-opt.jsonc" --globals "experiment_MyBenchmark.jsonc" --max_suggestions 10
+mlos_bench --config "./mlos_bench/mlos_bench/config/cli/azure-redis-opt.jsonc" --globals "experiment_MyBenchmark.jsonc" --max_suggestions 10 --trial-config-repeat-count 3
 ```
 
-Note that again we use the command line option `--max_suggestions` to override the default value from [`mlos_core_flaml.jsonc`](./mlos_bench/config/optimizers/mlos_core_flaml.jsonc).
+Note that again we use the command line option `--max_suggestions` to override the max. number of suggested configurations to trial from [`mlos_core_flaml.jsonc`](./mlos_bench/config/optimizers/mlos_core_flaml.jsonc).
+We also use `--trial-config-repeat-count` to benchmark each suggested configuration 3 times.
+That means, we will run 30 trials in total, 3 for each of the 10 suggested configurations.
 
 We don't have to specify the `"tunable_values"` for the optimization: the optimizer will suggest new values on each iteration and the framework will feed this data into the benchmarking environment.
 
--- a/mlos_bench/README.md
+++ b/mlos_bench/README.md
@@ -195,10 +195,10 @@ Searching for an optimal set of tunable parameters is very similar to running a 
 All we have to do is specifying the [`Optimizer`](./mlos_bench/optimizers/) in the top-level configuration, like in our [`azure-redis-opt.jsonc`](./mlos_bench/config/cli/azure-redis-opt.jsonc) example.
 
 ```sh
-mlos_bench --config "./mlos_bench/mlos_bench/config/cli/azure-redis-opt.jsonc" --globals "experiment_MyBenchmark.jsonc" --max_iterations 10
+mlos_bench --config "./mlos_bench/mlos_bench/config/cli/azure-redis-opt.jsonc" --globals "experiment_MyBenchmark.jsonc" --max_suggestions 10
 ```
 
-Note that again we use the command line option `--max_iterations` to override the default value from [`mlos_core_flaml.jsonc`](./mlos_bench/config/optimizers/mlos_core_flaml.jsonc).
+Note that again we use the command line option `--max_suggestions` to override the default value from [`mlos_core_flaml.jsonc`](./mlos_bench/config/optimizers/mlos_core_flaml.jsonc).
 
 We don't have to specify the `"tunable_values"` for the optimization: the optimizer will suggest new values on each iteration and the framework will feed this data into the benchmarking environment.
 
--- a/mlos_bench/mlos_bench/config/schemas/optimizers/optimizer-schema.json
+++ b/mlos_bench/mlos_bench/config/schemas/optimizers/optimizer-schema.json
@@ -22,8 +22,8 @@
                     "enum": ["min", "max"],
                     "example": "min"
                 },
-                "max_iterations": {
-                    "description": "The maximum number of additional (in the case of merging experiment data or resuming experiments) iterations to run when we launch the app.",
+                "max_suggestions": {
+                    "description": "The maximum number of additional (in the case of merging experiment data or resuming experiments) suggestions/trials to run when we launch the app.",
                     "type": "integer",
                     "minimum": 0,
                     "example": 100
--- a/mlos_bench/mlos_bench/config/schemas/optimizers/optimizer-schema.json
+++ b/mlos_bench/mlos_bench/config/schemas/optimizers/optimizer-schema.json
@@ -23,7 +23,7 @@
                     "example": "min"
                 },
                 "max_suggestions": {
-                    "description": "The maximum number of additional (in the case of merging experiment data or resuming experiments) config suggestions to run when we launch the app.  Note: configs may be repeated in more than one trial.",
+                    "description": "The maximum number of additional (in the case of merging experiment data or resuming experiments) config suggestions to run when we launch the app, or no limit if 0 is provided. Note: configs may be repeated in more than one trial.",
                     "type": "integer",
                     "minimum": 0,
                     "example": 100
--- a/mlos_bench/mlos_bench/config/schemas/optimizers/optimizer-schema.json
+++ b/mlos_bench/mlos_bench/config/schemas/optimizers/optimizer-schema.json
@@ -23,7 +23,7 @@
                     "example": "min"
                 },
                 "max_suggestions": {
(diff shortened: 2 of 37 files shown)


mlos_bench/mlos_bench/optimizers/base_optimizer.py
```diff
@@ -269,13 +267,15 @@ def suggest(self) -> TunableGroups:
             These are the same tunables we pass to the constructor,
             but with the values set to the next suggestion.
         """
+        self._iter += 1
```
bpkroth, line 270: Here's another alternative:
1. In `suggest`, place the config in a `pending` set.
  - As an extension on this, we could also internally loop on `_suggest` until we received a config not in the `pending` set, but I'd leave that for future work to explore.
1. In `register`, check to see if the provided configs are in the `pending` set.
  - If they are, remove them from the set and increment the iteration count.
  - Finish registering the config-scores regardless.

This allows the caller to "throw out" some "bad" configs (which they may filter with their own logic) but avoid those configs counting against them.

We have an API that's sort of for this in `mlos_core`: `register_pending()`
The basic idea was to allow us some means of avoiding the optimizer repeatedly suggesting the same config to us if we had some other means of identifying it as invalid or otherwise "pending" (e.g., doing parallel trials with different configs).

Obviously in the perfect world we'd also provide feedback to the optimizer about a config being no good (e.g., register it as 3x worst real config-score seen so far), but this allows us still a sane middle ground approach.

mlos_bench/mlos_bench/tests/launcher_run_test.py
```diff
@@ -93,8 +93,7 @@ def test_launch_main_app_opt(root_path: str, local_exec_service: LocalExecServic
     """
     _launch_main_app(
         root_path, local_exec_service,
-        # TODO: Reset --max_iterations to 3 after fixing the optimizer
-        "--config mlos_bench/mlos_bench/tests/config/cli/mock-opt.jsonc --trial_config_repeat_count 3 --max_iterations 9",
+        "--config mlos_bench/mlos_bench/tests/config/cli/mock-opt.jsonc --trial_config_repeat_count 3 --max_iterations 3",
```
bpkroth, line 96: Can we also use this PR to rename "max_iterations" to "num_of_suggested_configs" or "max_num_configs" or something like that?
#715

motus: sure, let's do that

mlos_bench/mlos_bench/optimizers/base_optimizer.py
motus: yeah, we can do that, but I don't think this extra functionality belongs to the base optimizer class. I would rather implement it in some derived class to keep the base interface simpler and also give the user more choice when implementing/instantiating the optimizers.

P.S. On a side note, I would be very careful when tracking the suggested/pending configurations. Exact matching might not work - e.g., consider two configurations that differ by 1e-6 in one parameter. Ideally, the user should be able to pick some similarity measure and closeness boundary for their configurations.

P.P.S. Same applies for termination criteria - e.g., we might want to stop the optimization not only when the acquisition function is small on the entire range, but also when, say, teh last 10 suggested configs are all close to the already tried configs, etc., etc.

mlos_bench/mlos_bench/tests/launcher_run_test.py
bpkroth: We already did the breaking change in the last one with how `--max-iterations` no longer is multiplied.  I'd say we finish it and cut a new release now.  Or we could finish this one and do it in the next PR.  Thoughts?

mlos_bench/mlos_bench/optimizers/base_optimizer.py
bpkroth: Termination criteria I think could be implemented separately, but your point about the first one is somewhat well taken.  However, part of me thinks that most of our configs are not continuous so that particular scenario would only show up except for floating point precision issues between numpy and native which we should mask.

bpkroth: I guess the main behavior I still somewhat prefer is that the count increases not when I call suggest, but when I call register, especially since I may wish to schedule that config at some point in the future and don't want to see it suggested again in the meantime.
But again, that count tracking should be independent of the number of times we repeat them.

bpkroth: That said, I'm not certain I'm super strong on any of these approaches.  I think they all have pros and cons.

mlos_bench/mlos_bench/tests/launcher_run_test.py
motus: I am always in favor of multiple small PRs 😸

mlos_bench/mlos_bench/optimizers/base_optimizer.py
motus: > that particular scenario would only show up except for floating point precision issues between numpy and native which we should mask.

oh no, it will be everywhere - e.g. for `sched_migration_cost_ns` (an integer, by the way) - are the values of 500000 and 500100 different or the same? how about 500000 and 500001? etc., etc.

mlos_bench/mlos_bench/tests/launcher_run_test.py
bpkroth: As always, it depends.  I prefer them too, unless it's too hard to see what's happening or there's breakage in the middle.  Anyways, let's move it to the next one.  Thanks!

mlos_bench/mlos_bench/optimizers/base_optimizer.py
bpkroth: I would argue those quantization issues are application domain specfic, and probably quantization specification issues in that case and are in fact different configs from the optimizer perspective.

motus: exactly. For me, that's another argument for factoring out the tracking of suggested/pending configs etc. into a separate class and PR

mlos_bench/mlos_bench/tests/launcher_run_test.py
motus: ok, I've updated it to `--max_suggestions` - hope you're ok with this!

mlos_bench/mlos_bench/config/schemas/optimizers/optimizer-schema.json
```diff
@@ -22,8 +22,8 @@
                     "enum": ["min", "max"],
                     "example": "min"
                 },
-                "max_iterations": {
-                    "description": "The maximum number of additional (in the case of merging experiment data or resuming experiments) iterations to run when we launch the app.",
+                "max_suggestions": {
+                    "description": "The maximum number of additional (in the case of merging experiment data or resuming experiments) suggestions/trials to run when we launch the app.",
```
bpkroth, line 26:
```suggestion
                    "description": "The maximum number of additional (in the case of merging experiment data or resuming experiments) config suggestions to run when we launch the app.  Note: configs may be repeated in more than one trial.",
```

mlos_bench/README.md
```diff
@@ -195,10 +195,10 @@ Searching for an optimal set of tunable parameters is very similar to running a
 All we have to do is specifying the [`Optimizer`](./mlos_bench/optimizers/) in the top-level configuration, like in our [`azure-redis-opt.jsonc`](./mlos_bench/config/cli/azure-redis-opt.jsonc) example.
 
 ```sh
-mlos_bench --config "./mlos_bench/mlos_bench/config/cli/azure-redis-opt.jsonc" --globals "experiment_MyBenchmark.jsonc" --max_iterations 10
+mlos_bench --config "./mlos_bench/mlos_bench/config/cli/azure-redis-opt.jsonc" --globals "experiment_MyBenchmark.jsonc" --max_suggestions 10
```
bpkroth, line 198: Perhaps expand this and the text below to also illustrate use of `--config_trial_repeat_count`.

mlos_bench/mlos_bench/schedulers/base_scheduler.py
```diff
@@ -177,21 +178,24 @@ def load_config(self, config_id: int) -> TunableGroups:
             _LOG.debug("Config %d ::\n%s", config_id, json.dumps(tunable_values, indent=2))
         return tunables
 
-    def _get_optimizer_suggestions(self, last_trial_id: int = -1, is_warm_up: bool = False) -> int:
+    def _get_optimizer_suggestions(self) -> bool:
```
bpkroth, line 181:
```suggestion
    def _schedule_new_optimizer_suggestions(self) -> bool:
```

bpkroth (commented):
A few minor comments on documentation and naming I think we should address, but mostly looks good.

I'm going to update the title and description to match.

mlos_bench/mlos_bench/schedulers/base_scheduler.py
motus: fixed

mlos_bench/README.md
motus: done

mlos_bench/mlos_bench/config/schemas/optimizers/optimizer-schema.json
```diff
@@ -22,8 +22,8 @@
                     "enum": ["min", "max"],
                     "example": "min"
                 },
-                "max_iterations": {
-                    "description": "The maximum number of additional (in the case of merging experiment data or resuming experiments) iterations to run when we launch the app.",
+                "max_suggestions": {
+                    "description": "The maximum number of additional (in the case of merging experiment data or resuming experiments) config suggestions to run when we launch the app.  Note: configs may be repeated in more than one trial.",
```
bpkroth, line 26:
```suggestion
                    "description": "The maximum number of additional (in the case of merging experiment data or resuming experiments) config suggestions to run when we launch the app, or no limit if 0 is provided. Note: configs may be repeated in more than one trial.",
```

bpkroth (approved):
Minor doc tweak.

PR #716, opened 2024-03-16
UTC datetime handling improvements

### LINKED ISSUE #718: Get rid of deprecated functions like `datetime.utcnow()`
Also, `.utcfromtimestamp()`

make sure that we always use UTC for our timestamps (esp. check the storage w/different backends)

motus:
Addresses some deprecation warnings about use of `datetime.utcnow()`.

Additionally, sqlite does not store timezone info even if provided, so retrieving the data may be interpreted according to the host machine's *local* timezone, which may or may not be UTC.

To mitigate this, this change ensures that all timestamps are 
1. first converted to UTC before storing into the DB,
2. converted (or augmented with zoneinfo) to UTC on retrieval

Additionally, we expand the tests to check for this behavior, first with some additional conversion matrixes when telemetry or status data is received in implicit local vs. explicit timezone data as well as executions where the implicit local timezone has be overridden with a `TZ` environment variable, to simulate different default timezone hosts.

Closes #718


bpkroth:
I started hacking on this a little locally as well.  Will post a PR against yours in a bit.

mlos_bench/mlos_bench/environments/local/local_env.py
```diff
@@ -240,6 +240,7 @@ def _datetime_parser(datetime_col: pandas.Series) -> pandas.Series:
         ValueError
             On parse errors.
         """
+        # FIXME: We should probably not be assuming results arrived in UTC timezone.
         new_datetime_col = pandas.to_datetime(datetime_col, utc=True)
```
bpkroth, line 244: This is the main problem with the remaining two sets of test cases that are failing.

Essentially, this conversion happens incorrectly, so we need to handle it a little differently.

Here's an example:
```sh
TZ="America/Chicago" pytest -n0 mlos_bench/mlos_bench/tests/environments/local/local_env_telemetry_test.py
```

bpkroth: Will take a look at this more tomorrow.

bpkroth: NOTE: that form of testing doesn't currently happen in the pipeline - will have to fix that later too.

bpkroth: The root cause is also fixed now.

bpkroth (approved):
LGTM, but I wrote half of it :)
@motus, do you want to take another look?

mlos_bench/mlos_bench/tests/__init__.py
```diff
@@ -6,21 +6,36 @@
...
+ZONE_NAMES = [
+    # Explicit time zones.
+    "UTC",
+    "America/Chicago",
+    "America/Los_Angeles",
+    # Implicit local time zone.
+    None,
+]
+ZONE_INFO: List[Optional[tzinfo]] = [
+    None if zone_name is None else pytz.timezone(zone_name)
```
motus, line 34:
```suggestion
    nullable(pytz.timezone, zone_name)
```

mlos_bench/mlos_bench/environments/local/local_env.py
```diff
@@ -240,7 +240,14 @@ def _datetime_parser(datetime_col: pandas.Series) -> pandas.Series:
...
         """
-        new_datetime_col = pandas.to_datetime(datetime_col, utc=True)
+        new_datetime_col = pandas.to_datetime(datetime_col, utc=False)
+        # If timezone data is missing, assume the local timezone.
+        if new_datetime_col.dt.tz is None:
+            local_tzinfo = datetime.now().astimezone().tzinfo
+            new_datetime_col = new_datetime_col.dt.tz_localize(local_tzinfo)
+        assert new_datetime_col.dt.tz is not None
+        # And convert it to UTC.
+        new_datetime_col = new_datetime_col.dt.tz_convert('UTC')
```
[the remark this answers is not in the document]
bpkroth: Done.

Funny thing is when I first added that function in a prior PR I did have it in `mlos_bench.util` :)

Issue #715, opened 2024-03-15
Revamp number of iterations vs. number of configs tracking


bpkroth:
Discussed offline:

- previously (#642) we artificially increased the iteration count so that when we repeated trials it wouldn't lessen the number of configs we sampled
- now that we're separating the optimizer (suggest config) and scheduler (repeat of configs), the we can
  - [x] move the tracking of number of configs suggested to the `suggest()` routine in the Optimizer (#711), and 
  - [x] `number of trials executed` to the Scheduler
- that opens up the door for more flexible stopping conditions (#427)
- to do that, we should also cleanup our terminology and 
  - [x] rename `max_iterations` to `num_config_suggestions`, and 
  - [x] add a new `max_trials` argument (separate PR though I think)

_Originally posted by @bpkroth in https://github.com/microsoft/MLOS/pull/710#discussion_r1526797056_

motus:
I think we've fixed this one in #713 and #719

PR #719, opened 2024-03-19
Add `max_trials` scheduler config parameter

motus:
maybe merge after #713 (they might conflict)

Part of issue #715

### DIFF
--- a/mlos_bench/mlos_bench/config/schedulers/sync_scheduler.jsonc
+++ b/mlos_bench/mlos_bench/config/schedulers/sync_scheduler.jsonc
@@ -6,7 +6,7 @@
 
     "config": {
         "trial_config_repeat_count": 3,
-        "max_trials": 100,
+        "max_trials": -1,  // Limited only in hte Optimizer logic/config.
         "teardown": false
     }
 }
--- a/mlos_bench/mlos_bench/config/schedulers/sync_scheduler.jsonc
+++ b/mlos_bench/mlos_bench/config/schedulers/sync_scheduler.jsonc
@@ -6,6 +6,7 @@
 
     "config": {
         "trial_config_repeat_count": 3,
+        "max_trials": 100,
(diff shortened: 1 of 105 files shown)


mlos_bench/mlos_bench/config/schemas/schedulers/scheduler-schema.json
```diff
@@ -25,6 +25,12 @@
                     "description": "Whether to teardown the experiment after running it.",
                     "type": "boolean"
                 },
+                "max_trials": {
+                    "description": "Max. number of trials to run",
+                    "type": "integer",
+                    "minimum": 1,
```
bpkroth, line 31: Should allow this to support -1 as well and document what that means.

mlos_bench/mlos_bench/schedulers/base_scheduler.py
```diff
@@ -234,10 +236,20 @@ def _run_schedule(self, running: bool = False) -> None:
         for trial in self.experiment.pending_trials(datetime.utcnow(), running=running):
             self.run_trial(trial)
 
+    def not_converged(self) -> bool:
```
bpkroth, line 239: May want to consider renaming this to "not_done" or something like that.

Converged means something specific in the optimizer case, but doesn't quite have the same meaning here.

motus: fixed

mlos_bench/mlos_bench/config/schedulers/sync_scheduler.jsonc
```diff
@@ -6,6 +6,7 @@
 
     "config": {
         "trial_config_repeat_count": 3,
+        "max_trials": 100,
```
bpkroth, line 9: Set the default to -1?

bpkroth (approved):

Issue #692, opened 2024-02-28
Integrate multi objective optimization


bpkroth:
Both FLAML and SMAC support multi objective optimization now.

We should extend mlos_core and mlos_bench to handle this.

PR #723, opened 2024-03-22
Pass multiple metrics into the optimizer

motus:
We will need this for multi-objective optimization as well as for training the optimizers that can take muti-dimensional input

* [x] Make `Storage.Experiment.load()` return multiple scores
* [x] Fix unit tests to check for loading multi-dimensional scores from the DB
* [x] Make `Optimizer.register()` and `.bulk_register()` take multi-dimensional trial scores
* [x] Fix the Optimizer unit tests to check for registering multi-dimensional scores
* [x] Check the Scheduler and optimization loop unit tests to see if we need to adjust the types etc.

**NOTE:** In this PR, we _do not_ change mlos_core: we will still pass a single scalar into it and do not change the API on mlos_core side. We will change mlos_core in the subsequent PR to minimize the diff.

Part of #692

### DIFF
--- a/mlos_bench/mlos_bench/storage/sql/experiment.py
+++ b/mlos_bench/mlos_bench/storage/sql/experiment.py
@@ -13,7 +13,7 @@ from typing import Optional, Tuple, List, Dict, Iterator, Any
 
 from pytz import UTC
 
-from sqlalchemy import Engine, Connection, CursorResult, Table, column, func
+from sqlalchemy import Engine, Connection, CursorResult, Table, column, func, select
 
 from mlos_bench.environments.status import Status
 from mlos_bench.tunables.tunable_groups import TunableGroups
@@ -172,10 +172,10 @@ class Experiment(Storage.Experiment):
         (E.g., configurations, results, and telemetry).
         """
         cur_result: CursorResult[Tuple[str, Any]] = conn.execute(
-            table.select().with_only_columns(
+            select(
                 column(f"{field}_id"),
                 column(f"{field}_value"),
-            ).where(
+            ).select_from(table).where(
                 *[column(key) == val for (key, val) in kwargs.items()]
             )
         )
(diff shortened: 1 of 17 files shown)


mlos_bench/mlos_bench/storage/sql/experiment.py
```diff
@@ -127,50 +127,59 @@ def load_telemetry(self, trial_id: int) -> List[Tuple[datetime, str, Any]]:
...
+        """
+        cur_result: CursorResult[Tuple[str, Any]] = conn.execute(
+            select(
+                column(f"{field}_id"),
+                column(f"{field}_value"),
+            ).select_from(table).where(
+                *[column(key) == val for (key, val) in kwargs.items()]
+            )
+        )
+        return dict(row._tuple() for row in cur_result.fetchall())  # pylint: disable=protected-access
```
motus, line 182: `_tuple()` is not a protected member; named tuple adds `_` to their methods to avoid name conflicts

mlos_bench/mlos_bench/tests/optimizers/grid_search_optimizer_test.py
```diff
@@ -239,7 +239,7 @@ def test_grid_search_async_order(grid_search_opt: GridSearchOptimizer) -> None:
     assert all(suggestion.get_param_values() not in suggested_shuffled for suggestion in suggested)
 
     grid_search_opt.bulk_register([suggestion.get_param_values() for suggestion in suggested],
-                                  [score] * len(suggested),
+                                  [{"score": score}] * len(suggested),
```
bpkroth, line 242: Add a test with multiple `{"metric_1": score_1, "metric_2": score_2}`?

motus: It requires a bit more plumbing, so I've implemented it in the next PR #726

PR #725, opened 2024-03-26
Allow multiple objectives in the optimizer config

motus:
* [x] Update JSON schema to use `"optimizer_target": {"score": "min"}` format
* [x] Update base Optimizer class to use the new config format (throw `NotImplementedError` if > 1 target)
* [x] Modify all unit tests to use the new format

Part of #692

Merge after #723

PR #726, opened 2024-03-29
Implement multi-objective optimization in mlos_bench base classes

motus:
* [x] Store multiple optimization targets and directions in base `Optimizer` class
* [x] Support multiple optimization targets in mock and grid search optimizers
* [x] Check for single objective in `MlosCoreOptimizer` class (will add support for multiple objectives after implementing that feature in `mlos_core` in subsequent PRs)
* [x] Update unit tests in mlos_bench

Merge after ~#723~ and ~#725~

### DIFF
--- a/mlos_bench/mlos_bench/optimizers/base_optimizer.py
+++ b/mlos_bench/mlos_bench/optimizers/base_optimizer.py
@@ -209,7 +209,7 @@ class Optimizer(metaclass=ABCMeta):     # pylint: disable=too-many-instance-attr
         return self._opt_target
 
     @property
-    def direction(self) -> str:
+    def direction(self) -> Literal['min', 'max']:
         """
         The direction to optimize the target metric (e.g., min or max).
         """
--- a/mlos_bench/mlos_bench/optimizers/base_optimizer.py
+++ b/mlos_bench/mlos_bench/optimizers/base_optimizer.py
@@ -20,6 +20,7 @@ from ConfigSpace import ConfigurationSpace
 from mlos_bench.config.schemas import ConfigSchema
 from mlos_bench.services.base_service import Service
 from mlos_bench.environments.status import Status
+from mlos_bench.tunables.tunable import TunableValue
 from mlos_bench.tunables.tunable_groups import TunableGroups
 from mlos_bench.optimizers.convert_configspace import tunable_groups_to_configspace
 
@@ -223,7 +224,9 @@ class Optimizer(metaclass=ABCMeta):     # pylint: disable=too-many-instance-attr
         return True
 
     @abstractmethod
-    def bulk_register(self, configs: Sequence[dict], scores: Sequence[Optional[float]],
+    def bulk_register(self,
+                      configs: Sequence[dict],
+                      scores: Sequence[Optional[Dict[str, TunableValue]]],
                       status: Optional[Sequence[Status]] = None) -> bool:
         """
         Pre-load the optimizer with the bulk data from previous experiments.
@@ -232,9 +235,9 @@ class Optimizer(metaclass=ABCMeta):     # pylint: disable=too-many-instance-attr
         ----------
         configs : Sequence[dict]
             Records of tunable values from other experiments.
-        scores : Sequence[float]
+        scores : Sequence[Optional[Dict[str, TunableValue]]]
             Benchmark results from experiments that correspond to `configs`.
-        status : Optional[Sequence[float]]
+        status : Optional[Sequence[Status]]
             Status of the experiments that correspond to `configs`.
 
         Returns
--- a/mlos_bench/mlos_bench/optimizers/base_optimizer.py
+++ b/mlos_bench/mlos_bench/optimizers/base_optimizer.py
@@ -79,7 +79,7 @@ class Optimizer(metaclass=ABCMeta):     # pylint: disable=too-many-instance-attr
         self._max_iter = int(self._config.pop('max_suggestions', 100))
 
         opt_targets: Dict[str, str] = self._config.pop('optimization_targets', {'score': 'min'})
-        self._opt_targets: Dict[str, Literal[1, -1]]
+        self._opt_targets: Dict[str, Literal[1, -1]] = {}
         for (opt_target, opt_dir) in opt_targets.items():
             if opt_dir == "min":
                 self._opt_targets[opt_target] = 1
--- a/mlos_bench/mlos_bench/optimizers/base_optimizer.py
+++ b/mlos_bench/mlos_bench/optimizers/base_optimizer.py
@@ -35,7 +35,7 @@ class Optimizer(metaclass=ABCMeta):     # pylint: disable=too-many-instance-attr
     BASE_SUPPORTED_CONFIG_PROPS = {
         "optimization_target",
         "optimization_direction",
-        "max_iterations",
+        "max_suggestions",
         "seed",
         "start_with_defaults",
     }
@@ -71,12 +71,12 @@ class Optimizer(metaclass=ABCMeta):     # pylint: disable=too-many-instance-attr
         experiment_id = self._global_config.get('experiment_id')
         self.experiment_id = str(experiment_id).strip() if experiment_id else None
 
-        self._iter = 1
+        self._iter = 0
         # If False, use the optimizer to suggest the initial configuration;
         # if True (default), use the already initialized values for the first iteration.
         self._start_with_defaults: bool = bool(
             strtobool(str(self._config.pop('start_with_defaults', True))))
-        self._max_iter = int(self._config.pop('max_iterations', 100))
+        self._max_iter = int(self._config.pop('max_suggestions', 100))
         self._opt_target = str(self._config.pop('optimization_target', 'score'))
         self._opt_sign = {"min": 1, "max": -1}[self._config.pop('optimization_direction', 'min')]
 
@@ -224,7 +224,7 @@ class Optimizer(metaclass=ABCMeta):     # pylint: disable=too-many-instance-attr
 
     @abstractmethod
     def bulk_register(self, configs: Sequence[dict], scores: Sequence[Optional[float]],
-                      status: Optional[Sequence[Status]] = None, is_warm_up: bool = False) -> bool:
+                      status: Optional[Sequence[Status]] = None) -> bool:
         """
         Pre-load the optimizer with the bulk data from previous experiments.
 
@@ -236,16 +236,13 @@ class Optimizer(metaclass=ABCMeta):     # pylint: disable=too-many-instance-attr
             Benchmark results from experiments that correspond to `configs`.
         status : Optional[Sequence[float]]
             Status of the experiments that correspond to `configs`.
-        is_warm_up : bool
-            True for the initial load, False for subsequent calls.
 
         Returns
         -------
         is_not_empty : bool
             True if there is data to register, false otherwise.
         """
-        _LOG.info("%s the optimizer with: %d configs, %d scores, %d status values",
-                  "Warm-up" if is_warm_up else "Load",
+        _LOG.info("Update the optimizer with: %d configs, %d scores, %d status values",
                   len(configs or []), len(scores or []), len(status or []))
         if len(configs or []) != len(scores or []):
             raise ValueError("Numbers of configs and scores do not match.")
@@ -257,10 +254,11 @@ class Optimizer(metaclass=ABCMeta):     # pylint: disable=too-many-instance-attr
             self._start_with_defaults = False
         return has_data
 
-    @abstractmethod
     def suggest(self) -> TunableGroups:
         """
         Generate the next suggestion.
+        Base class' implementation increments the iteration count
+        and returns the current values of the tunables.
 
         Returns
         -------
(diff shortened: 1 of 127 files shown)


mlos_bench/mlos_bench/optimizers/base_optimizer.py
```diff
@@ -199,28 +199,15 @@ def name(self) -> str:
...
-    @property
-    def target(self) -> str:
-        """
-        The name of the target metric to optimize.
-        """
-        return self._opt_target
-
     @property
-    def direction(self) -> Literal['min', 'max']:
+    def targets(self) -> Dict[str, str]:
```
bpkroth, line 203:
```suggestion
    def targets(self) -> Dict[str, Literal['min', 'max']]:
```

mlos_bench/mlos_bench/optimizers/base_optimizer.py
```diff
@@ -79,13 +79,10 @@ def __init__(self,
...
 
         opt_targets: Dict[str, str] = self._config.pop('optimization_targets', {'score': 'min'})
-        if not isinstance(opt_targets, dict):
-            raise ValueError(f"optimization_targets should be a dict: {opt_targets}")
-        # TODO: Implement multi-target optimization.
-        if len(opt_targets) != 1:
-            raise NotImplementedError("Multi-target optimization is not implemented.")
-        (self._opt_target, opt_dir) = list(opt_targets.items())[0]
-        self._opt_sign = {"min": 1, "max": -1}[opt_dir]
+        self._opt_targets: Dict[str, int] = {
```
bpkroth, line 82:
```suggestion
        self._opt_targets: Dict[str, Literal[1, -1]] = {
```

bpkroth: not sure about this one

mlos_bench/mlos_bench/optimizers/base_optimizer.py
bpkroth: Nice for autocompletes and static checks

mlos_bench/mlos_bench/optimizers/base_optimizer.py
```diff
@@ -317,27 +307,33 @@ def _get_score(self, status: Status, score: Optional[Union[float, Dict[str, floa
...
-                                     + f"in results for iteration {self._iter}: {score}")
-                score = score[self._opt_target]
-            return float(score) * self._opt_sign
-        assert score is None
-        return float("inf")
+
+        if not status.is_succeeded():
+            assert score is None
+            return {
+                opt_target: float("inf")
```
bpkroth, line 325: does this need to be adjusted based on target direction?

mlos_bench/mlos_bench/optimizers/base_optimizer.py
```diff
@@ -317,27 +307,33 @@ def _get_score(self, status: Status, score: Optional[Union[float, Dict[str, floa
         ----------
         status : Status
             Final status of the experiment (e.g., SUCCEEDED or FAILED).
-        score : Union[float, Dict[str, float]]
-            A scalar or a dict with the final benchmark results.
+        score : Optional[Dict[str, TunableValue]]
```
bpkroth, line 310:
```suggestion
        scores : Optional[Dict[str, TunableValue]]
```

mlos_bench/mlos_bench/optimizers/base_optimizer.py
```diff
@@ -293,22 +280,25 @@ def register(self, tunables: TunableGroups, status: Status,
...
+            from the dataframe that's being MINIMIZED.
         """
         _LOG.info("Iteration %d :: Register: %s = %s score: %s",
                   self._iter, tunables, status, score)
         if status.is_succeeded() == (score is None):  # XOR
             raise ValueError("Status and score must be consistent.")
         return self._get_score(status, score)
 
-    def _get_score(self, status: Status, score: Optional[Union[float, Dict[str, float]]]) -> Optional[float]:
+    def _get_score(self, status: Status,
```
bpkroth, line 299:
```suggestion
    def _get_scores(self, status: Status,
```

mlos_bench/mlos_bench/optimizers/base_optimizer.py
```diff
@@ -293,22 +280,25 @@ def register(self, tunables: TunableGroups, status: Status,
...
+                   score: Optional[Union[Dict[str, TunableValue], Dict[str, float]]]
```
bpkroth, line 300:
```suggestion
                   scores: Optional[Union[Dict[str, TunableValue], Dict[str, float]]]
```

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -99,10 +98,8 @@ def experiment(self, *,
         description : str
             Human-readable description of the experiment.
         tunables : TunableGroups
-        opt_target : str
-            Name of metric we're optimizing for.
-        opt_direction: Optional[str]
-            Direction to optimize the metric (e.g., min or max)
+        opt_targets : Dict[str, str]
```
bpkroth, line 101:
```suggestion
        opt_targets : Dict[str, Literal["min", "max"]]
```

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -212,14 +206,11 @@ def tunables(self) -> TunableGroups:
...
         @property
-        def opt_target(self) -> str:
-            """Get the Experiment's optimization target"""
-            return self._opt_target
-
-        @property
-        def opt_direction(self) -> Optional[str]:
-            """Get the Experiment's optimization target"""
-            return self._opt_direction
+        def opt_targets(self) -> Dict[str, str]:
```
bpkroth, line 209:
```suggestion
        def opt_targets(self) -> Dict[str, Literal["min", "max"]]:
```

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -356,18 +345,11 @@ def tunable_config_id(self) -> int:
             return self._tunable_config_id
 
         @property
-        def opt_target(self) -> str:
+        def opt_targets(self) -> Dict[str, str]:
```
bpkroth, line 348:
```suggestion
        def opt_targets(self) -> Dict[str, Literal["min", "max"]]:
```

mlos_bench/mlos_bench/storage/sql/experiment_data.py
```diff
@@ -52,56 +52,19 @@ def __init__(self, *,
...
-                _LOG.warning("Experiment %s has multiple trial optimization directions for optimization_target %s=%s",
-                             self, opt_target, objectives[opt_target])
-        for opt_tgt, opt_dir in objectives.items():
-            assert opt_dir in {None, "min", "max"}, f"Unexpected opt_dir {opt_dir} for opt_tgt {opt_tgt}."
-        return objectives
+            )
+            return {
+                objective.optimization_target: objective.optimization_direction
+                for objective in objectives_db_data.fetchall()
+            }
```
bpkroth, line 67: This was tricky to write when first implemented.  Should make sure we have some good test coverage for it (may already be there, but worth double checking).

mlos_bench/mlos_bench/optimizers/base_optimizer.py
motus: it does not - that's an internal function and it assumes that `self._opt_targets` has already been adjusted for target direction.

mlos_bench/mlos_bench/schedulers/base_scheduler.py
```diff
@@ -214,13 +213,14 @@ def schedule_trial(self, tunables: TunableGroups) -> None:
...
                 "optimizer": self.optimizer.name,
-                "opt_target": self.optimizer.target,
-                "opt_direction": self.optimizer.direction,
                 "repeat_i": repeat_i,
                 "is_defaults": tunables.is_defaults,
+                **{
+                    f"opt_{key}_{i}": val
+                    for (i, opt_target) in enumerate(self.optimizer.targets.items())
+                    for (key, val) in zip(["target", "direction"], opt_target)
+                }
```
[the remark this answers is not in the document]
bpkroth: ExperimentData, TrialData, etc.?

mlos_bench/mlos_bench/optimizers/base_optimizer.py
bpkroth: But, down below we do the multiplication by the opt_direction and here we don't.

Doesn't that mean that for unsuccessful trials, we always report `float("inf")` even if the `opt_dir` is `"max"` for that `opt_target`?

That seems like a bug to me.  Do we have a test case checking for this?

mlos_bench/mlos_bench/optimizers/base_optimizer.py
```diff
@@ -317,27 +311,33 @@ def _get_score(self, status: Status, score: Optional[Union[float, Dict[str, floa
...
+            return {
+                opt_target: float("inf")
+                for opt_target in self._opt_targets.keys()
+            }
+
+        assert scores is not None
+        target_metrics: Dict[str, float] = {}
+        for (opt_target, opt_dir) in self._opt_targets.items():
+            val = scores[opt_target]
+            assert val is not None
```
bpkroth, line 337:
```suggestion
```

Maybe unnecessary based on latest type changes?

mlos_bench/mlos_bench/optimizers/base_optimizer.py
```diff
@@ -347,13 +347,13 @@ def not_converged(self) -> bool:
         return self._iter < self._max_iter
 
     @abstractmethod
-    def get_best_observation(self) -> Union[Tuple[float, TunableGroups], Tuple[None, None]]:
+    def get_best_observation(self) -> Union[Tuple[Dict[str, float], TunableGroups], Tuple[None, None]]:
```
bpkroth, line 350:
```suggestion
    def get_best_observations(self) -> Union[Tuple[Dict[str, float], TunableGroups], Tuple[None, None]]:
```

Don't we need to return the full pareto frontier now?

bpkroth: Another option is we extend it to `get_best_observation(weights={
  "opt_target_1": 1,
  "opt_target_2": 0,
  "opt_target_3": 1,
})

As a means to flatten the parent frontier into a single point by means of multi-dimensional score weighting.

mlos_bench/mlos_bench/schedulers/base_scheduler.py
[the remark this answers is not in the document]
motus: I don't expect metadata to be the primary source of information about the optimization objectives, but yes, we have unit tests to verify that data in `ExperimentData` and `TrialData`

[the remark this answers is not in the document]
motus: on a second thought, we can add `TrialData.objectives` for convenience. I've opened an issue #729 - will send a separate PR with proper unit tests and stuff later

mlos_bench/mlos_bench/storage/sql/experiment_data.py
motus: I would not rely on trials' metadata to determine the experiment's objective. To me, the Experiment's objectives are defined when we create the experiment; if later we merge in some other trials into it, the experiment's objectives stay the same, even if some trials had been run with different objectives or no objectives at all. The info on individual trial's objectives we can get from the metadata, and I will soon implement an accessor `TrialData.objectives` to facilitate that (issue #729)

mlos_bench/mlos_bench/optimizers/base_optimizer.py
motus: Nah, the linter complains - `val` is of type `TunableValue` that theoretically can be `None`

mlos_bench/mlos_bench/optimizers/base_optimizer.py
motus: I would do it in the next PR after we add multi-target optimization to `mlos_core` - rn I am not even sure that any of our optimizers can return all points of the Pareto frontier. (also, it will require a lot of changes to our unit tests)

mlos_bench/mlos_bench/optimizers/base_optimizer.py
motus: `Optimizer._get_scores()` _always_ returns scores adjusted for **minimization** (I mention that in the docstring). That's why we do the multiplication here. If the trial is not successful, for minimization we must always return _positive_ infinity (or later maybe some other large _positive_ value).

I've added a unit test to verify that behavior

mlos_bench/mlos_bench/schedulers/base_scheduler.py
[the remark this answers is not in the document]
bpkroth: I'm not sure `TrialData` is the right place for that.  `ExperimentData.objectives` already exists and makes more sense to me when you consider that the data from individual Trials could be reused across Experiments.

mlos_bench/mlos_bench/storage/sql/experiment_data.py
bpkroth: I agree on the first part.  Using metadata was a means to backport functionality since early on we didn't have a separate `objectives` table that could be used to pull the data from the `Experiment` directly.

But I don't think we should expose the objectives for TrialData itself - its less relevant there versus `Experiment` for exactly the merging operation you point out.

bpkroth (approved):

Issue #523, opened 2023-10-03
mlos_bench: error handling improvements


bpkroth:
Sometimes user scripts don't return a `score` value, even though they exit 0 (indicating SUCCESS).

In that case we can do a couple of things:

- abort immediately in order to notify the experimenter and let them figure out what to do
- assume it's a bad config and that's why the benchmark aborted early
  - in which case we should fabricate a "fake" score that looks "bad" (i.e., much worse than any we've actually recorded with a good config) so that the optimizer learns that this is an infeasible region (there are already TODO markers in the code to implement this)
- some cominbation of the two
  for instance, tolerate no more than N "bad" configs in a row before we assume its a script error and abort entirely to notify the user that they should manually inspect and deal with things

bpkroth:
@eujing

bpkroth:
One thing that might make this easier to implement is if we clearly separated the phases of "setup" (e.g., basic system preparation) vs. "configure" (e.g., configure the target system with the tunables).

That way, if "setup" failed, we could alert that the script was the problem, whereas if "configure" failed, we could inform the optimizer that it was a bad region.

PR #730, opened 2024-05-01
Update mlos_core API to support with multi-factor optimization and contexts

motus:
* [x] Pass multi-column DataFrame instead of Sequence to `BaseOptimizer.register()` and other methods that deal with scores
* [x] Update mlos_bench `MlosCoreOptimizer` to support the new mlos_core API
* [ ] Update unit tests to work with the new API
* [ ] Add unit tests for end-to-end multi-target optimization

### DIFF
--- a/mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
+++ b/mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
@@ -114,7 +114,7 @@ class MlosCoreOptimizer(Optimizer):
         self._opt.register(df_configs, df_scores)
         if _LOG.isEnabledFor(logging.DEBUG):
             (score, _) = self.get_best_observation()
-            _LOG.debug("Warm-up END: %s = %s", self, score)
+            _LOG.debug("Warm-up END: %s :: %s", self, score)
         return True
 
     def _extract_target(self, scores: Optional[Dict[str, TunableValue]]) -> Optional[TunableValue]:
--- a/mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
+++ b/mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
@@ -46,11 +46,6 @@ class MlosCoreOptimizer(Optimizer):
                  service: Optional[Service] = None):
         super().__init__(tunables, config, global_config, service)
 
-        # TODO: Remove after implementing multi-target optimization in mlos_core
-        if len(self._opt_targets) != 1:
-            raise NotImplementedError(f"Multi-target optimization is not supported: {self}")
-        (self._opt_target, self._opt_sign) = list(self._opt_targets.items())[0]
-
         opt_type = getattr(OptimizerType, self._config.pop(
             'optimizer_type', DEFAULT_OPTIMIZER_TYPE.name))
 
@@ -99,26 +94,42 @@ class MlosCoreOptimizer(Optimizer):
                       configs: Sequence[dict],
                       scores: Sequence[Optional[Dict[str, TunableValue]]],
                       status: Optional[Sequence[Status]] = None) -> bool:
+
         if not super().bulk_register(configs, scores, status):
             return False
+
         df_configs = self._to_df(configs)  # Impute missing values, if necessary
-        df_scores = pd.Series(
-            [self._extract_target(score) for score in scores],
-            dtype=float) * self._opt_sign
+
+        df_scores = self._adjust_signs_df(
+            pd.DataFrame([{} if score is None else score for score in scores]))
+
+        opt_targets = list(self._opt_targets)
         if status is not None:
+            # Select only the completed trials, set scores for failed trials to +inf.
             df_status = pd.Series(status)
-            df_scores[df_status != Status.SUCCEEDED] = float("inf")
+            df_scores.loc[df_status != Status.SUCCEEDED, opt_targets] = float("inf")
             df_status_completed = df_status.apply(Status.is_completed)
             df_configs = df_configs[df_status_completed]
             df_scores = df_scores[df_status_completed]
-        self._opt.register(df_configs, df_scores)
+
+        # Pass opt_targets columns as scores, and the rest as context.
+        opt_context = set(df_scores.columns).difference(opt_targets)
+        self._opt.register(df_configs, df_scores[opt_targets],
+                           df_scores[list(opt_context)] if opt_context else None)
+
         if _LOG.isEnabledFor(logging.DEBUG):
             (score, _) = self.get_best_observation()
             _LOG.debug("Warm-up END: %s :: %s", self, score)
+
         return True
 
-    def _extract_target(self, scores: Optional[Dict[str, TunableValue]]) -> Optional[TunableValue]:
-        return None if scores is None else scores[self._opt_target]
+    def _adjust_signs_df(self, df_scores: pd.DataFrame) -> pd.DataFrame:
+        """
+        In-place adjust the signs of the scores for MINIMIZATION problem.
+        """
+        for (opt_target, opt_dir) in self._opt_targets.items():
+            df_scores[opt_target] *= opt_dir
+        return df_scores
 
     def _to_df(self, configs: Sequence[Dict[str, TunableValue]]) -> pd.DataFrame:
         """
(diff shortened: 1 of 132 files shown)


mlos_core/mlos_core/optimizers/__init__.py
```diff
@@ -62,6 +62,7 @@ class OptimizerFactory:
     @staticmethod
     def create(*,
                parameter_space: ConfigSpace.ConfigurationSpace,
+               optimization_targets: List[str],
```
bpkroth, line 65: Shouldn't this be:
```suggestion
               optimization_targets: List[Tuple[str, Literal[min, max]]],
```

bpkroth: Ah, nm, this was for the mlos_core optimizer, which we always minimize?

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -241,13 +257,14 @@ def _register(self, configurations: pd.DataFrame, scores: pd.Series, context: Op
         from smac.runhistory import StatusType, TrialInfo, TrialValue  # pylint: disable=import-outside-toplevel
 
         if context is not None:
-            raise NotImplementedError()
+            warn(f"Not Implemented: Ignoring context {list(context.columns)}", UserWarning)
```
bpkroth, line 260: Maybe leave that change for a separate PR?

mlos_core/mlos_core/optimizers/flaml_optimizer.py
```diff
@@ -137,7 +150,7 @@ def _target_function(self, config: dict) -> Union[dict, None]:
         """
         cs_config = normalize_config(self.optimizer_parameter_space, config)
         if cs_config in self.evaluated_samples:
-            return {'score': self.evaluated_samples[cs_config].score}
+            return {self._optimization_targets[0]: self.evaluated_samples[cs_config].score}
```
bpkroth, line 153: Nit: as a maintenance thing, for FLAML, I'd suggest we introduce a `self._optimization_target` property that returns `self._optimization_targets[0]` so that if they ever change it in the future it'll be a little easier to find the places to adjust.

mlos_core/mlos_core/optimizers/flaml_optimizer.py
```diff
@@ -26,35 +26,47 @@ class EvaluatedSample(NamedTuple):
...
+            If provided, calls np.random.seed() with the provided value to set the seed globally at init.
+        """
         super().__init__(
             parameter_space=parameter_space,
+            optimization_targets=optimization_targets,
             space_adapter=space_adapter,
         )
 
+        if len(self._optimization_targets) != 1:
+            raise ValueError("FLAML does not support multi-target optimization")
```
bpkroth, line 68: Hmm, this seems to indicate that multi-objective is supported?  Maybe we could reach out to Chi and discuss further.  Could also adjust the message for now and extend support for FLAML in a future PR.
https://microsoft.github.io/FLAML/docs/reference/tune/searcher/suggestion#optunasearch-objects

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -172,41 +178,43 @@ def register_pending(self, configurations: pd.DataFrame,
...
-        """Returns the best observation so far as a dataframe.
+        configs = pd.concat([config for config, _, _ in self._observations]).reset_index(drop=True)
+        scores = pd.concat([score for _, score, _ in self._observations]).reset_index(drop=True)
+        contexts = pd.concat([pd.DataFrame() if context is None else context
+                              for _, _, context in self._observations]).reset_index(drop=True)
+        return (configs, scores, contexts if len(contexts.columns) > 0 else None)
+
+    def get_best_observations(self, n_max: int = 1) -> Tuple[pd.DataFrame, pd.DataFrame, Optional[pd.DataFrame]]:
+        """
+        Get the N best observations so far as a triplet of DataFrames (config, score, context).
```
bpkroth, line 200: Add a note on how these are ordered?

For instance, this may produce the pareto frontier, or more than that if N is large, correct?

mlos_core/mlos_core/optimizers/flaml_optimizer.py
bpkroth: Actually, maybe it's just optuna that does that.  Not sure if we need a wrapper for a wrapper.  Maybe we could "just" incorporate optuna directly?

https://optuna.readthedocs.io/en/stable/index.html
https://optuna.readthedocs.io/en/stable/tutorial/10_key_features/002_configurations.html

mlos_core/mlos_core/optimizers/__init__.py
motus: exactly. in mlos_core we always minimize

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
motus: makes sense! smaller PRs are always good. I will do it now

mlos_core/mlos_core/optimizers/flaml_optimizer.py
motus: agree

mlos_core/mlos_core/optimizers/optimizer.py
motus: good point!

mlos_core/mlos_core/optimizers/flaml_optimizer.py
bpkroth: Oh, I thought SMAC did something a little different actually, but it's been a while since I looked.
I know that when Johannes chatted with them at NeurIPS they said the initial implementation was not the best and they were working on improving it.

motus: Just checked the SMAC3 code again. By default, they do the weighted average of the targets: [aggregation_strategy.py:44](https://github.com/automl/SMAC3/blob/main/smac/multi_objective/aggregation_strategy.py#L44)
There's also an implementation of [ParEGO](https://www.cs.bham.ac.uk/~jdk/UKCI-2015.pdf), but it is not used by default.

It would be pretty easy to implement weighted average aggregation strategy on top of FLAML; also, maybe we should pass the target weights as an optional parameter to all our optimizers?

motus:
Multi-target optimization in FLAML and optional weights for optimization targets implemented in the next PR #738

mlos_core/mlos_core/optimizers/flaml_optimizer.py
bpkroth: Ah, ParEGO was the one I was thinking of.  Johannes had done some experiments with that.

mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
```diff
@@ -99,26 +95,42 @@ def bulk_register(self,
...
             df_status = pd.Series(status)
-            df_scores[df_status != Status.SUCCEEDED] = float("inf")
+            df_scores.loc[df_status != Status.SUCCEEDED, opt_targets] = float("inf")
             df_status_completed = df_status.apply(Status.is_completed)
             df_configs = df_configs[df_status_completed]
             df_scores = df_scores[df_status_completed]
-        self._opt.register(df_configs, df_scores)
+
+        # Pass opt_targets columns as scores, and the rest as context.
+        opt_context = set(df_scores.columns).difference(opt_targets)
```
bpkroth, line 117: Disagree with this one for now.

There's so many things we might collect as metrics in the scores that aren't context that don't think we should default to that in this way.

For instance, imagine I collect throughput, latency, cost, elapsed time, and system metrics (CPU, disk, etc.).
In one run, I may want to optimize for increased throughput and lower elapsed time, where as in another run I may want to reduce CPU metrics and cost.
That data could be reused between experiments to prewarm an optimizer, but the omitted columns from one side vs. another don't tell me anything about the "context" of the test, only its results.

In my mind context are things about the execution environment (e.g., VM core count, VM RAM size, etc.) and the workload identification (WIP), but it wouldn't include execution metrics.

Let's leave this part out and make context handling something explicit in the future.

bpkroth (changes requested):
Let's back out the change to submitting all non-objective scores as context for now.  Thanks!

mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
```diff
@@ -99,26 +95,41 @@ def bulk_register(self,
...
+
+        df_scores = self._adjust_signs_df(
+            pd.DataFrame([{} if score is None else score for score in scores]))
+
+        opt_targets = list(self._opt_targets)
         if status is not None:
+            # Select only the completed trials, set scores for failed trials to +inf.
             df_status = pd.Series(status)
-            df_scores[df_status != Status.SUCCEEDED] = float("inf")
+            df_scores.loc[df_status != Status.SUCCEEDED, opt_targets] = float("inf")
```
bpkroth, line 111: We may still want to change this to something less awful, per #523.

For instance, 3-10x the worst *actually evaluated* sample recorded so far was shown in at least one paper I remember as performing better for models to learn on.

bpkroth: Maybe just add a comment to that effect for now.

bpkroth: Or could get handled in the base class somehow.
In any case, more a note for a future PR than anything else.

bpkroth (approved):
Minor comment request and I think so test failures, but LGTM

mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
motus: agree. I've added comments with the link to #523 in the places where we replace values with +inf

PR #735, opened 2024-05-13
Issue a warning on non-null optimization context

motus:


bpkroth (approved):
Discussed offline.  Feels a little bit like a no-op atm, but doesn't hurt anything, so its fine.

PR #741, opened 2024-05-15
Enable context for SMAC Optimizer

jsfreischuetz:

PR #751, opened 2024-05-22
Enable Optimizer state and Multi-Fidelity passthrough in SMAC

jsfreischuetz:
Enables changing the defaults in the SMAC optimizer to enable state passing (what SMAC internally calls context, but that doesn't match MLOS's notion of context) and multi-fidelity tuning.
This requires changes across the project to support the concept of state, as well as updating a few test cases.

This PR replaces the previous one (#741)


.vscode/settings.json
```diff
@@ -139,5 +138,8 @@
         "--log-level=DEBUG",
         "."
     ],
-    "python.testing.unittestEnabled": false
+    "python.testing.unittestEnabled": false,
+    "cSpell.words": [
+        "SOBOL"
+    ]
```
bpkroth, line 144: These go in different .cspell file, please move them there and revert the rest of the changes in this file.

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
```diff
@@ -388,14 +384,12 @@ def objective(point: pd.DataFrame) -> pd.DataFrame:
...
 
     (best_config, best_score, best_context) = optimizer.get_best_observations()
     assert isinstance(best_config, pd.DataFrame)
     assert isinstance(best_score, pd.DataFrame)
-    assert best_context is None
 
     (all_configs, all_scores, all_contexts) = optimizer.get_observations()
     assert isinstance(all_configs, pd.DataFrame)
     assert isinstance(all_scores, pd.DataFrame)
-    assert all_contexts is None
```
bpkroth, line 401: Can we assert that it *is* something expected now instead of just removing the check?

mlos_core/mlos_core/tests/optimizers/optimizer_multiobj_test.py
```diff
@@ -90,7 +89,6 @@ def objective(point: pd.DataFrame) -> pd.DataFrame:
     (all_configs, all_scores, all_contexts) = optimizer.get_observations()
     assert isinstance(all_configs, pd.DataFrame)
     assert isinstance(all_scores, pd.DataFrame)
-    assert all_contexts is None
```
bpkroth, line 93: Same here.  Can we assert that the contexts are something useful?

mlos_core/mlos_core/tests/optimizers/bayesian_optimizers_test.py
```diff
@@ -34,13 +34,10 @@ def test_context_not_implemented_warning(configuration_space: CS.ConfigurationSp
...
-    suggestion = optimizer.suggest()
+    suggestion, _ = optimizer.suggest()
     scores = pd.DataFrame({'score': [1]})
     context = pd.DataFrame([["something"]])
 
-    with pytest.raises(UserWarning):
-        optimizer.register(suggestion, scores, context=context)
-
     with pytest.raises(UserWarning):
         optimizer.suggest(context=context)
```
bpkroth, line 42: Hmm, so we can register context, but not suggest with it?

mlos_core/mlos_core/optimizers/random_optimizer.py
```diff
@@ -59,11 +61,15 @@ def _suggest(self, context: Optional[pd.DataFrame] = None) -> pd.DataFrame:
...
+
+        context : pd.DataFrame
+            Pandas dataframe with a single row containing the context.
+            Column names are the budget, seed, and instance of the evaluation, if valid.
         """
         if context is not None:
             # not sure how that works here?
             warn(f"Not Implemented: Ignoring context {list(context.columns)}", UserWarning)
-        return pd.DataFrame(dict(self.optimizer_parameter_space.sample_configuration()), index=[0])
+        return pd.DataFrame(dict(self.optimizer_parameter_space.sample_configuration()), index=[0]), None
```
bpkroth, line 72: Why should this return `None` for context?

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -26,7 +26,7 @@ class BaseOptimizer(metaclass=ABCMeta):
 
     def __init__(self, *,
                  parameter_space: ConfigSpace.ConfigurationSpace,
-                 optimization_targets: List[str],
+                 optimization_targets: str | List[str] | None = None,
```
bpkroth, line 29:
```suggestion
                 optimization_targets: Optional[Union[str, List[str]] = None,
```

We still need to support python 3.8

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
jsfreischuetz: No, because only SMAC uses context, where there other optimizers do not

mlos_core/mlos_core/tests/optimizers/optimizer_multiobj_test.py
jsfreischuetz: Same as above, unless we split out the SMAC tests

mlos_core/mlos_core/optimizers/random_optimizer.py
jsfreischuetz: I have interpreted the context that is passed in to the suggest to be different than the context that is returned. I am not exactly sure what the context in is supposed to mean

mlos_core/mlos_core/tests/optimizers/bayesian_optimizers_test.py
jsfreischuetz: It's more about not sending the warning not being being thrown anymore

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
bpkroth: Then check for that too.

mlos_core/mlos_core/tests/optimizers/optimizer_multiobj_test.py
bpkroth: Don't need to split them out, just add a branch in the test for them.
I want to make sure we're testing expected functionality in both directions.
It also makes it more obvious what needs to be done when we extend it elsewhere.

mlos_core/mlos_core/tests/optimizers/bayesian_optimizers_test.py
bpkroth: I understand what the code in this test is doing.  I'm asking why we aren't supporting suggestions with context yet when we're adding support for registering with context.  It makes it seem like a blackhole of information or only partial support somehow.

mlos_core/mlos_core/optimizers/random_optimizer.py
bpkroth: Here's what I expect:
- context is composed of some set of metrics describing the execution environment (e.g., vm ram, vm vcore count, workload descriptor, etc.)
- upon registering the optimizer is able to build a model for how different config+context pairs maps to different output metrics
- when requesting a new suggestion, possible for a different execution environment (e.g., larger vm size), described in the context, the optimizer should be able to use all of that information to provide a new config suggestion for that new context

Now, it's possible that the optimizer stinks at that part initially because it doesn't know how to compute suggestions for unseen contexts (future research work probably).
One option could be that we return a variable length (user specified) list of items, each stemming from different context of "known" items (ranked by higher confidence), instead of for a the new context and then the caller needs to sift through and decide what to do, or else we return a random config (explore) + the provided context and they could start trying them to help fill in that knowledge.
I think the exact specifics of that part is to be determined.

But high level, if someone asks for a suggestion for a given context, I think we should return something within context, not something for anywhere in the space.

Make sense?

.vscode/settings.json
bpkroth, line 1: revert again

mlos_core/mlos_core/tests/optimizers/bayesian_optimizers_test.py
jsfreischuetz: What does context for a suggestion mean?

mlos_core/mlos_core/optimizers/random_optimizer.py
jsfreischuetz: > context is composed of some set of metrics describing the execution environment (e.g., vm ram, vm vcore count, workload descriptor, etc.)

As far as I am aware this is not supported by any of the optimizers when asking for a suggestion. SMAC supports the idea of these values being defined on optimizer initialization, but which context to evaluate is determined by the optimizer, and there is still not support for requiring a specific context.

> But high level, if someone asks for a suggestion for a given context, I think we should return something within context, not something for anywhere in the space.

I think there is a problem with this. There are two ways to solve for this:
1. If we simply request configurations until we find a context that matches, we now have a long list of pending configurations that _should_ be evaluated according to the optimizer, but they are also stale
2. Alternatively, we can force the context to match, regardless of what the optimizer returns, but this is not actually sampling efficiently according to the optimizer.

I think realistically, we should remove the argument from suggest, but I tried to keep the changes as minimal as possible. Someone else before me also left a comment suggesting this.

mlos_core/mlos_core/optimizers/README.md
```diff
@@ -0,0 +1,18 @@
+This is a directory that contains wrappers for different optimizers to integrate into MLOS.
+This is implemented though child classes for the `BaseOptimizer` class defined in `optimizer.py`.
+
+The main goal of these optimizers is to take a suggest configurations based on prior samples to find an optimum based on some objective. This process is interacted with through and ask and tell interface.
```
bpkroth, line 4:
```suggestion
The main goal of these optimizers is to `suggest` configurations, possibly based on prior trial data to find an optimum based on some objective(s).
This process is interacted with through `register` and `suggest` interfaces.
```

Fixed some language.

Also, one sentence per line please.  Easier to review diffs later.

mlos_core/mlos_core/optimizers/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+The following defintions are useful for understanding the implementation
+- `configuration`: a vector representation of a configuration of a system to be evaluated.
+- `score`: the objective(s) associated with a configuration
+- `metadata`: additional information about the evaluation, such as the runtime budget used during evaluation.
+- `context`: additional information about the evaluation used to extend the internal model used for suggesting samples. This is not yet implemented.
```
bpkroth, line 10:
```suggestion
- `context`: additional (static) information about the evaluation used to extend the internal model used for suggesting samples.
   For instance, a descriptor of the VM size (vCore count and # of GB of RAM), and some descriptor of the workload.
   The intent being to allow either sharing or indexing of trial info between "similar" experiments in order to help make the optimization process more efficient for new scenarios.
   > Note: This is not yet implemented.
```

mlos_core/mlos_core/optimizers/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+The interface for these classes can be described as follows:
+
+- `register`: this is a function that takes a configuration, a score, and, optionally, metadata about the evaluation to update the model for future evaluations.
+- `suggest`: this function returns a new confiugration for evaluation. Some optimizers will return additional metadata for evaluation, that should be used durin the register phase. This function can also optionally take context (not yet implemented), and an argument to force the function to return the default configuration.
```
bpkroth, line 15:
```suggestion
- `suggest`: this function returns a new configuration for evaluation.

   Some optimizers will return additional metadata for evaluation, that should be used during the register phase.
   This function can also optionally take context (not yet implemented), and an argument to force the function to return the default configuration.
```

mlos_core/mlos_core/optimizers/README.md
```diff
@@ -0,0 +1,18 @@
...
+- `register_pending`: registers a configuration and metadata pair as pending to the optimizer.
+- `get_observations`: returns all observations reproted to the optimizer as a triplet of DataFrames (config, score, metadata).
+- `get_best_observations`: returns the best observation as A triplet of best (config, score, metadata) DataFrames.
```
bpkroth, line 18:
```suggestion
- `get_best_observations`: returns the best observation as a triplet of best (config, score, context, metadata) DataFrames.
```

mlos_core/mlos_core/optimizers/README.md
```diff
@@ -0,0 +1,18 @@
...
+- `score`: the objective(s) associated with a configuration
+- `metadata`: additional information about the evaluation, such as the runtime budget used during evaluation.
+- `context`: additional information about the evaluation used to extend the internal model used for suggesting samples. This is not yet implemented.
+
+The interface for these classes can be described as follows:
+
+- `register`: this is a function that takes a configuration, a score, and, optionally, metadata about the evaluation to update the model for future evaluations.
+- `suggest`: this function returns a new confiugration for evaluation. Some optimizers will return additional metadata for evaluation, that should be used durin the register phase. This function can also optionally take context (not yet implemented), and an argument to force the function to return the default configuration.
+- `register_pending`: registers a configuration and metadata pair as pending to the optimizer.
+- `get_observations`: returns all observations reproted to the optimizer as a triplet of DataFrames (config, score, metadata).
```
bpkroth, line 17:
```suggestion
- `get_observations`: returns all observations reproted to the optimizer as a triplet of DataFrames (config, score, context, metadata).
```

mlos_core/mlos_core/optimizers/flaml_optimizer.py
```diff
@@ -86,7 +86,7 @@ def __init__(self, *,   # pylint: disable=too-many-arguments
         self._suggested_config: Optional[dict]
 
     def _register(self, configurations: pd.DataFrame, scores: pd.DataFrame,
-                  context: Optional[pd.DataFrame] = None) -> None:
+                  metadata: Optional[pd.DataFrame] = None) -> None:
```
bpkroth, line 89: Please restore `context`.
The idea was to *add* `metadata` in addition to it, not replace it.

bpkroth: (requires associated changes throughout)

mlos_core/mlos_core/optimizers/random_optimizer.py
bpkroth: We discussed offline about the discrepency in terminology.

SMAC's notion of "context" is different from that in MLOS, which @jsfreischuetz rightly pointed out may or may not be supported directly yet.
SMAC's is much more about passing internal optimizer "state".

For now, we decided to

1. document the difference in the README.md above.
2. rename the SMAC required "state" to either `metadata` or `state`

mlos_core/mlos_core/tests/optimizers/bayesian_optimizers_test.py
bpkroth: We discussed offline and is documented in the suggestion in the README.md above now.

bpkroth (changes requested):
Please restore `context` and *add* `metadata` as new optional parameters.

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
jsfreischuetz: I made the associated changes

mlos_core/mlos_core/tests/optimizers/optimizer_multiobj_test.py
jsfreischuetz: I made the associated changes

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -26,7 +26,7 @@ class BaseOptimizer(metaclass=ABCMeta):
 
     def __init__(self, *,
                  parameter_space: ConfigSpace.ConfigurationSpace,
-                 optimization_targets: List[str],
+                 optimization_targets: Optional[Union[str, List[str]]] = None,
```
bpkroth, line 29: Why is this taking `None`?

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -78,34 +80,40 @@ def register(self, configurations: pd.DataFrame, scores: pd.DataFrame,
             Dataframe of configurations / parameters. The columns are parameter names and the rows are the configurations.
         scores : pd.DataFrame
             Scores from running the configurations. The index is the same as the index of the configurations.
-
         context : pd.DataFrame
-            Not Yet Implemented.
+            Not implemented yet.
+        metadata : pd.DataFrame
+            Implementaton depends on instance.
```
bpkroth, line 86:
```suggestion
            Implementation depends on instance (e.g., saved optimizer state to return).
```

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -78,34 +80,40 @@ def register(self, configurations: pd.DataFrame, scores: pd.DataFrame,
...
         """
         # Do some input validation.
-        assert set(scores.columns) == set(self._optimization_targets), \
-            "Mismatched optimization targets."
+        if type(self._optimization_targets) is str:
```
bpkroth, line 89:
```suggestion
        assert self._optimization_targets, "Missing or invalid optimization targets"
        if type(self._optimization_targets) is str:
```

bpkroth: Also assert not empty (see also comment above about accepting `None`)

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -78,34 +80,40 @@ def register(self, configurations: pd.DataFrame, scores: pd.DataFrame,
...
+            assert self._optimization_targets in scores.columns, "Mismatched optimization targets."
+        if type(self._optimization_targets) is list:
+            assert set(scores.columns) >= set(self._optimization_targets), "Mismatched optimization targets."
         assert self._has_context is None or self._has_context ^ (context is None), \
-            "Context must always be added or never be added."
+            "Metadata must always be added or never be added."
```
bpkroth, line 94: Incorrect comment.

```suggestion
            "Context must always be added or never be added."
```
Do you have other places to fix this?

mlos_core/mlos_core/optimizers/optimizer.py
jsfreischuetz: I don't think this makes sense given my comment above

mlos_core/mlos_core/optimizers/optimizer.py
bpkroth: Makes sense. But that's a separate change. Please submit a different PR for that so we can focus tests on that. Thanks!

bpkroth (changes requested):

mlos_core/mlos_core/optimizers/optimizer.py
bpkroth: OK, but separate PR for that one please

mlos_core/mlos_core/optimizers/flaml_optimizer.py
```diff
@@ -130,10 +135,10 @@ def _suggest(self, context: Optional[pd.DataFrame] = None) -> pd.DataFrame:
         if context is not None:
             warn(f"Not Implemented: Ignoring context {list(context.columns)}", UserWarning)
         config: dict = self._get_next_config()
-        return pd.DataFrame(config, index=[0])
+        return pd.DataFrame(config, index=[0]), None
 
     def register_pending(self, configurations: pd.DataFrame,
-                         context: Optional[pd.DataFrame] = None) -> None:
+                         metadata: Optional[pd.DataFrame] = None) -> None:
```
bpkroth, line 141: Restore `context`, add `metadata`

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -56,9 +56,11 @@ def __init__(self, *,
             raise ValueError("Number of weights must match the number of optimization targets")
 
         self._space_adapter: Optional[BaseSpaceAdapter] = space_adapter
-        self._observations: List[Tuple[pd.DataFrame, pd.DataFrame, Optional[pd.DataFrame]]] = []
+        self._observations: List[Tuple[pd.DataFrame, pd.DataFrame, Optional[pd.DataFrame], Optional[pd.DataFrame]]] = []
```
motus, line 59: this gets a bit unwieldy. Should we have a named tuple for `_pending_observations`, e.g.,
```python
class PendingObservation(NamedTuple):
    """A named tuple representing a pending observation."""

    configurations: pd.DataFrame
    context: Optional[pd.DataFrame]
    meta: Optional[pd.DataFrame]
```
and do the same for `_observations`? (not sure we can inherit `NamedTuple`s - most likely, we can't)

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -56,9 +56,11 @@ def __init__(self, *,
...
         self._has_context: Optional[bool] = None
-        self._pending_observations: List[Tuple[pd.DataFrame, Optional[pd.DataFrame]]] = []
+        self._pending_observations: List[Tuple[pd.DataFrame, Optional[pd.DataFrame], Optional[pd.DataFrame]]] = []
+        self.delayed_config: Optional[pd.DataFrame] = None
+        self.delayed_metadata: Optional[pd.DataFrame] = None
```
motus, line 63:
```suggestion
        self._delayed_config: Optional[pd.DataFrame] = None
        self._delayed_metadata: Optional[pd.DataFrame] = None
```
probably should be private, right?

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -78,34 +80,40 @@ def register(self, configurations: pd.DataFrame, scores: pd.DataFrame,
...
 
         if self._space_adapter:
             configurations = self._space_adapter.inverse_transform(configurations)
             assert configurations.shape[1] == len(self.optimizer_parameter_space.values()), \
                 "Mismatched configuration shape after inverse transform."
-        return self._register(configurations, scores, context)
+        return self._register(configurations, scores, metadata, context)
 
     @abstractmethod
     def _register(self, configurations: pd.DataFrame, scores: pd.DataFrame,
```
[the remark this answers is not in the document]
motus: Agree. Let's fix `_register` now and update the public `register` in the next PR

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -181,30 +207,34 @@ def register_pending(self, configurations: pd.DataFrame,
...
+    def get_observations(self) -> Tuple[pd.DataFrame, pd.DataFrame, Optional[pd.DataFrame], Optional[pd.DataFrame]]:
         """
-        Returns the observations as a triplet of DataFrames (config, score, context).
+        Returns the observations as a triplet of DataFrames (config, score, context, metadata).
 
         Returns
         -------
         observations : Tuple[pd.DataFrame, pd.DataFrame, Optional[pd.DataFrame]]
-            A triplet of (config, score, context) DataFrames of observations.
+            A triplet of (config, score, metadata) DataFrames of observations.
```
motus, line 223:
```suggestion
            A 4-tuple of (config, score, context, metadata) DataFrames of observations.
```
(or, better yet, a `NamedTuple`)

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -181,30 +207,34 @@ def register_pending(self, configurations: pd.DataFrame,
...
         """
         if len(self._observations) == 0:
             raise ValueError("No observations registered yet.")
-        configs = pd.concat([config for config, _, _ in self._observations]).reset_index(drop=True)
-        scores = pd.concat([score for _, score, _ in self._observations]).reset_index(drop=True)
+        configs = pd.concat([config for config, _, _, _ in self._observations]).reset_index(drop=True)
+        scores = pd.concat([score for _, score, _, _ in self._observations]).reset_index(drop=True)
```
motus, line 228: Yeah, let's have a `List[Observation]` NamedTuples - or, better yet, concatenate the dataframes right there in `_register` and forget about `List` and `NamedTuple`

mlos_core/mlos_core/optimizers/optimizer.py
motus: or, maybe, have `_observed_configs`, `_observed_scores`, `_observed_contexts` etc. and concatenate the dataframes instead of having a list of dataframes. I am pretty sure the schemas are the same from one `_register` call to the next

mlos_core/mlos_core/optimizers/utils.py
```diff
@@ -0,0 +1,36 @@
+import inspect
+from typing import Any, Callable, Dict, Optional
+import pandas as pd
+
+def to_metadata(metadata: Optional[pd.DataFrame]) -> Optional[List[pd.Series]]:
```
motus, line 5: docstring?

mlos_core/mlos_core/optimizers/utils.py
```diff
@@ -0,0 +1,36 @@
+import inspect
+from typing import Any, Callable, Dict, Optional
+import pandas as pd
+
```
motus, line 4:
```suggestion


```

mlos_core/mlos_core/optimizers/utils.py
```diff
@@ -0,0 +1,36 @@
...
+def to_metadata(metadata: Optional[pd.DataFrame]) -> Optional[List[pd.Series]]:
+    if metadata is None:
+        return None
+    return [idx_series[1] for idx_series in metadata.iterrows()]
+
```
motus, line 9:
```suggestion


```
I wonder why our linters don't complain here

mlos_core/mlos_core/optimizers/utils.py
```diff
@@ -0,0 +1,36 @@
+import inspect
```
motus, line 1:
```suggestion
#
# Copyright (c) Microsoft Corporation.
# Licensed under the MIT License.
#
"""
Docstring!
"""
import inspect
```
CI should not pass this - @bpkroth do you know why we are not detecting these style violations?

mlos_core/mlos_core/optimizers/utils.py
motus, line 1: It looks like we don't use this module anywhere - maybe remove it from this PR?

mlos_core/mlos_core/tests/optimizers/optimizer_multiobj_test.py
```diff
@@ -99,21 +99,29 @@ def objective(point: pd.DataFrame) -> pd.DataFrame:
...
+    else:
+        assert best_metadata is None
     assert best_context is None
     assert set(best_config.columns) == {'x', 'y'}
     assert set(best_score.columns) == {'main_score', 'other_score'}
     assert best_config.shape == (1, 2)
     assert best_score.shape == (1, 2)
 
-    (all_configs, all_scores, all_contexts) = optimizer.get_observations()
+    (all_configs, all_scores, all_metadata, best_context) = optimizer.get_observations()
```
motus, line 117: why `best_context`?? shouldn't it be `all_contexts` instead?

mlos_core/mlos_core/tests/optimizers/optimizer_multiobj_test.py
```diff
@@ -99,21 +99,29 @@ def objective(point: pd.DataFrame) -> pd.DataFrame:
...
     assert isinstance(all_configs, pd.DataFrame)
     assert isinstance(all_scores, pd.DataFrame)
-    assert all_contexts is None
+    if optimizer_class is OptimizerType.SMAC:
+        assert isinstance(all_metadata, pd.DataFrame) or all_metadata is None
+    else:
+        assert all_metadata is None
+    assert best_context is None
```
motus, line 124:
```suggestion
    assert all_contexts is None
```
?

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
```diff
@@ -48,7 +48,7 @@ def test_create_optimizer_and_suggest(configuration_space: CS.ConfigurationSpace
 
     assert optimizer.parameter_space is not None
 
-    suggestion = optimizer.suggest()
+    suggestion, _ = optimizer.suggest()
```
motus, line 51: here and in other places:
```suggestion
    suggestion, _metadata = optimizer.suggest()
```
it's better to spell out things

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
```diff
@@ -268,41 +266,39 @@ def objective(point: pd.DataFrame) -> pd.DataFrame:
...
         assert isinstance(best_config, pd.DataFrame)
         assert isinstance(best_score, pd.DataFrame)
-        assert best_context is None
         assert set(best_config.columns) == {'x', 'y'}
         assert set(best_score.columns) == {'score'}
 
-    (best_config, best_score, _context) = best_observation
-    (llamatune_best_config, llamatune_best_score, _context) = llamatune_best_observation
+    (best_config, best_score, _, _) = best_observation
+    (llamatune_best_config, llamatune_best_score, _metadata) = llamatune_best_observation
```
motus, line 291: I am confused. shouldn't it be 
```suggestion
    (llamatune_best_config, llamatune_best_score, _context, _metadata) = llamatune_best_observation
```
?
this should fail in tests

motus (commented):
Left a few comments. Let's have a meeting and discuss the changes. I am a bit confused about the purpose of the metadata and how it is different from the context

bpkroth:
> Left a few comments. Let's have a meeting and discuss the changes. I am a bit confused about the purpose of the metadata and how it is different from the context

It should be in the README I asked for and also some internal chats we had, but essentially, `metadata` is for internal optimizer saved state (e.g., something SMAC specific) and `context` is more of an index into a subspace of a larger optimization problem (e.g., optimize for VMs with 8GB RAM & 4 vCores and a transactional workload by reusing data from VMs with 4GB RAM and 2 vCores).

Issue #332, opened 2023-05-03
mlos_bench: add nightly runs of full end-to-end examples


bpkroth:
```[tasklist]
### Tasks
- [ ] Create resource group to hold resources
- [ ] Create service account to run things
- [ ] Stash service account's auth token in Github SECRETs for this repo
- [ ] Create CI rules that skip running that job when the SECRET isn't available
- [ ] Script the running of the jobs
- [ ] Make sure things are torn down when they fail
```

bpkroth:
#455 is probably a decent base for this

PR #754, opened 2024-06-12
Added missing 'shell_env_params' in redis.jsonc

DelphianCalamity:
These missing environment variables were causing the Redis benchmark to fail.


bpkroth (approved):
LGTM, thanks!
We'll eventually need to add end-to-end tests in #332, but not here.

PR #760, opened 2024-06-28
Require explicit arguments for mlos_core optimizers

jsfreischuetz:
This is a simple PR that makes all arguments explicit for optimizer-related function calls in preparation to add additional arguments in #751 and make it easier to review.

### DIFF
--- a/mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
+++ b/mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
@@ -117,7 +117,7 @@ class MlosCoreOptimizer(Optimizer):
 
         # TODO: Specify (in the config) which metrics to pass to the optimizer.
         # Issue: https://github.com/microsoft/MLOS/issues/745
-        self._opt.register(configs=df_configs, scores=df_scores[opt_targets].astype(float))
+        self._opt.register(configurations=df_configs, scores=df_scores[opt_targets].astype(float))
 
         if _LOG.isEnabledFor(logging.DEBUG):
             (score, _) = self.get_best_observation()
--- a/mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
+++ b/mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
@@ -117,7 +117,7 @@ class MlosCoreOptimizer(Optimizer):
 
         # TODO: Specify (in the config) which metrics to pass to the optimizer.
         # Issue: https://github.com/microsoft/MLOS/issues/745
-        self._opt.register(configurations=df_configs, scores=df_scores[opt_targets].astype(float))
+        self._opt.register(configs=df_configs, scores=df_scores[opt_targets].astype(float))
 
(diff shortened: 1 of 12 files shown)


mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
```diff
@@ -117,7 +117,7 @@ def bulk_register(self,
 
         # TODO: Specify (in the config) which metrics to pass to the optimizer.
         # Issue: https://github.com/microsoft/MLOS/issues/745
-        self._opt.register(df_configs, df_scores[opt_targets].astype(float))
+        self._opt.register(configurations=df_configs, scores=df_scores[opt_targets].astype(float))
```
bpkroth, line 120:
```suggestion
        self._opt.register(configs=df_configs, scores=df_scores[opt_targets].astype(float))
```

Proposal to also shorten this for brevity to just `configs` everywhere.

@motus do you agree?

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -231,7 +231,7 @@ def cleanup(self) -> None:
         Redefine this method in optimizers that require cleanup.
         """
 
-    def _from_1hot(self, config: npt.NDArray) -> pd.DataFrame:
+    def _from_1hot(self, *, config: npt.NDArray) -> pd.DataFrame:
```
bpkroth, line 234:
```suggestion
    def _from_1hot(self, config: npt.NDArray) -> pd.DataFrame:
```

Don't think we need to do this here - it only takes a single argument.

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -254,7 +254,7 @@ def _from_1hot(self, config: npt.NDArray) -> pd.DataFrame:
                     j += 1
         return pd.DataFrame(df_dict)
 
-    def _to_1hot(self, config: Union[pd.DataFrame, pd.Series]) -> npt.NDArray:
+    def _to_1hot(self, *, config: Union[pd.DataFrame, pd.Series]) -> npt.NDArray:
```
bpkroth, line 257:
```suggestion
    def _to_1hot(self, config: Union[pd.DataFrame, pd.Series]) -> npt.NDArray:
```

Same here, though again, I'm not against leaving it.

mlos_core/mlos_core/optimizers/optimizer.py
bpkroth: I suppose it might be confusing if `context` needs to be 1 hot encoded in the future and we want to reuse this routine, but we can also address that in the future.

bpkroth (changes requested):
I'd like to change the `configurations` parameter to `configs` at the same time for brevity and make sure the tests pass, but other than that LGTM.

bpkroth (approved):
LGTM, thanks!

bpkroth:
Ignoring the Windows CI issue for now - #768

PR #762, opened 2024-06-28
Staged Makefile Improvements

bpkroth:
- Introduce a new `make format` target that currently only calls `licenseheaders` but will call `isort` and `black` in the future
- Introduce new variables to help make dependency tracking easier
- Simplify `pytest` rules and allow running only a single module at a time.

This PR is a precursor to pyproject.toml related build changes as well.

PR #763, opened 2024-06-28
Makefile improvements with incremental rebuild support for check targets

bpkroth:
Builds off of #762 

Change rules to only feed changed files since the last successful rebuild to various check targets.
This should help with incremental rebuilds.


bpkroth:
I'm not quite satisfied with this one - there's some issues of it not detecting the need to rescan the whole set of files after conda rebuilds for instance - so leaving it in draft mode for now.

PR #770, opened 2024-07-01
Add Metadata to optimizers

jsfreischuetz:
Adds metadata to respond from suggest, and be passable into register.

This is in support of adding multi-fidelity support (#751)

### DIFF
--- a/mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
+++ b/mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
@@ -117,7 +117,7 @@ class MlosCoreOptimizer(Optimizer):
 
         # TODO: Specify (in the config) which metrics to pass to the optimizer.
         # Issue: https://github.com/microsoft/MLOS/issues/745
-        self._opt.register(configs=df_configs, scores=df_scores[opt_targets].astype(float))
+        self._opt.register(configurations=df_configs, scores=df_scores[opt_targets].astype(float))
 
         if _LOG.isEnabledFor(logging.DEBUG):
             (score, _) = self.get_best_observation()
--- a/mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
+++ b/mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
@@ -117,7 +117,7 @@ class MlosCoreOptimizer(Optimizer):
 
         # TODO: Specify (in the config) which metrics to pass to the optimizer.
         # Issue: https://github.com/microsoft/MLOS/issues/745
-        self._opt.register(configurations=df_configs, scores=df_scores[opt_targets].astype(float))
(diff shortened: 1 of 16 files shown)


mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
```diff
@@ -180,7 +180,7 @@ def suggest(self) -> TunableGroups:
         tunables = super().suggest()
         if self._start_with_defaults:
             _LOG.info("Use default values for the first trial")
-        df_config = self._opt.suggest(defaults=self._start_with_defaults)
+        df_config, _ = self._opt.suggest(defaults=self._start_with_defaults)
```
bpkroth, line 183:
```suggestion
        df_config, _metadata = self._opt.suggest(defaults=self._start_with_defaults)
```

mlos_core/mlos_core/tests/optimizers/bayesian_optimizers_test.py
```diff
@@ -34,7 +34,7 @@ def test_context_not_implemented_warning(configuration_space: CS.ConfigurationSp
         optimization_targets=['score'],
         **kwargs
     )
-    suggestion = optimizer.suggest()
+    suggestion, _ = optimizer.suggest()
```
bpkroth, line 37:
```suggestion
    suggestion, _metadata = optimizer.suggest()
```

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -120,7 +120,7 @@ def _register(self, *, configs: pd.DataFrame, scores: pd.DataFrame,
         """
         pass    # pylint: disable=unnecessary-pass # pragma: no cover
 
-    def suggest(self, *, context: Optional[pd.DataFrame] = None, defaults: bool = False) -> pd.DataFrame:
+    def suggest(self, *, context: Optional[pd.DataFrame] = None, defaults: bool = False) -> Tuple[pd.DataFrame, Optional[pd.DataFrame]]:
```
bpkroth, line 123: Mostly looks ok, though am wondering if we should have this `Tuple[DataFrame, DataFrame]` a named data structure as well.

jsfreischuetz: I thought about this, but if we make that change, it has to make a quite significant change to the overall API, since then how you call register should change too, and

bpkroth: Yeah, I think that's where I landed as well.  We can always tweak it later.

bpkroth (approved):

mlos_core/mlos_core/optimizers/optimizer.py
bpkroth, line 123: @jsfreischuetz I think you have lint issue here (line too long)

PR #774, opened 2024-07-03
Enable black, isort, and doc formatters and checks

bpkroth:
Follow on work to #766.

This enabled both formatters and applies their changes to the repo.

Additionally, since `black` does not make changes to comments nor docstrings, we also enable `docformatter` to reformat docstrings which better aligns with `pydocstyle` rules as well.
Without this additional change (and some manual fixups), `pycodestyle` and `pylint` would still complain about line lengths, for instance.

Finally, we make a minor adjustment to the max line length setting it to 99 (which is also accepted and mentioned in pep8) instead of 88 to avoid some comment (especially linter overrides) wrapping.


pyproject.toml
```diff
@@ -1,9 +1,16 @@
 [tool.black]
-line-length = 88
+line-length = 99
```
bpkroth, line 2: pep8 actually has a note that 99 is also acceptable, and I think works better for larger screens, and less of a nusance for us to change to, so considering this a compromise

PR #775, opened 2024-07-03
bugfix: MySQL can round microseconds into the future causing scheduler to skip trials

### LINKED ISSUE #756: mlos_bench: Trial repetition
I would like to run each config 30 times but not 30 at same time. For example, a default config should be running 30 times but not concurrently. 

Currently trial repeat param for mlos_bench will run the trial repeat number all at the same ts_start

motus:
Solution: truncate microseconds of the current timestamp when scheduling new trials to be executed ASAP

Closes #756

### DIFF
--- a/mlos_bench/mlos_bench/storage/sql/experiment.py
+++ b/mlos_bench/mlos_bench/storage/sql/experiment.py
@@ -253,7 +253,8 @@ class Experiment(Storage.Experiment):
 
     def new_trial(self, tunables: TunableGroups, ts_start: Optional[datetime] = None,
                   config: Optional[Dict[str, Any]] = None) -> Storage.Trial:
-        # bugfix: MySQL can round microseconds into the future causing scheduler to skip trials.
+        # MySQL can round microseconds into the future causing scheduler to skip trials.
+        # Truncate microseconds to avoid this issue.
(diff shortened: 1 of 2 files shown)


mlos_bench/mlos_bench/storage/sql/experiment.py
```diff
@@ -253,7 +253,9 @@ def _get_config_id(self, conn: Connection, tunables: TunableGroups) -> int:
 
     def new_trial(self, tunables: TunableGroups, ts_start: Optional[datetime] = None,
                   config: Optional[Dict[str, Any]] = None) -> Storage.Trial:
-        ts_start = utcify_timestamp(ts_start or datetime.now(UTC), origin="local")
+        # MySQL can round microseconds into the future causing scheduler to skip trials.
+        # Truncate microseconds to avoid this issue.
+        ts_start = utcify_timestamp(ts_start or datetime.now(UTC), origin="local").replace(microsecond=0)
```
bpkroth, line 258: Can we please add a test for this?

bpkroth: Future PR is fine for now so we can get this thru.

bpkroth (approved):
LGTM

Test failures seem unrelated

PR #777, opened 2024-07-07
patch for managed identity storage authentication

DelphianCalamity:
MLOS can no longer work with `storage account keys` for storage authentication because account key access must be disabled for security reasons. Users must enable storage authentication with managed identities instead.

That's a temporary patch that switches to managed identity authentication.
Users must create a managed identity and assign to it proper "contributor" roles that allow it to manage the file share.
Then they must pass the managed identity's info `managedIdentityClientId` and `managedIdentityName` to MLOS as global parameters. 

The managed identity must be assigned to the MLOS host in order for the `local file share service` to work.
```bash 
az vm identity assign -g <resource-group> -n <vm-name> --identities <managed-identity-name>
```

Implementation Details:
- The managed identity must be assigned to all provisioned VMs.
Currently, I assign the identity using the REST API right after the VM has been provisioned. A better approach would be to assign the identity directly using the ARM template. The reason I did not use the better/easier approach is that I was somehow tricked after reading [this](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/how-to-configure-managed-identities?pivots=qs-configure-portal-windows-vm#assign-a-user-assigned-identity-during-the-creation-of-a-vm) into thinking that it was not supported! I intend to switch soon.

- [Mounting file shares doesn't work with managed identities](https://learn.microsoft.com/en-us/answers/questions/1620270/how-do-i-mount-an-azure-file-share-in-azure-contai). The VMs provisioned by MLOS access scripts uploaded by the host by mounting the file share. My temporary workaround is to download/upload files using the functionality from `azure-file-share-service` instead which I updated to use the Managed Identity Credential. To make that script available to the provisioned VMs I packaged it and [uploaded it to PyPI ](https://pypi.org/project/azure-file-share-service/). Then, scripts booting/setting up the VM must install the package (e.g. in boot-ubuntu.jsonc) in order to be able to interact with the file share. The proper, long-term solution would be to enable mounting again after adding to it support for managed identities.

I updated the config files for the `RedisBench` experiment and verified that the flow works.

Opening this PR to raise concern about the deprecation of storage account keys and how this currently breaks MLOS (no longer working because retrieving the keys is prohibited). Will be using this patch for my personal work until there is a more formal solution.

### DIFF
--- a/mlos_bench/mlos_bench/config/cli/azure-redis-bench.jsonc
+++ b/mlos_bench/mlos_bench/config/cli/azure-redis-bench.jsonc
@@ -14,9 +14,9 @@
     "environment": "environments/root/root-azure-redis.jsonc",
     "storage": "storage/sqlite.jsonc",
 
-    "globals": [
-        "global_config_azure.jsonc"
-    ],
+    // "globals": [
+    //     "global_config_azure.jsonc"
+    // ],
 
     "teardown": false,
 
--- a/mlos_bench/mlos_bench/config/environments/apps/redis/redis.jsonc
+++ b/mlos_bench/mlos_bench/config/environments/apps/redis/redis.jsonc
@@ -51,19 +51,22 @@
                         "trial_id"
                     ],
                     "setup": [
-                        "$mountPoint/$experiment_id/$trial_id/scripts/setup-workload.sh",
-                        "$mountPoint/$experiment_id/$trial_id/scripts/setup-app.sh"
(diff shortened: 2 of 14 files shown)


motus (commented):
This is very helpful! There is more to be done for this update (fix the unit tests, configs, etc.), but that's a great start. I'll split this PR into several smaller updates and work on them together with @DelphianCalamity 
We'll close that PR once all its functionality is in master. For now I'll mark it as draft

PR #779, opened 2024-07-09
Use token-based authentication instead of the access key in the AzureFileShareService

motus:
Make `AzureFileShareService` class use token-based credential instead of the access key. This PR is part of #777

**NOTE:**
* We have to use late initialization of the `_share_client` to avoid issues with JSON schema validation tests.
* More PRs will follow:
   * Use to `azcopy` on the remote VMs instead of mounting the file share using the access key
   * Remove references to `"storageAccountKey"` from configurations and tests and document the new mechanism


bpkroth:
Part of why we went with the `mount` smb method was reduced dependencies on the target VM.

Use of `azcopy` instead might be a little more challenging on that front, but let's see.

bpkroth:
As long as we're removing account storage key, does removing fileshare also make sense?  We could use blobs instead I guess.

mlos_bench/mlos_bench/services/remote/azure/azure_fileshare.py
```diff
@@ -52,23 +53,30 @@ def __init__(
...
-            credential=self.config["storageAccountKey"],
-        )
+        self._share_client: Optional[ShareClient] = None
+
+    def _get_share_client(self) -> ShareClient:
+        """Get the Azure file share client object."""
+        if self._share_client is None:
+            assert self._parent is not None and isinstance(
+                self._parent, SupportsAuth
+            ), "Authorization service not provided. Include service-auth.jsonc?"
```
bpkroth, line 70: I think that check should be in `__init__` for earlier feedback on bad configs.

bpkroth (approved):
Mostly looks good. Only feedback was to move the `SupportsAuth` check to `__init__` so we can check for proper configs at startup time rather than at file transfer need time which might be after some experiment trial has already run.

mlos_bench/mlos_bench/services/remote/azure/azure_fileshare.py
motus: I wish I could move all that functionality to `__init__()` where it actually belongs! Unfortunately, that breaks our unit tests that instantiate each Service class with test configurations to validate the JSONs. `_get_share_client()` is just a workaround for this, just like a bunch of other methods with delayed initialization in other Services, Environments, etc.

bpkroth: Proposed fix to that:
https://github.com/motus/MLOS/pull/14

bpkroth: (we already do that in one or two other places)

PR #786, opened 2024-07-10
Workaround pylance issue with pyproject.toml related changes and pip editable modules install format

bpkroth:
See Also: 
- https://github.com/microsoft/pylance-release/issues/3473

May also affect `mypy`:
- https://github.com/python/mypy/issues/16988
- https://github.com/python/mypy/issues/12313


bpkroth:
Slurped into #783

PR #787, opened 2024-07-10
Update azure credentials to be more flexible

eujing:
These changes are mainly to add flexibility to the auth service in supporting azure authenticating via:

- A personal identity via az CLI for the local dev environment. This will still run into expiring issues, but is probably fine for local dev.
- A personal or managed identity in an azure hosted environment. The flexibility to use the managed identity allows us to avoid the expiring issue without needing a separate SP, key-vault and cert setup, aligning with new security recommendations.


motus (approved):
Makes sense! Let's merge it in

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
```diff
@@ -64,9 +64,9 @@ def __init__(self,
         self._access_token = "RENEW *NOW*"
         self._token_expiration_ts = datetime.now(UTC)  # Typically, some future timestamp.
 
-        # Login as ourselves
-        self._cred: Union[azure_id.AzureCliCredential, azure_id.CertificateCredential]
-        self._cred = azure_id.AzureCliCredential()
+        # Login as the first identity available, usually ourselves or a managed identity
```
bpkroth, line 67: I like the direction, but I don't follow how this will allow ManagedIdentity to work?  Can you provide some additional information, documentation links, comments, etc?

DelphianCalamity: @bpkroth my understanding is that the [DefaultCredential](https://learn.microsoft.com/en-us/dotnet/api/azure.identity.defaultazurecredential?view=azure-dotnet) is more generic that the `AzureCliCredential`. The first one comprises a collection of various credential classes (azure cli and managed identities included) and then it will try to authenticate by `trying out` all of them one by one. 

If the VM running mlos is assigned to a managed identity (as is the case in my dev setup) the managed identity authentication will succeed.

I personally integrated this PR to avoid doing `az login` all the time!

eujing: Yes exactly what @DelphianCalamity mentioned! @bpkroth the chain of authentication methods in `DefaultCredential` is the recommended way to supporting both the local dev environment (e.g. cli credentials) and "production" environment (e.g. managed identity) with minimal code changes.

If I remember correctly, assigning a managed identity to a VM sets some azure-specific environment variables in the VM that `DefaultCredential` is able to pickup and authenticate as early in the chain, otherwise it goes and tries other methods until one is available, like `AzureCliCredential`.

motus (approved):

mlos_bench/mlos_bench/services/remote/azure/azure_auth.py
bpkroth: Kk.  Next time, please include a documentation reference either in the source code or the PR so that it's in the history.  Thanks!

PR #789, opened 2024-07-10
Fix the coercion of scores to floats in the optimizer

### LINKED ISSUE #785: Resume experiments fail when trying to bulk register previous metrics to the optimizer
During bulk register, existing metrics from the results db are loaded as strings as that is how they are stored.

However, applying `self._adjust_signs_df` to the string target columns results in un-intended results, especially when in a `max` case.
The string column *= -1 results in empty strings, which fail to cast into `float` later on in bulk register.
This issue probably never turned up in `min` cases as the string columns *= 1 has no effect.

The casting to float should probably be done on all target columns before adjusting signs.

image

Example output from the debug terminal:
image

motus:
Closes #785 

Also, add more unit tests to make sure the optimizer handles string inputs correctly.


mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
```diff
@@ -104,20 +104,19 @@ def bulk_register(self,
...
             pd.DataFrame([{} if score is None else score for score in scores]))
 
-        opt_targets = list(self._opt_targets)
         if status is not None:
             # Select only the completed trials, set scores for failed trials to +inf.
             df_status = pd.Series(status)
             # TODO: Be more flexible with values used for failed trials (not just +inf).
             # Issue: https://github.com/microsoft/MLOS/issues/523
-            df_scores.loc[df_status != Status.SUCCEEDED, opt_targets] = float("inf")
+            df_scores[df_status != Status.SUCCEEDED] = float("inf")
```
bpkroth, line 112: What kind of tests can we put in for this?  For backwards compatibility and forwards bug fix checks?

motus: good point - I'll add tests to this PR

mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
```diff
@@ -127,11 +126,12 @@ def bulk_register(self,
...
         """
-        In-place adjust the signs of the scores for MINIMIZATION problem.
+        Coerce optimization target scores to floats
+        and adjust the signs for MINIMIZATION problem.
         """
-        for (opt_target, opt_dir) in self._opt_targets.items():
-            df_scores[opt_target] *= opt_dir
-        return df_scores
+        # NOTE: We use `.astype()` instead of `.apply(pd.to_numeric)`
+        # to catch incorrect values early.
```
bpkroth, line 133: So, if this errors, it's basically a problem in the user script, right?

Can we handle this error in order to expose that comment directly to the user so they get some feedback on what to do?

For instance:

```python
try:
  ret = df_scores[list(self._opt_targets)].astype(float) * self._opt_targets.values()
except SomeExceptionThatPandasThrowsOnConversionErrors:
  _LOG.error("Failed to convert scores to floats. Double check parse script to ensure valid values are being provided: %s", json.dumps(the_scores)))
  raise e
```

mlos_bench/mlos_bench/tests/optimizers/mlos_core_opt_df_test.py
```diff
@@ -81,3 +85,38 @@ def test_df(mlos_core_optimizer: MlosCoreOptimizer, mock_configs: List[dict]) ->
...
+    """
+    Test `MlosCoreOptimizer._adjust_signs_df()` on different types of inputs.
+    """
+    df_scores_input = pandas.DataFrame({
+        'latency': [88.88, 66.66, 99.99, None],
+        'throughput': [111, 222, 333, None],
+    })
+
+    df_scores_output = pandas.DataFrame({
+        'latency': [88.88, 66.66, 99.99, float("NaN")],
```
bpkroth, line 112: Maybe add some strings as well?

mlos_bench/mlos_bench/tests/optimizers/mlos_core_opt_df_test.py
```diff
@@ -81,3 +85,38 @@ def test_df(mlos_core_optimizer: MlosCoreOptimizer, mock_configs: List[dict]) ->
...
+        'throughput': [-111, -222, -333, float("NaN")],
+    })
+
+    # Make sure we adjust the signs for minimization.
+    df_scores = mlos_core_optimizer._adjust_signs_df(df_scores_input)
+    assert df_scores.equals(df_scores_output)
+
+    # Check that the same operation works for string inputs.
+    df_scores = mlos_core_optimizer._adjust_signs_df(df_scores_input.astype(str))
+    assert df_scores.equals(df_scores_output)
```
bpkroth, line 122: Maybe also add a test for exception being thrown on bad inputs.

mlos_bench/mlos_bench/tests/optimizers/mlos_core_opt_df_test.py
bpkroth, line 112: Add a comment on how we except "NaN" to be handled?

mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
motus: In this particular case, the problem is not just in the user script but also in our code that allowed such data to be stored in the database. I will create another PR with more checks there

mlos_bench/mlos_bench/tests/optimizers/mlos_core_opt_df_test.py
motus: I have that in the next test, but I'll add a few more

mlos_bench/mlos_bench/tests/optimizers/mlos_core_opt_df_test.py
motus: done

mlos_bench/mlos_bench/tests/optimizers/mlos_core_opt_df_test.py
motus: yep. I will add a few more unit tests on that

mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
bpkroth: Since that's been pretty common, and it's not always clear what to do on a random exception, whether it's user based or internal issue, I think still some warning that describes at least "please ensure the scores are valid floats" would be a good idea.

bpkroth: A comment in the code that explains that it might be due to bad values being mistakenly stored in the DB could also be helpful.

motus: done. there are unit tests, too

mlos_bench/mlos_bench/tests/optimizers/mlos_core_opt_df_test.py
```diff
@@ -74,3 +78,85 @@ def test_df(mlos_core_optimizer: MlosCoreOptimizer, mock_configs: List[dict]) ->
...
+            pandas.DataFrame(
+                {
+                    "latency": ["INVALID"],
+                    "throughput": ["no input"],
+                }
+            )
+        )
+
+    with pytest.raises(ValueError):
+        mlos_core_optimizer._adjust_signs_df(
```
bpkroth, line 155:
```suggestion
        mlos_core_optimizer._adjust_signs_df(  # pylint: disable=protected-access
```

Missing this?  Or should we do the test on a public method?

bpkroth (approved):
Minor comment about a test and pylint protected access complaints vs. public method test. Otherwise, LGTM :)

mlos_bench/mlos_bench/tests/optimizers/mlos_core_opt_df_test.py
bpkroth: Oh nm, you did that at the top of this file.  All good :)

PR #782, opened 2024-07-09
bugfix: is_defaults() is a method, not property

motus:
Closes #781

PR #790, opened 2024-07-11
Test that the config provided is serializable

bpkroth:
Test for #782

Makes `new_trial` a wrapper with a sanity check and renames the previous `new_trial` abstractmethod to `_new_trial`.


mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -308,6 +308,36 @@ def new_trial(self, tunables: TunableGroups, ts_start: Optional[datetime] = None
                 An object that allows to update the storage with
                 the results of the experiment trial run.
             """
+            # Check that `config` is json serializable (e.g., no callables)
+            if config:
+                try:
+                    _config = DictTemplater(config).expand_vars()
```
bpkroth, line 314: Quick hack - makes use of the fact that DictTemplater only allows primitive types throughout a nested dict structure, though we don't actually care about expanding anything.

bpkroth:
Note: this needs #791 to merge before unrelated tests can pass.

motus:
dunno, I'd rather simply check these values in the concrete SQL-specific implementation of the method. also, the issue that Yaseen ran into was with the scores, not config parameters' values

motus (approved):

PR #796, opened 2024-07-12
_update_status was called twice leading to mysql.connector.errors.IntegrityError

DelphianCalamity:
tweaked the `update_telemetry` function so that it doesn't update the status again.


motus:
This is a benign warning that we issue when re-writing the existing status with the same timestamp. It happens in the synchronous scheduler when we update the telemetry and final status of the trial, and they have the same timestamp (end of the trial). We can downgrade the severity of the message to DEBUG, but we definitely should not comment out the update. I'll close it

Issue #805, opened 2024-07-18
LlamaTune inverse_trainsform() produces invalid configuration


motus:
`LlamaTuneAdapter.inverse_transform()` apparently produces small `float` values like `1.5` and `0.998` and assigns them to `int` hyperparameters with range `[1, 10000]`.
If we add validation immediately after producing such config, it will fail, causing the corresponding unit test failure:

`FAILED mlos_bench/mlos_bench/tests/optimizers/llamatune_opt_test.py::test_llamatune_optimizer - ConfigSpace.exceptions.IllegalValueError: Value 1.5: (<class 'float'>) is not allowed for hyperparameter dim_0, Type: UniformInteger, Range: [1, 10000], Default: 5000`

Interestingly, in ConfigSpace 1.0 same `float` inputs get scaled up from `[0, 1]` to `int` range of the corresponding hyperparameter. That also produces invalid configuration because value of `1.5` got scaled to `15000`, i.e., outside of the hyperparameter's `[1, 10000]` range.

motus:
@kkanellis can you please take a look at it? I am bit lost in LlamaTune code. Use #806 to reproduce the problem, then run
```bash
pytest -vv -n1 mlos_bench/mlos_bench/tests/optimizers/llamatune_opt_test.py::test_llamatune_optimizer
```
Thank you!

PR #806, opened 2024-07-18
Validate the configuration produced by LlamaTune inverse_transform()

motus:
Code to reproduce #805

### DIFF
--- a/mlos_core/mlos_core/spaces/adapters/llamatune.py
+++ b/mlos_core/mlos_core/spaces/adapters/llamatune.py
@@ -35,7 +35,7 @@ class LlamaTuneAdapter(BaseSpaceAdapter):  # pylint: disable=too-many-instance-a
     discretization is used.
     """
 
-    def __init__(
+    def __init__(  # pylint: disable=too-many-arguments
         self,
         *,
         orig_parameter_space: ConfigSpace.ConfigurationSpace,
--- a/mlos_core/mlos_core/spaces/adapters/llamatune.py
+++ b/mlos_core/mlos_core/spaces/adapters/llamatune.py
@@ -129,6 +129,11 @@ class LlamaTuneAdapter(BaseSpaceAdapter):  # pylint: disable=too-many-instance-a
                 if getattr(self, "_pinv_matrix", None) is None:
                     self._try_generate_approx_inverse_mapping()
 
+                # FIXME: Give some variables names for debugging.
+                config_scaler = self._config_scaler
+                q_scaler = self._q_scaler
+                pinv_matrix = self._pinv_matrix
+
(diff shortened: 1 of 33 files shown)


mlos_core/mlos_core/spaces/adapters/llamatune.py
bpkroth, line 1: Somewhat unrelated, we should add "quantization + llamatune" tests as well.

mlos_core/mlos_core/spaces/adapters/llamatune.py
```diff
@@ -139,6 +139,8 @@ def inverse_transform(self, configurations: pd.DataFrame) -> pd.DataFrame:
                     self.target_parameter_space,
                     vector=target_config_vector,
                 )
+                self.target_parameter_space.check_configuration(target_config)
```
bpkroth, line 142: Might just need to clip them to the range in the case that it's too large (e.g., due to quantization issues).
Are there other failure cases than that?

bpkroth: I took a slightly closer look at this.
`max_unique_values_per_param` is forcing the lower dimensions space to be `UniformIntegerHyperparatmers`, but the `vector` produced that is then dot producted with the approximate inverted matrix returns floats.

So, I think there's an additional scaling operation required to put that vector back into the `max_unique_values_per_param` range when that option `is not None`.

bpkroth: Ok, I've added the q_scaler inverse transform here as well, which makes the values look much better, but one is still out of bounds.  I think it's maybe just needing clipping either there, or in a previous step.

bpkroth: @kkanellis

bpkroth:
```sh
pytest -n0 mlos_bench/mlos_bench/tests/optimizers/llamatune_opt_test.py
```

to reproduce

mlos_core/mlos_core/spaces/adapters/llamatune.py
```diff
@@ -261,7 +313,7 @@ def _transform(self, configuration: dict) -> dict:
             # Clip value to force it to fall in [0, 1]
             # NOTE: HeSBO projection ensures that theoretically but due to
             #       floating point ops nuances this is not always guaranteed
-            value = max(0.0, min(1.0, norm_value))  # pylint: disable=redefined-loop-name
+            value = max(0.0, min(1.0, norm_value))
```
bpkroth, line 316: unnecessary pylint exception?

bpkroth:
Taking this over as a fix as well in a new PR.

PR #819, opened 2024-07-24
Fixups to AzureFileShareService auth token 

### LINKED ISSUE #818: Fix issues in AzureFileShareService introduced by #779
1. `_get_share_client` is now only called once, but needs to be refreshed on demand using the `AuthService`
2. This change caused tests to try and authenticate, but we need to mock that for tests:

              Proposed fix to that:
https://github.com/motus/MLOS/pull/14

_Originally posted by @bpkroth in https://github.com/microsoft/MLOS/pull/779#discussion_r1688806116_

bpkroth:
1. Make `_get_share_client` refresh the auth service token on demand.
2. Use the `MockAuthService` in config loading tests and require `SupportsAuth` `Service` parent at config load time for early feedback.

Closes #818


motus:
I assume this update went to #820

bpkroth:
> I assume this update went to #820

Parts of it did, but I don't think all of it.  Especially the refresh logic.

@eujing I see that you added some of the test fixes in.  Have you tested #820 to confirm that it already handles credential refresh as necessary?

eujing:
> > I assume this update went to #820
> 
> Parts of it did, but I don't think all of it. Especially the refresh logic.
> 
> @eujing I see that you added some of the test fixes in. Have you tested #820 to confirm that it already handles credential refresh as necessary?

Yes I have been using the #820 changes, along with #837 for long running experiments. So at least for `DefaultAzureCredential` with the MI setup, the token refresh has been handled as expected.

PR #820, opened 2024-07-25
Proposed fixes for token-based auth in azure fileshare service

eujing:
To use token-based authentication for `ShareClient`, I think we should be passing in the credential object derived from `TokenCredential` (in our case, some instance of `DefaultAzureCredential`.

Previously, we were passing specific string tokens to the `credential` argument, which is being intepreted as a SAS token.
This leads to errors like: `azure.core.exceptions.ClientAuthenticationError: Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.`

[ShareClient](https://learn.microsoft.com/en-us/python/api/azure-storage-file-share/azure.storage.fileshare.shareclient?view=azure-python) documentation on the `credential` argument.

By passing in the whole `TokenCredential` object, I believe `ShareClient` will manage the token lifecycle and we won't need to do so as mentioned in #818.


eujing:
Want to add that I haven't been able to test this fully on my side yet.
Currently getting a permissions error, so I am trying to grant myself the `Storage File Data Privileged Contributor` role to see if that works. Currently only have overall `Contributor` role to the storage account.

`azure.core.exceptions.HttpResponseError: This request is not authorized to perform this operation using this permission.
RequestId:15e8cf1a-301a-0068-0618-de1302000000
Time:2024-07-24T22:29:41.5702312Z
ErrorCode:AuthorizationPermissionMismatch`

mlos_bench/mlos_bench/services/types/azure_authenticator_type.py
```diff
@@ -0,0 +1,24 @@
...
+#
+"""Protocol interface for authentication for the Azure services."""
+
+from typing import Protocol, runtime_checkable
+from mlos_bench.services.types.authenticator_type import SupportsAuth
+import azure.core.credentials as azure_cred
+
+
+@runtime_checkable
+class SupportsAzureAuth(SupportsAuth, Protocol):
```
bpkroth, line 13: Hmm, not sure I love this approach, but it might be the best option since 

Another option:
Change `SupportsAuth` to add `get_credential` to return something like `Union[str, "azure_cred.TokenCredential"]`

Where `str` is really only there to make the `Union` part happy, and
`azure_cred.TokenCredential` is in `"` so that we can do the import hidden behind a block like this:

```python
if TYPE_CHECKING:
    import azure.core.credentials as azure_cred
```

Then we could add AWS's credential object type to that `Union` as well, but it still might require an `assert` on the caller end to make the type checker happy about the specific type it received and that's messy too.

Another option would be to still just return the string and then convert it back to the appropriate Credential type as necessary, but that has to happen at all callsites, which still might make sense if different clients need different credentials types (token vs credentia, etc.)

There are also various test mocks that will also need to be fixed up though.
Have a look at #818 for some additional related fixups that are required.

eujing: I agree those are possible options too, this was mainly to satisfy mypy checking. I had the impression `SupportsAuth` was supposed to be cloud agnostic, so did not want to introduce the azure-specific types there.

Actually another option might be to use a generic protocol instead like `SupportsAuth[TokenCredential]`, what do you think? The generic could then be used for other similar cloud credential objects too, but I need to check if we can do runtime asserts on `SupportsAuth[TokenCredential]`

eujing: Oh no I just tried, it turns out `isinstance` cannot be used to check for `SupportsAuth[TokenCredential]` explicitly, only `SupportsAuth`. But in the latter case, I guess it would be possible to use an incompatible auth service with the fileshare without triggering the assert.



The mypy error is `Parameterized generics cannot be used with class or instance checksmypy[misc](https://mypy.readthedocs.io/en/stable/_refs.html#code-misc)`

eujing: @bpkroth @motus just pushed a change, making `SupportsAuth` generic to support cloud-agnostic credential objects.
Let me know what you think of this approach.

Also incorporated the fixes from #818 for the relevant test mocks.

motus (commented):
Looks good, although I suspect that we need to make sure that the credential has been properly initialized in `get_credential()` implementation - just like we do in `get_auth_token()`. I've pushed an update along with a few cosmetic fixes to https://github.com/eujing/MLOS/pull/1 - @eujing please take a look

motus (approved):

mlos_bench/mlos_bench/services/remote/azure/azure_fileshare.py
```diff
@@ -60,20 +61,25 @@ def __init__(
...
     def _get_share_client(self) -> ShareClient:
         """Get the Azure file share client object."""
         if self._share_client is None:
-            assert self._parent is not None and isinstance(
-                self._parent, SupportsAuth
-            ), "Authorization service not provided. Include service-auth.jsonc?"
+            credential = self._auth_service.get_credential()
+            assert isinstance(
+                credential, TokenCredential
+            ), f"Expected a TokenCredential, but got {type(credential)} instead."
```
bpkroth, line 76: This is still a late error that I was trying to turn into a config load error in #819

bpkroth: Alright, I made a new change to #819 to handle that now.
From what I can tell the rest of this one is good, but it'd be good to get confirmation that `DefaultCredential` doesn't need refreshed, especially with the SP part.

PR #821, opened 2024-07-25
MLOS_VIZ: View panel!

yshady:
OSS Version of experiment view panel


yshady:
Should we have this under mlos_viz?

yshady:
WIll close and re org

mlos_viz/backend.py
yshady, line 1: @motus  @bpkroth  @eujing  this might wanna be deleted but not too horrible to just ignore honestly

PR #822, opened 2024-07-26
mlos_vizpanel: A Gui for mlos experiments

yshady:
Visualize mlos experiments with an intuitive GUI


yshady:
@bpkroth @motus  This should make more sense than having the viz panel in mlos_viz 
Should be separate folders for better organization :)

Issue #826, opened 2024-07-31
Update mlos_viz to support matplotlib >= 3.9.* 


motus:
New version of matplotlib seem to have breaking changes in the API

P.S. A speculation: also, a new version of matplotlib seems to pull in numpy 2.* binaries (?) that break LlamaTune tests, e.g.,
https://github.com/microsoft/MLOS/actions/runs/10187168908/job/28181896008?pr=802

PR #827, opened 2024-08-01
Restrict matplotlib version to < 3.9 for Windows conda env

bpkroth:
Temporary fix to conda env to make it match the `pip` `setup.py` restrictions for now.

Eventually, we'll wan to relax these, along with increased `numpy` versions, etc.

See Also #826


motus (approved):

conda-envs/mlos-windows.yml
```diff
@@ -16,7 +16,8 @@ dependencies:
   - jupyter
   - ipykernel
   - nb_conda_kernels
-  - matplotlib
+  - matplotlib<3.9
+  - matplotlib-base<3.9
```
bpkroth, line 20: @motus do we need `pyqt` at all?
https://conda-forge.org/docs/maintainer/knowledge_base/#matplotlib

motus: good question! Maybe we can get by with just `matplotlib-base` here and in `setup.py`? The reason I've added `matplotlib-base` explicitly was because without it our environment ends up with `matplotlib=3.9.1` and `matplotlib-base=3.8.4` both installed, and that led to conflicts in (i believe) common binaries

motus: anyway, let me experiment with it now. stay tuned! :)

bpkroth: `matplotlib-base` would only be in `conda` - it's not a `pypi` package

motus: yep, figured that out already. everything seems to work with `matplotlib-base` only

conda-envs/mlos-windows.yml
```diff
@@ -16,7 +16,8 @@ dependencies:
   - jupyter
   - ipykernel
   - nb_conda_kernels
-  - matplotlib
+  - matplotlib<3.9
```
motus, line 19:
```suggestion
```
It looks like we can use just `matplotlib-base` here. I'll send a PR shortly

motus (approved):

motus (approved):

PR #802, opened 2024-07-17
Migrate to ConfigSpace 1.*

motus:
Summary of changes:
* [x] Update dependencies to ConfigSpace >= 1.0.
* [x] Update dependencies to SMAC3 >= 2.2.
* [x] Migrate to new ConfigSpace API.
* [x] Better definition of ConfigSpace fixture in unit tests.
* [x] Fix mypy and pylint issues triggered by the migration.

Notes:
* Side effect: **Works with Python 3.12!**
* ~Supersedes #799~
* ~Depends on upcoming SMAC3 release with [SMAC3 #1124](https://github.com/automl/SMAC3/pull/1124) merged in and a new release cut with it.~
* Quantization will be added back later:
    * #803 
* ~Tests fail due to LlamaTune issue:~
    * #805 

Closes #727

PR #829, opened 2024-08-01
Add CI build for Python 3.12

motus:
This PR enables Python 3.12 in Windows environment and does not restrict Python version in the default environments on Windows and Linux. Note that although `mlos-3.12.yml` exists, its usage is still commented out in `linux.yml`, because Python 3.12 is still not available on GitHub CI. Issue #832 opened to track it

P.S. Also, we now use a lighter `matplotlib-base` instead of `matplotlib` proper

~To be merged after #802~


conda-envs/mlos-windows.yml
```diff
@@ -16,14 +16,14 @@ dependencies:
   - jupyter
   - ipykernel
   - nb_conda_kernels
-  - matplotlib
+  - matplotlib<2.9
+  - matplotlib-base<2.9
```
bpkroth, line 20: I think this might need to be 3.9, not 2.9 since that's what's in `mlos_viz/setup.py`

Anyways, hopefully goes away once dabl cuts a new version with my fixes there.

motus: ooh right

.github/workflows/linux.yml
```diff
@@ -37,6 +37,7 @@ jobs:
           - "3.9"
           - "3.10"
           - "3.11"
+          - "3.12"
```
motus, line 40:
```suggestion
          # - "3.12"
```
Temporarily disable Python 3.12 for Linux CI

.github/workflows/linux.yml
```diff
@@ -37,6 +37,7 @@ jobs:
           - "3.9"
           - "3.10"
           - "3.11"
+          # - "3.12"
```
motus, line 40: It looks like Linux CI environment on GitHub only has Python 3.12.0.rc3 which causes conflicts with other packages. We will enable 3.12 once a proper 3.12.0 release is available on GitHub CI. #832 opened to track the issue

conda-envs/mlos.yml
```diff
@@ -15,14 +15,13 @@ dependencies:
   - jupyter
   - ipykernel
   - nb_conda_kernels
-  - matplotlib
+  - matplotlib-base
```
bpkroth, line 18: Does this still allow plotting to work correctly?

.github/workflows/linux.yml
bpkroth: I left some comments in the issue.  I think we might want to consider switching to conda-forge as the priority everywhere again to let this work.

bpkroth:
BTW, some of this might be eased by finishing some fixups for dabl so that we can release other constraints on numpy and whatnot.

.github/workflows/linux.yml
bpkroth: I want to see the error message so I pushed a small change to re-enable this for now.

conda-envs/mlos.yml
motus: yes, I've just checked it locally. There was a weird unit test failure on the first run after I had rebuilt the conda env:
`FAILED mlos_viz/mlos_viz/tests/test_dabl_plot.py::test_dabl_plot - _tkinter.TclError: couldn't read file "C:/Users/sergiym/AppData/Local/anaconda3/envs/mlos/Library/lib/tcl8.6/auto.tcl": no such file or directory`
when I checked, the tcl file was there and I could not reproduce the failure again. Maybe this file is generated the first time matplotlib is used and there was some race condition when I first ran the unit tests.

.github/workflows/linux.yml
bpkroth: FWIW, we can also do this locally like so:

```sh
make CONDA_ENV_NAME=mlos-3.12 conda-env
```

bpkroth: Indeed it was a conda-forge channel priority issue.

bpkroth: Switching all of the conda env files over to that for consistency.

bpkroth (approved):

bpkroth:
Should mark the new one as required in the CI settings once this goes through.

bpkroth:
> Should mark the new one as required in the CI settings once this goes through.

Did this just now too.

PR #834, opened 2024-08-06
Fixups to example notebook for new APIs

bpkroth:
- [x] enable notebook execution in CI pipelines to prevent these mistakes in the future


bpkroth:
Should really enable this for some of the example repos as well.

motus (approved):

PR #837, opened 2024-08-08
Add managed identities support to RG setup scripts

eujing:
- Creates the managed identity (MI) in the specified RG. If existing, operation is idempotent.
- Assigns the MI as a user-assigned MI to the created control VM
- Grants the MI required roles for usage with the storage account file shares.

TODO:

- [ ] Add support to the powershell script too.
- [ ] Consider MI access to the results DB too (part of incoming security wave requirements).

### DIFF
--- a/mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/rg-template.json
+++ b/mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/rg-template.json
@@ -363,6 +363,13 @@
         "kvName": {
             "type": "string",
             "value": "[variables('kvName')]"
+        },
+        "storageAccountNames": {
+            "type": "array",
+            "copy": {
+                "count": "[length(parameters('storageAccountLocations'))]",
+                "input": "[concat(variables('storageAccountName'), parameters('storageAccountLocations')[copyIndex()])]"
+            }
         }
     }
 }
--- a/mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.sh
+++ b/mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.sh
@@ -19,6 +19,12 @@ $script_name
     --resultsDbArmParamsFile someResultsDbArmParamsFile.json
     --servicePrincipalName someServicePrincipalName
     --certName someCertName
+OR
+$script_name
+    --resourceGroupName someRgName
+    --controlPlaneArmParamsFile someControlPlaneArmParamsFile.json
+    --resultsDbArmParamsFile someResultsDbArmParamsFile.json
+    --managedIdentityName someManagedIdentityName
 USAGE
     exit 1
 }
@@ -36,7 +42,12 @@ case $1 in
         resultsDbArmParamsFile="$2"
         shift 2
         ;;
-    # Other params
+    # Managed Identity params
+    --managedIdentityName)
+        managedIdentityName="$2"
+        shift 2
+        ;;
+    # Service Principal params
     --servicePrincipalName)
         servicePrincipalName="$2"
         shift 2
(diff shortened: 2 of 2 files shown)


mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/README.md
bpkroth, line 1: Has any of this been developed with an eye towards rectifying an existing setup?

eujing: @bpkroth that is a good point. Unfortunately I have only tested this with setting up from scratch.
However, in theory I think this should work for existing resource groups.

Assuming the existing RG was set up with these scripts, I think all the az cli calls (need to confirm) are idempotent.
For example, using the same ARM template and parameters should give us back the required details of the resources without changing them too much (same naming conventions)

Then the new blocks of logic for managed identity should then be apply to those existing resources (VMs, storage accounts, RG role)

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/README.md
```diff
@@ -29,23 +29,42 @@ The *control plane RG* is a container for the *persistent* resources of MLOS (re
...
     ./setup-rg.ps1 `
         -controlPlaneArmParamsFile $controlPlaneArmParamsFile `
-        -resultsDbArmParamsFile $resultsDbArmParamsFile  # If provisioning results DB, otherwise omit `
-        -servicePrincipalName $servicePrincipalName `
         -resourceGroupName $resourceGroupName `
+        -resultsDbArmParamsFile $resultsDbArmParamsFile `
+        -servicePrincipalName $servicePrincipalName `
         -certName $certName
+
+    # If setting up with a Managed Identity
```
bpkroth, line 42: Can you give more context in the README about this part?
When should one use a Managed Identity? How does one set it up, etc.

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/README.md
```diff
@@ -29,23 +29,42 @@ The *control plane RG* is a container for the *persistent* resources of MLOS (re
...
+    ./setup-rg.ps1 `
+        -controlPlaneArmParamsFile $controlPlaneArmParamsFile `
+        -resourceGroupName $resourceGroupName `
+        -resultsDbArmParamsFile $resultsDbArmParamsFile `
+        -managedIdentityname $managedIdentityName
```
bpkroth, line 47:
```suggestion
        -managedIdentityName $managedIdentityName
```

consistent camelCasing please

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/setup-rg.ps1
```diff
@@ -11,14 +11,17 @@ param(
...
     [Parameter(Mandatory=$False)]
     [string] $resultsDbArmParamsFile,
-    # Other params
-    [Parameter(Mandatory=$True)]
-    [string] $servicePrincipalName,
     [Parameter(Mandatory=$True)]
     [string] $resourceGroupName,
-    [Parameter(Mandatory=$True)]
+    # Managed Identity params
+    [Parameter(Mandatory=$True, ParameterSetName="ByMI")]
```
bpkroth, line 17: Are using ServicePrincipals and ManagedIdentity mutually exclusive?

mlos_bench/mlos_bench/config/services/remote/azure/scripts/setup-rg/README.md
bpkroth: I think we might need to add end-to-end tests to validate these finally, though maybe in a separate PR.

PR #851, opened 2024-08-20
Move quantization_bins into meta field of NumericalHyperaparameter

bpkroth:


mlos_core/mlos_core/tests/spaces/adapters/llamatune_test.py
```diff
@@ -322,15 +348,15 @@ def test_special_parameter_values_biasing() -> None:  # pylint: disable=too-comp
...
-    ][100]
+    assert (1 - eps) * int(num_configs * bias_percentage / 2) <= (
+        special_values_instances["int_1"][1]
+    )
+    assert (1 - eps) * int(num_configs * bias_percentage / 2) <= (
+        special_values_instances["int_2"][2]
+    )
+    assert (1 - eps) * int(num_configs * bias_percentage * 1.5) <= (
+        special_values_instances["int_2"][100]
+    )
```
bpkroth, line 359: Minor reformatting - unrelated.

mlos_bench/mlos_bench/optimizers/convert_configspace.py
```diff
@@ -140,7 +143,10 @@ def _tunable_to_configspace(
     if tunable.quantization_bins:
         # Temporary workaround to dropped quantization support in ConfigSpace 1.0
         # See Also: https://github.com/automl/ConfigSpace/issues/390
-        monkey_patch_quantization(range_hp, tunable.quantization_bins)
+        new_meta = dict(range_hp.meta or {})
+        new_meta[QUANTIZATION_BINS_META_KEY] = tunable.quantization_bins
+        range_hp.meta = new_meta
+        monkey_patch_hp_quantization(range_hp)
```
bpkroth, line 149: Part of me thinks this monkey patching could be entirely moved to the `mlos_core` and all we should have to do in `mlos_bench` is set `"quantization_bins"` in `meta` and let `mlos_core` handle the rest.

bpkroth: @motus - the refinement I was thinking of

bpkroth: Actually, I moved it just now into the `BaseOptimizer` class.  Have a look and lemme know what you think.  At a high level I think this works, but probably needs more tests.

mlos_bench/mlos_bench/optimizers/convert_configspace.py
```diff
@@ -77,6 +77,10 @@ def _tunable_to_configspace(
     meta: Dict[Hashable, TunableValue] = {"cost": cost}
     if group_name is not None:
         meta["group"] = group_name
+    if tunable.is_numerical and tunable.quantization_bins:
+        # Temporary workaround to dropped quantization support in ConfigSpace 1.0
+        # See Also: https://github.com/automl/ConfigSpace/issues/390
+        meta[QUANTIZATION_BINS_META_KEY] = tunable.quantization_bins
```
bpkroth, line 83: From here, `mlos_core.Optimizer` should be able to apply the `monkey_patch_cs_quantization` in the base class and `mlos_bench` basically shouldn't need to be aware of it.

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -44,6 +45,9 @@ def __init__(
         space_adapter : BaseSpaceAdapter
             The space adapter class to employ for parameter space transformations.
         """
+        # Temporary workaround to dropped quantization support in ConfigSpace 1.0
+        # See Also: https://github.com/automl/ConfigSpace/issues/390
+        monkey_patch_cs_quantization(parameter_space)
         self.parameter_space: ConfigSpace.ConfigurationSpace = parameter_space
         self.optimizer_parameter_space: ConfigSpace.ConfigurationSpace = (
             parameter_space if space_adapter is None else space_adapter.target_parameter_space
```
bpkroth, line 53: Hmm, not entirely certain how LlamaTune will interact with this though ...

mlos_bench/mlos_bench/optimizers/convert_configspace.py
bpkroth: As discussed in https://github.com/bpkroth/MLOS/pull/60, this isn't actually true.
Not all underlying optimizers actually use ConfigSpace🤦, so we'd have to quantize at the Tunables wrapper layer and that might interact poorly with the underlying optimizers if we end up not actually registering the originally suggested value with it, but maybe that's something we can experiment with and address in the future as well.

motus (approved):

mlos_core/mlos_core/spaces/converters/util.py
```diff
@@ -4,36 +4,83 @@
...
         ConfigSpace hyperparameter to patch.
-    quantization_bins : int
-        Number of bins to quantize the hyperparameter into.
+
+    Returns
+    -------
+    hp : Hyperparameter
+        Patched hyperparameter.
     """
+
```
motus, line 31:
```suggestion
```

PR #811, opened 2024-07-23
Introduce Named Data Classes for mlos_core

jsfreischuetz:
Introduces `@dataclass` types for return values from mlos_core.Optimizers to improve code readability.

See also: https://github.com/microsoft/MLOS/pull/751

PR #852, opened 2024-08-21
Rewrite for named tuples

jsfreischuetz:
## Title

Refactor mlos_core APIs to encapsulate related data fields.

## Description

Refactors the mlos_core Optimizer APIs to accept new data types `Observation`, `Observations` and return `Suggestion`, instead of a mess of `Tuple[DataFrame, DataFrame, Optional[DataFrame], Optional[DataFrame]]` that must be named and checked everywhere.

Additionally, this makes it more explicit that `_register` is a bulk operation that is not actually supported currently by the underlying optimizers, though leaves notes on how we can do that in the future.

## Type of Change

- Refactor

---

## Testing

Usual CI plus some new unit tests for new data type operations.

---

## Additional Notes

A more significant rewrite of named tuple support inside mlos_core.
This is based on comments in #811 as well as conversations with @bpkroth 

---


mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+    if type(left) != type(right):
+        return False
+    elif left is not None and right is not None:
+        if not left.equals(right):
+            return False
+    return True
+
+
+def compare_optional_dataframe(
+    left: Optional[pd.DataFrame], right: Optional[pd.DataFrame]
```
bpkroth, line 40:
```suggestion
    left: Optional[pd.DataFrame], right: Optional[pd.DataFrame],
```

nit: add a trialing comma and let black reformat to multiple lines

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+        The right Series to compare
+
+    Returns
+    -------
+    bool
+        Compare the equality of two Optional[pd.Series] objects
+    """
+    if type(left) != type(right):
+        return False
+    elif left is not None and right is not None:
```
bpkroth, line 33:
```python
    if isinstance(left, pd.Series) and isinstance(right, pd.Series):
        return left.equals(right)
    return (left is None and right is None)
```

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+        context: Optional[pd.Series] = None,
+        metadata: Optional[pd.Series] = None,
+    ):
+        self.config = config
+        self.context = context
+        self.metadata = metadata
+
+    def complete(self, score: pd.Series) -> Observation:
+        """
+        Completes the suggestion.
```
bpkroth, line 320:
```suggestion
        Completes the suggestion by adding a score.
```

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+            The observation of the suggestion.
+        """
+        return Observation(
+            config=self.config,
+            score=score,
+            context=self.context,
+            metadata=self.metadata,
+        )
+
+    def config_to_configspace(self, space: ConfigurationSpace) -> Configuration:
```
bpkroth, line 339:
```suggestion
    def to_configspace_config(self, space: ConfigurationSpace) -> Configuration:
```

?

Since it's actually a Suggestion atm.

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+            return False
+        if not compare_optional_series(self.context, other.context):
+            return False
+        if not compare_optional_series(self.metadata, other.metadata):
+            return False
+
+        return True
+
+    def __ne__(self, other: Any) -> bool:
+        return not self.__eq__(other)
```
bpkroth, line 361: Is this necessary?  Or is it automatically handled?

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+                config=self.config.loc[[idx]],
+                score=self.score.loc[[idx]],
+                context=None if self.context is None else self.context.loc[[idx]],
+                metadata=None if self.metadata is None else self.metadata.loc[[idx]],
+            )
+
+    def __repr__(self) -> str:
+        return (
+            f"Observation(config={self.config}, score={self.score},"
+            + " context={self.context}, metadata={self.metadata})"
```
bpkroth, line 275:
```suggestion
            " context={self.context}, metadata={self.metadata})"
```

Python doesn't need the + since the () wrap the expression already.

bpkroth: I usually also like to leave the trailing space on the line before (nit).

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+            ), "config and metadata must have the same index"
+        else:
+            assert metadata is None, (
+                "context of appending observation must be null"
+                + " if metadata of prior observation is null"
+            )
+
+    def to_list(self) -> List[Observation]:
+        """
+        Converts the observations to a list of observations.
```
bpkroth, line 243:
```suggestion
        Converts the Observations object to a list of Observation objects.
```

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+                + " if metadata of prior observation is not null"
+            )
+            self.metadata = pd.concat([self.metadata, metadata]).reset_index(drop=True)
+            assert self.config.index.equals(
+                self.metadata.index
+            ), "config and metadata must have the same index"
+        else:
+            assert metadata is None, (
+                "context of appending observation must be null"
+                + " if metadata of prior observation is null"
```
bpkroth, line 238:
```suggestion
                " if metadata of prior observation is null"
```

same comment about the trailing whitespace on the previous line here.

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+    config : pd.Series
+        The configuration observed.
+    score : pd.Series
+        The score metrics observed.
+    context : Optional[pd.Series]
+        The context in which the configuration was evaluated.
+        Not Yet Implemented.
+    metadata: Optional[pd.Series]
+        The metadata in which the configuration was evaluated
+        Not Yet Implemented.
```
bpkroth, line 80:
```suggestion
```

Metadata is implemented, right?

bpkroth: Or is metadata dropped once it's an Observation and only relevant when it's a Suggestion?

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+    """
+
+    def __init__(
+        self,
+        *,
+        config: pd.Series = pd.Series(),
```
bpkroth, line 86:
```suggestion
        config: pd.Series,
```

I think that's generally considered to be a bad idea to do since it makes a single reference to all new instances.

Better to just add asserts in the constructor to require the value to be provided.

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+        score: pd.Series = pd.Series(),
+        context: Optional[pd.Series] = None,
+        metadata: Optional[pd.Series] = None,
+    ):
+        self.config = config
```
bpkroth, line 91: Is there a reason to leave these mutable?

Shouldn't they be `self._config` and provide (ideally read-only) accessors instead?

mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
```diff
@@ -198,10 +199,10 @@ def suggest(self) -> TunableGroups:
         tunables = super().suggest()
         if self._start_with_defaults:
             _LOG.info("Use default values for the first trial")
-        df_config, _metadata = self._opt.suggest(defaults=self._start_with_defaults)
+        suggestion = self._opt.suggest(defaults=self._start_with_defaults)
         self._start_with_defaults = False
-        _LOG.info("Iteration %d :: Suggest:\n%s", self._iter, df_config)
-        return tunables.assign(configspace_data_to_tunable_values(df_config.loc[0].to_dict()))
+        _LOG.info("Iteration %d :: Suggest:\n%s", self._iter, suggestion.config)
+        return tunables.assign(configspace_data_to_tunable_values(suggestion.config.to_dict()))
```
bpkroth, line 205: Ahh, that's much more readable actually :)

mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
```diff
@@ -221,18 +222,20 @@ def register(
...
             )
         return registered_score
 
     def get_best_observation(
         self,
     ) -> Union[Tuple[Dict[str, float], TunableGroups], Tuple[None, None]]:
-        (df_config, df_score, _df_context) = self._opt.get_best_observations()
-        if len(df_config) == 0:
+        best_observations = self._opt.get_best_observations()
+        if len(best_observations.config.index) == 0:
```
bpkroth, line 236:
```suggestion
        if len(best_observations.config) == 0:
```

?

mlos_bench/mlos_bench/tests/optimizers/toy_optimization_loop_test.py
```diff
@@ -40,10 +41,13 @@ def _optimize(env: Environment, opt: Optimizer) -> Tuple[float, TunableGroups]:
             # pylint: disable=protected-access
             if isinstance(opt, MlosCoreOptimizer) and isinstance(opt._opt, SmacOptimizer):
                 config = tunable_values_to_configuration(tunables)
-                config_df = config_to_dataframe(config)
+                config_df = config_to_series(config)
```
bpkroth, line 44:
```suggestion
                config_series = config_to_series(config)
```

bpkroth: Though, see comment below as well.

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
+#
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+#
+"""Contains classes for observations and suggestions."""
+
+from __future__ import annotations
```
bpkroth, line 7: Unnecesssary?

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+
+from typing import Any, Iterator, List, Optional
+
+import pandas as pd
+from ConfigSpace import Configuration, ConfigurationSpace
+
+
+def compare_optional_series(left: Optional[pd.Series], right: Optional[pd.Series]) -> bool:
```
bpkroth, line 15: This should probably go in `mlos_core/util.py`

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+        score: pd.Series = pd.Series(),
+        context: Optional[pd.Series] = None,
+        metadata: Optional[pd.Series] = None,
+    ):
+        self.config = config
+        self.score = score
+        self.context = context
+        self.metadata = metadata
+
+    def to_suggestion(self) -> Suggestion:
```
bpkroth, line 96:
```suggestion
    def to_suggestion(self) -> "Suggestion":
```

bpkroth: That should let you work around the order of class definitions.

bpkroth: You can add something like the following at the top as well (in case of splitting things out):

```python
if TYPE_CHECKING:
    from mlos_core.data_types.suggestion import Suggestion
```

mlos_core/mlos_core/optimizers/observations.py
bpkroth: Maybe you "just" need to use the quoted form of class names when giving type hints?

e.g.,
```python
def get_suggestion(...) -> "Suggestion":
   ...
```

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
+#
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+#
+"""Contains classes for observations and suggestions."""
```
bpkroth, line 5: I wonder if we can split this out a bit more, or there's just too many circular import issues.

```
mlos_core/data_types/
  __init__.py
  observation.py
  observations.py
  suggestion.py
```

mlos_core/mlos_core/optimizers/observations.py
bpkroth: It'd make it easier to discover and useful for other needs.

bpkroth: Also easier to add independent tests.

mlos_core/mlos_core/optimizers/observations.py
bpkroth, line 1: These new classes and functions all need tests it looks like.

mlos_core/mlos_core/optimizers/observations.py
bpkroth: Also nicer for locating tests.

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+            assert len(config.index) == len(
+                score.index
+            ), "config and score must have the same length"
+
+            if sum([obs.context is None for obs in observations]) == 0:
+                context = pd.concat([obs.context.to_frame().T for obs in observations])  # type: ignore
+            else:
+                context = None
+            if sum([obs.metadata is None for obs in observations]) == 0:
+                metadata = pd.concat([obs.metadata.to_frame().T for obs in observations])  # type: ignore
```
bpkroth, line 152:
```suggestion
                metadata = pd.concat([obs.metadata.to_frame().T for obs in observations])  # type: ignore[specific-rule]
```

Let's not do blanket ignores at the very least, being more specific makes it more clear what's going on here.

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+        score: pd.DataFrame = pd.DataFrame(),
+        context: Optional[pd.DataFrame] = None,
+        metadata: Optional[pd.DataFrame] = None,
+        observations: List[Observation] = [],
+    ):
+
+        if len(observations) > 0:
+            config = pd.concat([obs.config.to_frame().T for obs in observations])
+            score = pd.concat([obs.score.to_frame().T for obs in observations])
+            assert len(config.index) == len(
```
bpkroth, line 143:
```suggestion
            assert (len(config.index) == len(score.index)), "config and score must have same length"
```

This is a nit, but I wonder if adding additional parens convinces black to keep the comparison part on the same line and split on the message.

mlos_core/mlos_core/optimizers/bayesian_optimizers/bayesian_optimizer.py
```diff
@@ -5,57 +5,38 @@
...
     """Abstract base class defining the interface for Bayesian optimization."""
 
     @abstractmethod
-    def surrogate_predict(
-        self,
-        *,
-        configs: pd.DataFrame,
-        context: Optional[pd.DataFrame] = None,
-    ) -> npt.NDArray:
+    def surrogate_predict(self, *, suggestion: Suggestion) -> npt.NDArray:
```
bpkroth, line 19: Maybe `*` is not necessary if `surrogate_predict` only takes on value now.

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -269,63 +272,49 @@ def _dummy_target_func(config: ConfigSpace.Configuration, seed: int = 0) -> None
...
+                f"Not Implemented: Ignoring context {list(observation.context.index)}",
+                UserWarning,
             )
-            value = TrialValue(cost=list(score.astype(float)), time=0.0, status=StatusType.SUCCESS)
-            self.base_optimizer.tell(info, value, save=False)
+
+        # Retrieve previously generated TrialInfo (returned by .ask()) or create
+        # new TrialInfo instance
+        config = ConfigSpace.Configuration(
+            self.optimizer_parameter_space, values=observation.config.dropna().to_dict()
```
bpkroth, line 299:
```suggestion
            self.optimizer_parameter_space, values=observation.config.dropna().to_dict(),
```
let black reformat with trailing comma

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -269,63 +272,49 @@ def _dummy_target_func(config: ConfigSpace.Configuration, seed: int = 0) -> None
...
+        )
+        info: TrialInfo = self.trial_info_map.get(
+            config,
+            TrialInfo(config=config, seed=self.base_optimizer.scenario.seed),
+        )
+        value = TrialValue(
+            cost=list(observation.score.astype(float)), time=0.0, status=StatusType.SUCCESS
```
bpkroth, line 306: same black trailing comma comment here

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -269,63 +272,49 @@ def _dummy_target_func(config: ConfigSpace.Configuration, seed: int = 0) -> None
...
 
-    def _register(
-        self,
-        *,
-        configs: pd.DataFrame,
-        scores: pd.DataFrame,
-        context: Optional[pd.DataFrame] = None,
-        metadata: Optional[pd.DataFrame] = None,
-    ) -> None:
+    def _register(self, *, observation: Observation) -> None:
```
bpkroth, line 275: Same `*` comment here now that all of the values are named in the observation.

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -73,201 +75,182 @@ def space_adapter(self) -> Optional[BaseSpaceAdapter]:
     def register(
         self,
         *,
-        configs: pd.DataFrame,
-        scores: pd.DataFrame,
-        context: Optional[pd.DataFrame] = None,
-        metadata: Optional[pd.DataFrame] = None,
+        observations: Optional[Observations] = None,
+        observation: Optional[Observation] = None,
```
bpkroth, line 79:
```suggestion
        observations: Optional[Union[Observation|Observations]] = None,
```

Proposal: accept `observations` as either a single `Observation` or an `Observations` and automatically convert to `Observations` when you find an `Observation`.

That allows you to:

- drop the `*` from the `register()` signature to simplify it
- drop `register_single()` or else turn it into an private internal implementation detail.

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -73,201 +75,182 @@ def space_adapter(self) -> Optional[BaseSpaceAdapter]:
...
+            observations is not None and observation is None
+        ), "Only one of observations or observation can be provided."
 
-        context : pd.DataFrame
-            Not Yet Implemented.
+        if observation is not None:
+            self.register_single(observation=observation)
+        if observations is not None:
+            for obs in observations.to_list():
+                self.register_single(observation=obs)
```
bpkroth, line 100: Whether this is a for loop over singleton registations or some other method of say rebuilding the internal model in bulk should be an implementation detail left up to the various backends, not done for all Optimizers in the wrapper class.

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -73,201 +75,182 @@ def space_adapter(self) -> Optional[BaseSpaceAdapter]:
...
 
-        metadata : Optional[pd.DataFrame]
-            Not Yet Implemented.
+    def register_single(self, *, observation: Observation) -> None:
```
bpkroth, line 102: See above.

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -73,201 +75,182 @@ def space_adapter(self) -> Optional[BaseSpaceAdapter]:
...
+            The observation to register.
         """
         pass  # pylint: disable=unnecessary-pass # pragma: no cover
 
     def suggest(
         self,
         *,
-        context: Optional[pd.DataFrame] = None,
+        context: Optional[pd.Series] = None,
         defaults: bool = False,
```
bpkroth, line 158: Can we make context a `data_type` too?

mlos_core/mlos_core/optimizers/observations.py
bpkroth: I think we can, though for the conversion methods, you'd have to place the `from ... import` call inside the method and tell `pylint: disable=not-top-level-import`, in order to avoid `cyclic-import` issues.

mlos_core/mlos_core/optimizers/observations.py
bpkroth: You can also do something like the following if necessary:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from mlos_core.optimizers.data_types.suggestion import Suggestion
```

But I think the quoted version + the internal import like I described above should be fine.

mlos_core/mlos_core/optimizers/observations.py
bpkroth: Yes, I recognize that returning a reference to an object is still mutable, but at least helps the type checkers along the way towards catching unintended write-errors and would make it easier to return a `.copy()` instead.

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+    config : pd.DataFrame
+        Pandas dataframe containing configurations. Column names are the parameter names.
+    score : pd.DataFrame
+        The score metrics observed in a dataframe.
+    context : Optional[pd.Series]
+        The context in which the configuration was evaluated.
+        Not Yet Implemented.
+    metadata: Optional[pd.Series]
+        The metadata in which the configuration was evaluated
+        Not Yet Implemented.
```
bpkroth, line 127: same comment as above about metadata being Not Yet Implemented

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+            if sum([obs.metadata is None for obs in observations]) == 0:
+                metadata = pd.concat([obs.metadata.to_frame().T for obs in observations])  # type: ignore
+            else:
+                metadata = None
+
+        assert len(config.index) == len(score.index), "config and score must have the same length"
+        self.config = config
+        self.score = score
+        self.context = context
+        self.metadata = metadata
```
bpkroth, line 160: Similar comment about private members and public accessors.

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+        index : pd.Index
+            The indices to filter by.
+
+        Returns
+        -------
+        Observation
+            The filtered observation.
+        """
+        return Observations(
+            config=self.config.loc[index],
```
bpkroth, line 177:
```suggestion
            config=self.config.loc[index].copy(),
```

?

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+            score=self.score.loc[index],
+            context=None if self.context is None else self.context.loc[index],
+            metadata=None if self.metadata is None else self.metadata.loc[index],
+        )
+
+    def append(self, other: Observation) -> None:
```
bpkroth, line 183: May need a way to append a whole set of Obervations in bulk as well.

mlos_core/mlos_core/optimizers/observations.py
```diff
@@ -0,0 +1,361 @@
...
+
+        return True
+
+    def __ne__(self, other: Any) -> bool:
+        return not self.__eq__(other)
+
+
+class Suggestion:
+    """
+    A single suggestion for a configuration.
```
bpkroth, line 299:
```suggestion
    A single suggestion for a configuration.
    
    A Suggestion is an Observation that has not yet been scored.
    Evaluating the Suggestion and calling `complete(scores)` can convert it to an Observation. 
```

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -73,201 +75,182 @@ def space_adapter(self) -> Optional[BaseSpaceAdapter]:
...
 
     @abstractmethod
     def _register(
         self,
         *,
-        configs: pd.DataFrame,
-        scores: pd.DataFrame,
-        context: Optional[pd.DataFrame] = None,
-        metadata: Optional[pd.DataFrame] = None,
+        observation: Observation,
```
bpkroth, line 142: See comment above about accepting multiple here and leaving the implementation of whether to loop or not up to the backend.

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -276,14 +259,17 @@ def get_best_observations(
...
-        if len(self._observations) == 0:
+        observations = self.get_observations()
+        if len(observations) == 0:
             raise ValueError("No observations registered yet.")
-        (configs, scores, contexts) = self.get_observations()
-        idx = scores.nsmallest(n_max, columns=self._optimization_targets, keep="first").index
-        return (configs.loc[idx], scores.loc[idx], None if contexts is None else contexts.loc[idx])
+
+        idx = observations.score.nsmallest(
+            n_max, columns=self._optimization_targets, keep="first"
```
bpkroth, line 270:
```suggestion
            n_max, columns=self._optimization_targets, keep="first",
```

same nit

mlos_core/mlos_core/spaces/adapters/adapter.py
```diff
@@ -74,15 +74,15 @@ def inverse_transform(self, configurations: pd.DataFrame) -> pd.DataFrame:
 
         Parameters
         ----------
-        configurations : pd.DataFrame
-            Dataframe of configurations / parameters, which belong to the original parameter space.
+        configurations : pd.Series
+            Series of configurations / parameters, which belong to the original parameter space.
```
bpkroth, line 78:
```suggestion
            A Series of configuration parameters, which belong to the original parameter space.
```

mlos_core/mlos_core/spaces/adapters/adapter.py
```diff
@@ -44,28 +44,28 @@ def target_parameter_space(self) -> ConfigSpace.ConfigurationSpace:
         pass  # pylint: disable=unnecessary-pass # pragma: no cover
 
     @abstractmethod
-    def transform(self, configuration: pd.DataFrame) -> pd.DataFrame:
+    def transform(self, configuration: pd.Series) -> pd.Series:
```
bpkroth, line 47: Trying to decide if we should still have a bulk version of this or use it as an `apply` over a DataFrame's rows.

mlos_core/mlos_core/spaces/adapters/adapter.py
```diff
@@ -74,15 +74,15 @@ def inverse_transform(self, configurations: pd.DataFrame) -> pd.DataFrame:
...
+        configurations : pd.Series
+            Series of configurations / parameters, which belong to the original parameter space.
             The columns are the parameter names the original parameter space and the
             rows are the configurations.
 
         Returns
         -------
-        configurations : pd.DataFrame
-            Dataframe of the translated configurations / parameters.
+        configurations : pd.Series
```
bpkroth, line 84:
```suggestion
        configuration : pd.Series
```

mlos_core/mlos_core/tests/optimizers/optimizer_multiobj_test.py
```diff
@@ -93,39 +95,41 @@ def objective(point: pd.DataFrame) -> pd.DataFrame:
...
+        suggestion = optimizer.suggest()
+        assert isinstance(suggestion, Suggestion)
+        assert isinstance(suggestion.config, pd.Series)
+        assert suggestion.metadata is None or isinstance(suggestion.metadata, pd.Series)
+        assert set(suggestion.config.index) == {"x", "y"}
         # Check suggestion values are the expected dtype
-        assert isinstance(suggestion.x.iloc[0], np.integer)
-        assert isinstance(suggestion.y.iloc[0], np.floating)
+        assert isinstance(suggestion.config["x"], int)  # type: ignore
+        assert isinstance(suggestion.config["y"], float)  # type: ignore
```
bpkroth, line 105: Does `np.integer` not work here anymore?  Would be nice to drop the type ignore (or at least improve it's specificity)

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
```diff
@@ -87,8 +88,9 @@ def test_basic_interface_toy_problem(
         # number of max iterations.
         kwargs["max_trials"] = max_iterations * 2
 
-    def objective(x: pd.Series) -> pd.DataFrame:
-        return pd.DataFrame({"score": (6 * x - 2) ** 2 * np.sin(12 * x - 4)})
+    def objective(x: float) -> pd.Series:
+        ret: pd.Series = pd.Series({"score": (6 * x - 2) ** 2 * np.sin(12 * x - 4)})
```
bpkroth, line 92: Is the type hint necessary?

mlos_core/mlos_core/spaces/adapters/adapter.py
bpkroth: This part might have actually been better split to a separate PR.  I'm not sure how necessary it is now, and maybe we leave it, but it could have made some of the other changes simplier to leave them as DataFrames.

mlos_core/mlos_core/tests/optimizers/optimizer_test.py
jsfreischuetz: Do you not like type hints?

mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
jsfreischuetz: This runs faster, but I'll accept the change

mlos_core/mlos_core/optimizers/observations.py
jsfreischuetz: This is required for something I did, I can't remember exactly where, but it is not automatic.

mlos_core/mlos_core/util.py
```diff
@@ -10,9 +10,9 @@
 from ConfigSpace import Configuration, ConfigurationSpace
 
 
-def config_to_dataframe(config: Configuration) -> pd.DataFrame:
+def config_to_series(config: Configuration) -> pd.Series:
```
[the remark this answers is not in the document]
jsfreischuetz: This gets used in the internal representation and in some of the test cases. The main reason to have it out is because of the test cases.

mlos_core/mlos_core/optimizers/observations.py
jsfreischuetz: I tried this, and it is incredibly painful to get it to work

mlos_core/mlos_core/optimizers/optimizer.py
jsfreischuetz: I'm not sure that this makes sense, at least for now. Context is exclusively a series.

jsfreischuetz:
@bpkroth

jsfreischuetz:
If these warnings need to be resolved, then I don't think that the data type members should be protected

README.md
```diff
@@ -67,8 +67,8 @@ To do this this repo provides two Python modules, which can be used independentl
...
-  - provide some helpers for [automating optimization experiment](https://microsoft.github.io/MLOS/overview.html#mlos-bench-api) runner loops and data collection
+  - an "optimizer" service [abstraction](https://microsoft.github.io/MLOS/source_tree_docs/index.html) (e.g. [`register()`](https://microsoft.github.io/MLOS/autoapi/mlos_bench/optimizers/base_optimizer/index.html#mlos_bench.optimizers.base_optimizer.Optimizer.register) and [`suggest()`](https://microsoft.github.io/MLOS/autoapi/mlos_bench/optimizers/base_optimizer/index.html#mlos_bench.optimizers.base_optimizer.Optimizer.suggest)) so we can easily swap out different implementations methods of searching (e.g. random, BO, LLM, etc.)
+  - provide some helpers for [automating optimization experiment](https://microsoft.github.io/MLOS/source_tree_docs/mlos_bench/index.html) runner loops and data collection
```
bpkroth, line 71: Will move this to a new PR

mlos_bench/mlos_bench/__init__.py
```diff
@@ -105,7 +105,7 @@
 Here's the shell command you'd actually run:
 # mlos_bench --config mlos_bench/mlos_bench/tests/config/cli/test-cli-local-env-bench.jsonc --globals experiment_test_local.jsonc --tunable_values tunable-values/tunable-values-local.jsonc
 >>> # Now we run the command and check the output.
->>> result = run(cmd, shell=True, capture_output=True, text=True, check=True)
+>>> result = run(cmd, shell=True, capture_output=True, text=True)
```
bpkroth, line 108: revert

mlos_bench/mlos_bench/tests/optimizers/toy_optimization_loop_test.py
```diff
@@ -95,7 +99,7 @@ def test_mock_optimization_loop_no_defaults(
 def test_flaml_optimization_loop(mock_env_no_noise: MockEnv, flaml_opt: MlosCoreOptimizer) -> None:
     """Toy optimization loop with mock environment and FLAML optimizer."""
     (score, tunables) = _optimize(mock_env_no_noise, flaml_opt)
-    assert score == pytest.approx(60.15, 0.01)
+    assert score == pytest.approx(60, 0.01)
```
bpkroth, line 102: Why'd this change?

mlos_core/mlos_core/__init__.py
```diff
@@ -62,39 +62,39 @@
...
-y    7
-Name: 1, dtype: int64
+>>> assert len(observations._config) == 1
+>>> assert len(observations._score) == 1
+>>> observations._config
+    x
+0  10
+>>> observations._score
+   y
+0  7
```
bpkroth, line 97: 👍

mlos_core/mlos_core/data_classes.py
```diff
@@ -0,0 +1,350 @@
...
+    Attributes
+    ----------
+    config : pd.Series
+        The configuration observed.
+    score : pd.Series
+        The score metrics observed.
+    context : Optional[pd.Series]
+        The context in which the configuration was evaluated.
+        Not Yet Implemented.
+    metadata: Optional[pd.Series]
```
bpkroth, line 31: Might need to change these to `pandas.` instead of `pd.`

README.md
jsfreischuetz: It won't pass tests if I don't do this

mlos_core/mlos_core/data_classes.py
```diff
@@ -0,0 +1,350 @@
...
+    """
+
+    def __init__(
+        self,
+        *,
+        config: pd.DataFrame = pd.DataFrame(),
+        score: pd.DataFrame = pd.DataFrame(),
+        context: Optional[pd.DataFrame] = None,
+        metadata: Optional[pd.DataFrame] = None,
+        observations: List[Observation] = [],
```
bpkroth, line 113:
```suggestion
        observations: Optional[List[Observation]] = None,
```

Don't pass objects - they are mutable (pylint will probably also complain about this.

mlos_core/mlos_core/data_classes.py
```diff
@@ -0,0 +1,350 @@
...
+                metadata = pd.concat(
+                    [obs._metadata.to_frame().T for obs in observations]  # type: ignore
+                )
+            else:
+                metadata = None
+        assert len(config.index) == len(score.index), "config and score must have the same length"
+        if context is not None:
+            l1 = len(config.index)
+            l2 = len(context.index)
+            assert l1 == l2, "config and context must have the same length"
```
bpkroth, line 136: skip variable assignment, include the `len()` check in the assert and use a line break for the message?

mlos_core/mlos_core/data_classes.py
jsfreischuetz: context of "might need to change"? There are no other places where this was done

mlos_core/mlos_core/data_classes.py
```diff
@@ -0,0 +1,350 @@
...
+
+            if sum([obs._context is None for obs in observations]) == 0:
+                context = pd.concat(
+                    [obs._context.to_frame().T for obs in observations]  # type: ignore
+                )
+            else:
+                context = None
+            if sum([obs._metadata is None for obs in observations]) == 0:
+                metadata = pd.concat(
+                    [obs._metadata.to_frame().T for obs in observations]  # type: ignore
```
bpkroth: Add the specific type to the ignore.

`# type: ignore[foo]`

bpkroth: And anywhere else here.

mlos_bench/mlos_bench/tests/optimizers/toy_optimization_loop_test.py
jsfreischuetz: I have reverted this, but there is a problem with this test case. There are some random times where it seems like the return value can change because of a seed change

mlos_core/mlos_core/data_classes.py
```diff
@@ -0,0 +1,350 @@
...
+        context = None if other._context is None else other._context.to_frame().T
+        metadata = None if other._metadata is None else other._metadata.to_frame().T
+        if len(self._config.index) == 0:
+            self._config = config
+            self._score = score
+            self._context = context
+            self._metadata = metadata
+            assert set(self._config.index) == set(
+                self._score.index
+            ), "config and score must have the same index"
```
bpkroth, line 189: complete nit, but can you put the string in parens so it breaks on that instead of the equality expression?

mlos_core/mlos_core/data_classes.py
```diff
@@ -0,0 +1,350 @@
...
+            return
+
+        self._config = pd.concat([self._config, config]).reset_index(drop=True)
+        self._score = pd.concat([self._score, score]).reset_index(drop=True)
+        assert set(self._config.index) == set(
+            self._score.index
+        ), "config and score must have the same index"
```
bpkroth, line 196: same here

mlos_core/mlos_core/data_classes.py
```diff
@@ -0,0 +1,350 @@
...
+                + " if metadata of prior observation is not null"
+            )
+            self._metadata = pd.concat([self._metadata, metadata]).reset_index(drop=True)
+            assert self._config.index.equals(
+                self._metadata.index
+            ), "config and metadata must have the same index"
+        else:
+            assert metadata is None, (
+                "context of appending observation must be null"
+                " if metadata of prior observation is null"
```
bpkroth, line 224: Above you use an explicit concat and here you use an implicit one.
Maybe just pick one style and stick to it?

bpkroth: Also, preference for trailing whitespace on the previous line.

mlos_core/mlos_core/data_classes.py
```diff
@@ -0,0 +1,350 @@
...
+        if not self._score.equals(other._score):
+            return False
+        if not compare_optional_dataframe(self._context, other._context):
+            return False
+        if not compare_optional_dataframe(self._metadata, other._metadata):
+            return False
+
+        return True
+
+    # required as per: https://stackoverflow.com/questions/30643236/does-ne-use-an-overridden-eq
```
bpkroth, line 279: 👍

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -150,7 +153,7 @@ def __init__(
             self.optimizer_parameter_space,
             objectives=self._optimization_targets,
             name=run_name,
-            output_directory=Path(output_directory),
+            output_directory=Path(output_directory) if output_directory else Path("smac3_output"),
```
bpkroth, line 154: This shouldn't happen, right?
(we set it to a temporary path above)

Please revert - seems unrelated.

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -356,18 +354,15 @@ def _suggest(
...
+        suggestion: Suggestion
+            The suggestion to evaluate.
         """
         if TYPE_CHECKING:
             # pylint: disable=import-outside-toplevel,unused-import
             from smac.runhistory import TrialInfo
 
         if context is not None:
-            warn(f"Not Implemented: Ignoring context {list(context.columns)}", UserWarning)
+            warn(f"Not Implemented: Ignoring context {list(context.index)}", UserWarning)
```
bpkroth, line 363: Why is this switching?

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
```diff
@@ -377,31 +372,21 @@ def _suggest(
...
-        configs: pd.DataFrame,
-        context: Optional[pd.DataFrame] = None,
-    ) -> npt.NDArray:
+    def surrogate_predict(self, suggestion: Suggestion) -> npt.NDArray:
         # pylint: disable=import-outside-toplevel
         from smac.utils.configspace import convert_configurations_to_array
 
-        if context is not None:
-            warn(f"Not Implemented: Ignoring context {list(context.columns)}", UserWarning)
+        if suggestion._context is not None:
```
bpkroth: Private accessor?

mlos_core/mlos_core/optimizers/flaml_optimizer.py
```diff
@@ -101,54 +102,47 @@ def __init__(
...
-        *,
-        configs: pd.DataFrame,
-        scores: pd.DataFrame,
-        context: Optional[pd.DataFrame] = None,
-        metadata: Optional[pd.DataFrame] = None,
+        observations: Optional[Union[Observation, Observations]] = None,
     ) -> None:
         """
-        Registers the given configs and scores.
+        Registers the given config and scores.
```
bpkroth, line 108:
```suggestion
        Registers the given config(s) and scores.
```

mlos_core/mlos_core/data_classes.py
jsfreischuetz: black formats it back

mlos_core/mlos_core/optimizers/flaml_optimizer.py
```diff
@@ -101,54 +102,47 @@ def __init__(
...
-
-        for (_, config), (_, score) in zip(configs.astype("O").iterrows(), scores.iterrows()):
-            # Remove None values for inactive config parameters
-            config_dict = drop_nulls(config.to_dict())
-            cs_config: ConfigSpace.Configuration = ConfigSpace.Configuration(
-                self.optimizer_parameter_space,
-                values=config_dict,
+        assert isinstance(
+            observations, Observation
+        ), "Internal implementation does not support Observations."
```
bpkroth, line 117: Why not?
At the very least it could convert an Observations object with multiple Observation instances to a singleton.

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
jsfreischuetz: because series don't have columns

mlos_core/mlos_core/optimizers/flaml_optimizer.py
bpkroth: I'd rather not mix too many changes at once here.

If the previous implementation handled registering multiple scores at once, we should retain that.

If you want to refactor it later to restrict it to only handle registering one at a time or doing that via a loop internally, we should make that change separately.

This one should only be about changing the signatures to take a new encapsulated data type.

mlos_core/mlos_core/optimizers/flaml_optimizer.py
```diff
@@ -161,24 +155,18 @@ def _suggest(
 
         Returns
         -------
-        configuration : pd.DataFrame
+        configuration : pd.Series
```
bpkroth, line 158:
```suggestion
        configuration : Suggestion
```

mlos_core/mlos_core/optimizers/optimizer.py
```diff
@@ -83,207 +85,173 @@ def space_adapter(self) -> Optional[BaseSpaceAdapter]:
...
+        """
+        if isinstance(observations, Observation):
+            self.register_single(observation=observations)
+        if isinstance(observations, Observations):
+            for obs in observations.to_list():
+                self.register_single(observation=obs)
 
-        context : pandas.DataFrame
-            Not Yet Implemented.
+    def register_single(self, observation: Observation) -> None:
```
bpkroth, line 105: Please make this a private method.

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
jsfreischuetz: This is to fix a linter error that slipped through a pervious PR. It won't  pass linting otherwise

mlos_core/mlos_core/data_classes.py
jsfreischuetz: black formats it back

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
bpkroth: Sorry, I was asking why it was private.  I think `suggestion.context` is probably better, no?

mlos_core/mlos_core/data_classes.py
bpkroth: `make doc` nitpicking might complain otherwise.  If it not it's fine, leave it.

README.md
bpkroth: Yeah, I know.  I moved it to a new PR though - it's unrelated.
See #883 for more details.

mlos_core/mlos_core/util.py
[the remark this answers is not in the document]
bpkroth: static methods can be called in test methods too.

It's purely an organizational thing.

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
bpkroth: Add an `assert output_directory` above this instead then.

bpkroth: (sometimes the linter logic changes)

mlos_bench/mlos_bench/optimizers/convert_configspace.py
```diff
@@ -7,7 +7,8 @@
 """
 
 import logging
-from typing import Dict, Hashable, List, Optional, Tuple, Union
+from collections.abc import Hashable
+from typing import Dict, List, Optional, Tuple, Union
```
bpkroth, line 11: Let's split this out to another PR too - also unrelated change in the linter rules.

mlos_bench/mlos_bench/optimizers/mlos_core_optimizer.py
```diff
@@ -221,18 +222,20 @@ def register(
...
             )
         return registered_score
 
     def get_best_observation(
         self,
     ) -> Union[Tuple[Dict[str, float], TunableGroups], Tuple[None, None]]:
-        (df_config, df_score, _df_context) = self._opt.get_best_observations()
-        if len(df_config) == 0:
+        best_observations = self._opt.get_best_observations()
+        if len(best_observations.config) == 0:
```
bpkroth, line 236: Since for `Observations` `len(config) == len(scores)`, let's add `__len__` to `Observations` so this can be `len(best_observations) == 0`

mlos_bench/mlos_bench/tests/dict_templater_test.py
```diff
@@ -85,7 +85,7 @@ def test_os_env_expansion(source_template_dict: Dict[str, Any]) -> None:
 
     results = DictTemplater(source_template_dict).expand_vars(use_os_env=True)
     assert results == {
-        "extra_str-ref": f"{environ['extra_str']}-ref",
+        "extra_str-ref": f"{environ['extra_str']}-ref",  # pylint: disable=inconsistent-quotes
```
bpkroth, line 88: Unrelated

mlos_core/mlos_core/__init__.py
```diff
@@ -62,39 +62,39 @@
...
->>> assert len(configs_df) == 1
->>> assert len(scores_df) == 1
->>> configs_df.iloc[0]
-x    10
-Name: 1, dtype: int64
->>> scores_df.iloc[0]
-y    7
-Name: 1, dtype: int64
+>>> assert len(observations.config) == 1
+>>> assert len(observations.score) == 1
```
bpkroth, line 91: Similar comment about `len(observations)`

mlos_core/mlos_core/optimizers/bayesian_optimizers/smac_optimizer.py
jsfreischuetz: It is now, but I changed it to be private based on a previous comment, and then decided that it doesn't work and have reverted it back

mlos_core/mlos_core/data_classes.py
```diff
@@ -0,0 +1,365 @@
+#
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+#
+"""Data classes for mlos_core used to pass around configurations, observations, and
+suggestions.
+"""
+from __future__ import annotations
```
bpkroth, line 8: Thing I noted earlier that we shouldn't need this if we use `"QuotedClassNameAsTypeHint"`

mlos_core/mlos_core/data_classes.py
```diff
@@ -0,0 +1,365 @@
...
+        The score metrics observed in a dataframe.
+    context : Optional[pd.Series]
+        The context in which the configuration was evaluated.
+        Not Yet Implemented.
+    metadata: Optional[pd.Series]
+        The metadata in which the configuration was evaluated
+        Not Yet Implemented.
+    """
+
+    def __init__(  # pylint: disable=dangerous-default-value
```
bpkroth, line 106:
```suggestion
    def __init__(
```

Fix those

mlos_core/mlos_core/data_classes.py
jsfreischuetz: The linter gets pretty unhappy about this since you are using the constructors ect

mlos_bench/mlos_bench/optimizers/convert_configspace.py
jsfreischuetz: This is not a change in linter rules -- Hashable was depreciated

mlos_core/mlos_core/util.py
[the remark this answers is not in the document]
jsfreischuetz: I think that this function fits with all of the other functions in this file. It is used across multiple classes, and it is used in multiple test cases, but it would not be used outside of mlos_core

mlos_core/mlos_core/tests/spaces/spaces_test.py
bpkroth, line 1: revert

mlos_core/mlos_core/tests/spaces/monkey_patch_quantization_test.py
bpkroth, line 1: revert

mlos_core/mlos_core/optimizers/optimizer.py
jsfreischuetz: It's a little weird since context becomes a dataframe in Observations. Also if we did this, we should also do it for metadata, and probably also config, both of which also have this problem that in Observations they go from series to dataframes.

jsfreischuetz: I have a minimal implementation in the PR now, but really not sure what to do with observations, other than take a list of series, which feels pretty awkward

jsfreischuetz: @bpkroth

bpkroth: K, fine, but then let's leave the '*' param so that it's always called as

`opt.suggest(context=my_context_series)`

mlos_core/mlos_core/data_classes.py
bpkroth: Fixed

mlos_bench/mlos_bench/tests/optimizers/toy_optimization_loop_test.py
bpkroth: Please file a bug and submit a separate PR on that.  Let's keep unrelated changes in separate PRs.

bpkroth (approved):

bpkroth:
All good now.  Thanks for your patience :)

mlos_core/mlos_core/util.py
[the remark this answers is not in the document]
bpkroth: Let's just leave it as is for now for now so we can get this in.  Thanks!

bpkroth:
Found a few more issues.  Gonna try and tackle them myself.

bpkroth (approved):

PR #720, opened 2024-03-19
Introduce TrialRunner Abstraction

bpkroth:
This is another step in adding support for parallel trial execution #380.

Here we separate out the running of an individual trial to a single class - TrialRunner.

Multiple TrialRunners are instantiated at CLI invocation with the `--num-trial-runners` argument. 
Each TrialRunner associated with a single copy of the root Environment and its Services (important for eventual parallelization reasons), and made unique by means of a unique `trial_runner_id` value that's also included in that Environment's global_config.

In future PRs we will add:
- New Scheduler implementations to run TrialRunners in parallel.
- Async polling of status results in each TrialRunner independently.

PR #853, opened 2024-08-23
Move to azure VM managed run commands

eujing:
A comparison of the existing action run command API vs new managed version: https://learn.microsoft.com/en-us/azure/virtual-machines/run-command-overview

The main benefits of moving to the new managed version are:
- Much longer remote execution timeouts can be set beyond the current 90 mins. (Supposedly supports up to days)
- Returns the exit code of script ran.
- Can run multiple managed run commands in parallel.
- Supports storing large script output into blob storage (not used in this PR)

The first point is really needed for steps like the benchbase loading phase, where loading TPCC scale factor 200+ is a struggle due to hitting the previous 90 min timeout limit.

This should be possible by setting `"pollTimeout": 5400` to much higher values as needed in the relevant VM service configs.


mlos_bench/mlos_bench/environments/remote/remote_env.py
```diff
@@ -175,10 +176,14 @@ def _remote_exec(self, script: Iterable[str]) -> Tuple[Status, datetime, Optiona
             Status is one of {PENDING, SUCCEEDED, FAILED, TIMED_OUT}
         """
         env_params = self._get_env_params()
+        command_name = re.sub(r"\W+", "-", self.name).lower()
```
motus, line 179: I am not sure that using the `Environment` name as command ID will work. The environment can have separate commands for setup, run, and teardown, at least. In addition to that, other environments can inherit from `RemoteEnv` and make additional calls to `_remote_exec()` for their internal needs (e.g., early termination).

So maybe pass the command name as an additional parameter to `_remote_exec()`?

**P.S.** We also have to think about the parallelization of the experiments, when (theoretically) several trials may simultaneously run different commands on the same VM

motus, line 179: Also, precompile your regex:
```suggestion
    _RE_SPECIAL = re.compile(r"\W+")  # class-level const
        command_name = self._RE_SPECIAL.sub("-", self.name).lower()
```

motus (commented):
no objections in general, but I have some concerns about the way we assign IDs to the commands. I am afraid they are insufficiently unique. Also, how to deal with versioning, unfinished commands, and such?

motus:
Hmm, this issue is new - 
```

LibMambaUnsatisfiableError: Encountered problems while solving:
  - nothing provides _python_rc needed by python-3.12.0rc3-rc3_hab00c5b_1_cpython

Could not solve for environment specs
The following packages are incompatible
├─ build is installable with the potential options
│  ├─ build 0.10.0 would require
│  │  └─ python >=3.10,<3.11.0a0 , which can be installed;
│  ├─ build 0.10.0 would require
│  │  └─ python >=3.11,<3.12.0a0 , which can be installed;
│  ├─ build 0.10.0 would require
│  │  └─ python >=3.8,<3.9.0a0 , which can be installed;
│  └─ build 0.10.0 would require
│     └─ python >=3.9,<3.10.0a0 , which can be installed;
└─ python 3.12**  is not installable because there are no viable options
   ├─ python [3.12.0|3.12.1|...|3.12.5] conflicts with any installable versions previously reported;
   └─ python 3.12.0rc3 would require
      └─ _python_rc, which does not exist (perhaps a missing channel).
```
I can swear we used to have Python 3.12.0 in Linux, and now it somehow is back to rc3
@bpkroth do you know what's going on? must be a conda issue as I see it on my MacOS, too

mlos_bench/mlos_bench/environments/remote/remote_env.py
eujing: Agreed. For the previous behavior, setup, run and teardown would just keep overriding the managed run command.
Added the command name as an additional parameter, and now each step adds a related suffix to the name.

eujing: Updated to use precompiled regex

bpkroth: > The first point is really needed for steps like the benchbase loading phase, where loading TPCC scale factor 200+ is a struggle due to hitting the previous 90 min timeout limit.

Another option here is make these operations all two phase.  e.g., they could all have a "start" and "finish" sub-phase where "start_setup" is used to kick off the loading process in the background and "finish_setup" is used to wait for its result and report that back to MLOS.

That would allow the script writer to create async operations and have polling in the background and it would work across multiple backend services (e.g., SshService).

bpkroth:
> no objections in general, but I have some concerns about the way we assign IDs to the commands. I am afraid they are insufficiently unique. Also, how to deal with versioning, unfinished commands, and such?

Agreed.

> The first point is really needed for steps like the benchbase loading phase, where loading TPCC scale factor 200+ is a struggle due to hitting the previous 90 min timeout limit.

Another option here is make these operations all two phase.  e.g., they could all have a "start" and "finish" sub-phase where "start_setup" is used to kick off the loading process in the background and "finish_setup" is used to wait for its result and report that back to MLOS.

That would allow the script writer to create async operations and have polling in the background and it would work across multiple backend services (e.g., SshService).

bpkroth:
> Hmm, this issue is new -
> 
> ```
> 
> LibMambaUnsatisfiableError: Encountered problems while solving:
>   - nothing provides _python_rc needed by python-3.12.0rc3-rc3_hab00c5b_1_cpython
> 
> Could not solve for environment specs
> The following packages are incompatible
> ├─ build is installable with the potential options
> │  ├─ build 0.10.0 would require
> │  │  └─ python >=3.10,<3.11.0a0 , which can be installed;
> │  ├─ build 0.10.0 would require
> │  │  └─ python >=3.11,<3.12.0a0 , which can be installed;
> │  ├─ build 0.10.0 would require
> │  │  └─ python >=3.8,<3.9.0a0 , which can be installed;
> │  └─ build 0.10.0 would require
> │     └─ python >=3.9,<3.10.0a0 , which can be installed;
> └─ python 3.12**  is not installable because there are no viable options
>    ├─ python [3.12.0|3.12.1|...|3.12.5] conflicts with any installable versions previously reported;
>    └─ python 3.12.0rc3 would require
>       └─ _python_rc, which does not exist (perhaps a missing channel).
> ```
> 
> I can swear we used to have Python 3.12.0 in Linux, and now it somehow is back to rc3 @bpkroth do you know what's going on? must be a conda issue as I see it on my MacOS, too

Fixed this in #856

mlos_bench/mlos_bench/environments/remote/remote_env.py
bpkroth: > P.S. We also have to think about the parallelization of the experiments, when (theoretically) several trials may simultaneously run different commands on the same VM

I'm not certain about that part.  There should only ever be one trial executing on an Environment (VM) at a time.

We *will* want to have different Environment copies for parallel Trial execution across Workers though (#720).

bpkroth: > Agreed. For the previous behavior, setup, run and teardown would just keep overriding the managed run command. Added the command name as an additional parameter, and now each step adds a related suffix to the name.

Here's my proposal:

1. Assemble a json (dict) of 

```python
command_id_info = {
  "environment_id": self.id,
  "command": command_str,
  "config": config,
  "globals": globals,
}
```

  Basically it should include anything that goes into making that execution unique.
  In particular, the environment id should include something about the name of the environment and its type, the config should include the particulars for this invocation and maybe also the trial_id, the globals may incorporate any other substitutions, the command is obviously the particular command being executed, etc.

2. Create a digest of the info: `sha256(json.dumps(command_info)).hex` or somesuch

bpkroth: The question about outstanding unfinished commands is import though.

Can we check for those prior to issuing a new exec?

eujing: > 2\. Create a digest of the info: `sha256(json.dumps(command_info)).hex` or somesuch

This sounds like a good idea, although I need to confirm this fits within the length and character limits of command names.

There is currently a big blocker for including trial-specific info like the `trial_id` and `parameters` into the command naming. Currently, there is a limit of 25 managed run commands per instance: https://learn.microsoft.com/en-us/azure/virtual-machines/run-command-overview#when-to-use-action-or-managed-commands
This would severely limit how many trials we can run, even if assuming each trial only makes one remote exec call.


> The question about outstanding unfinished commands is import though.
> 
> Can we check for those prior to issuing a new exec?

Yes this can be done, either by issuing a GET on the endpoints `runcommands` or `runcommands/{command_name}`, where we can see the execution status. I think it is a good idea to check unfinished commands before issuing a new one, but maybe better done in a separate PR as we don't support that even currently.

bpkroth: > This sounds like a good idea, although I need to confirm this fits within the length and character limits of command names.

sha256 produces a 64 character hex hash.  Just truncate it to fit whatever the limit is.  git uses sha1 and often only reports the first 7 characters.  That still gives (26+10)^7 ~ 78 billion possibilities.  It doesn't need to be perfect.  Probabilistic is still very good for this case since commands will eventually be retired (completed and removed from the history on the backend) over time.

bpkroth: > There is currently a big blocker for including trial-specific info like the trial_id and parameters into the command naming. Currently, there is a limit of 25 managed run commands per instance: https://learn.microsoft.com/en-us/azure/virtual-machines/run-command-overview#when-to-use-action-or-managed-commands
This would severely limit how many trials we can run, even if assuming each trial only makes one remote exec call.

I see the issue now.  Managed run scripts are a resource you create and associate with a VM and then invoke.

So, to work around the limits, the workflow would have to be something like:

1. Create managed script resource.
2. Invoke it.
3. Wait for the result.
4. Delete the managed script resource.

That adds some additional back and forth with ARM which can be slow unfortunately.

Question:

Can we we create somewhat more generic managed scripts objects that can be invoked with arguments and/or environment variables (or maybe fed via a file) in order to adjust their behavior.
*If* there are less than 25 of them, then we can create the objects once and execute as necessary.

bpkroth: Actually, given the output blob URIs are also fixed, we might have to do this create/invoke/wait/delete operation every time anyways.

motus (approved):
I think we can merge it. This fix is a clear improvement over the existing code, and all additional functionality that we discuss in the comments can be implemented later when we actually need it. @bpkroth , WDYT?

mlos_bench/mlos_bench/services/remote/azure/azure_vm_services.py
```diff
@@ -231,6 +242,28 @@ def wait_host_operation(self, params: dict) -> Tuple[Status, dict]:
         params.setdefault(f"{params['vmName']}-deployment")
         return self._wait_while(self._check_operation_status, Status.RUNNING, params)
 
+    def wait_remote_exec_operation(self, params: dict) -> Tuple["Status", dict]:
```
bpkroth, line 245: Wondering if we need to add this to the `SupportsRemoteExec` type as well.

eujing:
> Generally, also OK to proceed, though I haven't tested it myself, so kinda on you to support @eujing.
> 
> Maybe to address your concern about the uniqueness of the command names, simply prepend "mlos_{env_name}_{phase}" for now?
> 
> Thanks!

I have been using this version in my most recent experiments, so it has been working on my side!

For command names, it is currently similar to what you have proposed, roughly "{env_name}-{phase}".

mlos_bench/mlos_bench/environments/remote/remote_env.py
```diff
@@ -72,6 +75,7 @@ def __init__(  # pylint: disable=too-many-arguments
         )
 
         self._wait_boot = self.config.get("wait_boot", False)
+        self._command_prefix = self._RE_SPECIAL.sub("-", self.name).lower() + "-"
```
bpkroth, line 78:
```suggestion
        self._command_prefix = "mlos-" + self._RE_SPECIAL.sub("-", self.name).lower() + "-"
```

Just to help namespace them a tad more.

bpkroth (approved):
LGTM

mlos_bench/mlos_bench/services/remote/azure/azure_vm_services.py
bpkroth: Let's skip this for now.

PR #855, opened 2024-09-04
CyclicConfigSuggestor

eujing:
Currently used for validating a default configuration against some other specific configuration, in an alternating manner on the same MySQL instance. This sets up the collected pairs of metrics for something like a matched-pair t-test, without the noise of making comparisons across different MySQL instances.

More generally, multiple configs can be specified via the list `cycle_tunable_values` to cycle through in optimization mode.

EDIT: Originally implemented as a scheduler, now is a "ManualOptimizer" instead.


motus:
I am not sure that this functionality belongs to the Scheduler. The Scheduler determines *when* the trial should be executed, not the *configuration* of this trial. The latter belongs to the Optimizer, and we already have the one-shot optimizer (effectively, a benchmark) and a grid search optimizer. I think your cyclic scheduler belongs there - and it will be much shorter and simpler as the Optimizer instance.

P.S. Also, maybe we should use a better word for the Optimizers, since not all of them actually optimize stuff. A ConfigurationProposer, maybe? ConfigExplorer? @bpkroth do you have any suggestions?

bpkroth:
> I am not sure that this functionality belongs to the Scheduler. The Scheduler determines _when_ the trial should be executed, but not the _configuration_ of this trial. The latter belongs to the Optimizer, and we already have one-shot optimizer (effectively, a benchmark) and grid search optimizer that both, just like your "cyclic scheduler", ignore the trial results and just go on with a pre-determined sequence of configurations. I think your code should actually be an Optimizer - and its implementation will be much shorter and simpler that way.

I actually thing it's a fine fit for the Scheduler.

This isn't determining *which* config might improve confidence or scores, that's still handled by the Optimizer.

Instead, it's adjusting the method of trial execution in order to try and come up with an adjusted score to feed back to the Optimizer when evaluating the suggestion, one that is hopefully easier for the Optimizer to learn from.

Our other efforts on noisy tuning do something similar.

> P.S. Also, maybe we should use a better word for the Optimizers, since not all of them actually optimize stuff. A ConfigurationProposer, maybe? ConfigExplorer? @bpkroth do you have any suggestions?

That's true I suppose, but I'm not certain I want to tackle that change atm.  Let's take it up separately.

mlos_bench/mlos_bench/config/schemas/schedulers/scheduler-schema.json
```diff
@@ -62,7 +62,9 @@
             "$comment": "required",
             "enum": [
                 "mlos_bench.schedulers.SyncScheduler",
-                "mlos_bench.schedulers.sync_scheduler.SyncScheduler"
+                "mlos_bench.schedulers.sync_scheduler.SyncScheduler",
+                "mlos_bench.schedulers.CyclicScheduler",
+                "mlos_bench.schedulers.cyclic_scheduler.CyclicScheduler"
```
bpkroth, line 67: nit: alphabetize these (the diff will be smaller too)

mlos_bench/mlos_bench/config/schemas/schedulers/scheduler-schema.json
```diff
@@ -83,7 +85,9 @@
                     "class": {
                         "enum": [
                             "mlos_bench.schedulers.SyncScheduler",
-                            "mlos_bench.schedulers.sync_scheduler.SyncScheduler"
+                            "mlos_bench.schedulers.sync_scheduler.SyncScheduler",
+                            "mlos_bench.schedulers.CyclicScheduler",
+                            "mlos_bench.schedulers.cyclic_scheduler.CyclicScheduler"
```
bpkroth, line 90: same here

mlos_bench/mlos_bench/schedulers/__init__.py
```diff
@@ -6,8 +6,10 @@
 
 from mlos_bench.schedulers.base_scheduler import Scheduler
 from mlos_bench.schedulers.sync_scheduler import SyncScheduler
+from mlos_bench.schedulers.cyclic_scheduler import CyclicScheduler
```
bpkroth, line 9: Hmm, surprised isort didn't complain about this

mlos_bench/mlos_bench/schedulers/__init__.py
```diff
@@ -6,8 +6,10 @@
...
 
 __all__ = [
     "Scheduler",
     "SyncScheduler",
+    "CyclicScheduler",
```
bpkroth, line 14: same

mlos_bench/mlos_bench/schedulers/cyclic_scheduler.py
```diff
@@ -0,0 +1,80 @@
...
+from mlos_bench.schedulers.sync_scheduler import SyncScheduler
+from mlos_bench.storage.base_storage import Storage
+from mlos_bench.util import merge_parameters
+
+_LOG = logging.getLogger(__name__)
+
+
+class CyclicScheduler(SyncScheduler):
+    """
+    A simple single-threaded synchronous benchmarking loop implementation
```
bpkroth, line 19:
```suggestion
    A simple synchronous benchmarking loop implementation
```

mlos_bench/mlos_bench/schedulers/cyclic_scheduler.py
```diff
@@ -0,0 +1,80 @@
...
+    that cycles through a list of provided configuration IDs.
+
+    This scheduler does not use any optimizer and is mainly used for benchmarking
+    specific configurations against each other only.
```
bpkroth, line 23: Let's remove that.

This should be composable with Optimizers.

If you want to only evaluate a small set of tunable-param configs, we have (or can make) other Optimizers (e.g., OneShot) that we can make use of.

mlos_bench/mlos_bench/schedulers/cyclic_scheduler.py
```diff
@@ -0,0 +1,80 @@
...
+
+    `cycle_config_ids` is the  list of existing configuration IDs to cycle through.
+    It can be provided via the scheduler's config, or overwritten through the global config.
```
bpkroth, line 26: As discussed above - please remove this.  That functionality should be constrained to an Optimizer (or ConfigSuggestor) as @motus mentioned.

mlos_bench/mlos_bench/schedulers/cyclic_scheduler.py
```diff
@@ -0,0 +1,80 @@
...
+        *,
+        config: Dict[str, Any],
+        global_config: Dict[str, Any],
+        environment: Environment,
+        optimizer: Optimizer,
+        storage: Storage,
+        root_env_config: str,
+    ):
+        """
+        Create a new instance of the scheduler. The constructor of this and the derived
```
bpkroth, line 40: Do we need this comment?  Since it's in the parent class, maybe remove it to avoid extra maintenance issues, unless there's something specific to this one.

mlos_bench/mlos_bench/schedulers/cyclic_scheduler.py
bpkroth, line 1: Needs tests.

mlos_bench/mlos_bench/schedulers/cyclic_scheduler.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+        self._cycle_config_ids: List[int] = config.get("cycle_config_ids", [0])
+
+    def _schedule_new_optimizer_suggestions(self) -> bool:
+        not_done = self.not_done()
+        if not_done:
+            config_id = self._cycle_config_ids[self._trial_count % len(self._cycle_config_ids)]
+            tunables = self.load_config(config_id)
+            self.schedule_trial(tunables)
+        return not_done
```
bpkroth, line 80: I thought the goal was to test the "default" vs. the "suggested config" and come up with a "normalized" score for this backend instance?

mlos_bench/mlos_bench/schedulers/cyclic_scheduler.py
bpkroth: Eventually, it'd be good to have a parallel version of this that can evaluate the default vs. new_suggestion across multiple nodes, so please also rename this to CyclicSyncScheduler.

mlos_bench/mlos_bench/schedulers/cyclic_scheduler.py
eujing: I think that is a good idea we discussed for optimization, but I intended this to just support validation of a suggested config vs the default, in benchmarking mode only. With the pairs (or generally tuples) of measurements, I can then do something like a matched t-test for significance testing.

I wanted a way to repeatedly measure: default vs some suggested config, on the same server instance.
Previously, I would be comparing across separate benchmarking runs that end up using different server instances, which have varying baseline performances.

bpkroth: Ah, I see.  Then I agree with Sergiy.  You want a very specific version of the "ConfigSuggestor".

Let's call it "ManualOptimizer" then.  It can extend OneShotOptimizer or MockOptimizer and simply run a provided array of `tunable-values` and no more.

motus:
@bpkroth says:
> This isn't determining which config might improve confidence or scores, that's still handled by the Optimizer.
> 
> Instead, it's adjusting the method of trial execution in order to try and come up with an adjusted score to feed back to the Optimizer when evaluating the suggestion, one that is hopefully easier for the Optimizer to learn from.

I am afraid I can't agree. I see the Environment, Optimizer, and Scheduler as three orthogonal components:
* `Optimizer`: suggests configurations and puts them on the queue
* `Scheduler`: schedules execution of each configuration extracted from the queue
* `Environment`: executes trials, one config at a time

In that design, any instance of the Environment can be used with any instance of Optimizer and Scheduler. `CyclicScheduler` breaks this: it does not make sense to combine it with any Optimizer (Grid Search, for example); worse yet, since it replaces the Scheduler, it cannot be combined with other schedulers (e.g., for parallel execution).

Note that we don't need to feed the scores to the (Bayesian) optimizer immediately after we receive them from the environment - it will pick that data from Storage after we restart the experiment, e.g., switching from CyclicOptimizer to MlosCoreOptimizer.

Finally, nothing prevents us from implementing a `CompositeOptimizer` just like we did for the Environment. To the rest of the framework, it will look like an Optimizer that generates a sequence of the configurations; internally, it can be, e.g., a sequence of grid search, followed by a certain number of FLAML iterations, followed by a few iterations on baseline and best configs. All that can be combined with ParallelScheduler, once we implement it.

bpkroth:
> @bpkroth says:
> 
> > This isn't determining which config might improve confidence or scores, that's still handled by the Optimizer.
> > Instead, it's adjusting the method of trial execution in order to try and come up with an adjusted score to feed back to the Optimizer when evaluating the suggestion, one that is hopefully easier for the Optimizer to learn from.
> 
> I am afraid I can't agree. I see the Environment, Optimizer, and Scheduler as three orthogonal components:
> 
> * `Optimizer`: suggests configurations and puts them on the queue
> * `Scheduler`: schedules execution of each configuration extracted from the queue
> * `Environment`: executes trials, one config at a time
> 
> In that design, any instance of the Environment can be used with any instance of Optimizer and Scheduler. `CyclicScheduler` breaks this: it does not make sense to combine it with any Optimizer (Grid Search, for example); worse yet, since it replaces the Scheduler, it cannot be combined with other schedulers (e.g., for parallel execution).
> 
> Note that we don't need to feed the scores to the (Bayesian) optimizer immediately after we receive them from the environment - it will pick that data from Storage after we restart the experiment, e.g., switching from CyclicOptimizer to MlosCoreOptimizer.
> 
> Finally, nothing prevents us from implementing a `CompositeOptimizer` just like we did for the Environment. To the rest of the framework, it will look like an Optimizer that generates a sequence of the configurations; internally, it can be, e.g., a sequence of grid search for warm-up, followed by a certain number of FLAML iterations, followed by a few iterations on baseline and best configs. All that can be combined with ParallelScheduler, once we implement it.

I think there was some additional discussion lost here.  I agree with everything you've said above and corrected myself in a different thread.

For summary here, I was mistaken in my original understanding of the intent of this PR.
From previous discussions, I thought @eujing was attempting to create a method for "denoising" the signal produced by running a Trial via "normalizing" it against a run of the default configs on that same instance.

That (which isn't what he actually implemented here) very well could be implemented by a Scheduler (e.g., whenever a config (produced by an Optimizer or manually or whatever) is pulled off the queue, we execute both it and the default and compute the normalized score prior to submitting results back to the Storage layer).

But I agree that you're right about what is actually implemented - it's much more of a "config suggestor", and should be handled in an Optimizer layer instead, given the current abstractions.

motus:
Oh, I see now. We're on the same page, then!

eujing:
@motus @bpkroth I have re-implemented this as a `ManualOptimizer` instead, where an array of explicit tunable values can be provided, either via the optimizer's config or overwritten via globals.

For example:

```json
"cycle_tunable_values": [
            {
                "require_secure_transport": "OFF",
                "sql_generate_invisible_primary_key": "OFF",
                "innodb_change_buffer_max_size": 25,
                "innodb_change_buffering": "all"
            },
            {
                "require_secure_transport": "OFF",
                "sql_generate_invisible_primary_key": "OFF",
                "innodb_change_buffer_max_size": 50,
                "innodb_change_buffering": "inserts"
            }
        ]
```

mlos_bench/mlos_bench/schedulers/cyclic_scheduler.py
eujing: Re-implemented this `ManualOptimizer` now!

mlos_bench/mlos_bench/config/optimizers/manual_opt.jsonc
```diff
@@ -0,0 +1,11 @@
+// Mock optimizer to test the benchmarking framework.
+{
+    "$schema": "https://raw.githubusercontent.com/microsoft/MLOS/main/mlos_bench/mlos_bench/config/schemas/optimizers/optimizer-schema.json",
+
+    "class": "mlos_bench.optimizers.ManualOptimizer",
+
+    "config": {
+        "max_suggestions": 30,
```
bpkroth, line 8:
```suggestion
        "max_cycles": 30,
```

(and the internally limit by `max_suggestions` since I think that's part of the base class config)

mlos_bench/mlos_bench/config/schemas/optimizers/optimizer-schema.json
```diff
@@ -201,6 +203,46 @@
...
+                                    }
+                                }
+                            }
+                        ],
+                        "$comment": "disallow other properties",
+                        "unevaluatedProperties": false
+                    }
+                }
+            },
+            "else": false
```
bpkroth, line 245: Future work, but we should probably split this schema up into smaller files like we do for the other schemas.

mlos_bench/mlos_bench/config/optimizers/manual_opt.jsonc
bpkroth: My thought is it makes intuitive sense to me to talk about testing the entire set of configs N times, rather than stop part way through.  Else it would ruin your comparisons later on.

mlos_bench/mlos_bench/config/schemas/optimizers/optimizer-schema.json
```diff
@@ -201,6 +203,46 @@
...
+            "then": {
+                "properties": {
+                    "config": {
+                        "type": "object",
+                        "allOf": [
+                            { "$ref": "#/$defs/config_base_optimizer" },
+                            {
+                                "type": "object",
+                                "properties": {
+                                    "cycle_tunable_values": {
```
bpkroth, line 230: Add `max_cycles` here as another way to limit `max_suggestions`

mlos_bench/mlos_bench/optimizers/manual_optimizer.py
```diff
@@ -0,0 +1,46 @@
...
+    """Optimizer that proposes an explicit sequence of tunable values."""
+
+    def __init__(
+        self,
+        tunables: TunableGroups,
+        config: dict,
+        global_config: Optional[dict] = None,
+        service: Optional[Service] = None,
+    ):
+        super().__init__(tunables, config, global_config, service)
```
bpkroth, line 28: Add max_cycles vs max_suggestions computation here.

mlos_bench/mlos_bench/tests/config/schemas/optimizers/test-cases/good/partial/manual_opt_schema_partial_config.jsonc
```diff
@@ -0,0 +1,7 @@
+{
+
+    "class": "mlos_bench.optimizers.ManualOptimizer",
+    "config": {
+        "cycle_tunable_values": []
```
bpkroth, line 5: This should be rejected, right?  There's nothing to do.

mlos_bench/mlos_bench/tests/config/schemas/optimizers/test-cases/good/partial/manual_opt_no_schema.jsonc
```diff
@@ -0,0 +1,4 @@
+{
+
```
bpkroth, line 2:
```suggestion
```

nit

mlos_bench/mlos_bench/tests/optimizers/conftest.py
```diff
@@ -154,3 +157,16 @@ def smac_opt_max(tunable_groups: TunableGroups) -> MlosCoreOptimizer:
...
+
+
+@pytest.fixture
+def manual_opt(tunable_groups: TunableGroups, mock_configs: List[dict]) -> ManualOptimizer:
+    """Test fixture for ManualOptimizer."""
+    return ManualOptimizer(
+        tunables=tunable_groups,
+        service=None,
+        config={
+            "max_suggestions": 2 * len(mock_configs),
```
bpkroth, line 169: Exactly the sort of config annoyance I'm thinking to avoid.

Set `"max_cycles": 2` and let `max_suggestions = min(max_suggestions, max_cycles * len(cycle_tunable_values))` internally.

mlos_bench/mlos_bench/tests/optimizers/manual_opt_test.py
```diff
@@ -0,0 +1,19 @@
...
+from mlos_bench.environments.status import Status
+from mlos_bench.optimizers.manual_optimizer import ManualOptimizer
+
+# pylint: disable=redefined-outer-name
+
+
+def test_manual_optimizer(manual_opt: ManualOptimizer, mock_configs: list) -> None:
+    """Make sure that manual optimizer produces consistent suggestions."""
+
+    for i in range(2 * len(mock_configs)):
```
bpkroth, line 16: Would be probably easier to maintain if you kept that dynamic (i.e., didn't expect coordination between the definition in the other file and this one).

bpkroth (changes requested):
Mostly feedback about `max_cycles`, but otherwise looks fine.

mlos_bench/mlos_bench/config/optimizers/manual_opt.jsonc
eujing: `max_cycles` is a good idea, will implement it!

mlos_bench/mlos_bench/tests/config/schemas/optimizers/test-cases/good/partial/manual_opt_schema_partial_config.jsonc
eujing: With the current logic, if an empty array is provided we first emit a warning for that, and then default to an array with just the default tunable values. Kind of falling back to `OneShotOptimizer` behavior.

I was thinking this could be a sane default, to be overwritten via the globals file with specific values compatible with the current tunable groups. Otherwise, someone copying this default optimizer config would fail immediately.

But if you guys think the latter is better, I can change the behavior to that!

mlos_bench/mlos_bench/optimizers/manual_optimizer.py
eujing: Added the max_cycles computation!

mlos_bench/mlos_bench/config/optimizers/manual_opt.jsonc
```diff
@@ -0,0 +1,11 @@
+// Mock optimizer to test the benchmarking framework.
```
bpkroth, line 1:
```suggestion
// Manual optimizer to run fixed set of tunable values on repeat via the benchmarking framework.
```

mlos_bench/mlos_bench/config/optimizers/manual_opt.jsonc
```diff
@@ -0,0 +1,11 @@
...
+{
+    "$schema": "https://raw.githubusercontent.com/microsoft/MLOS/main/mlos_bench/mlos_bench/config/schemas/optimizers/optimizer-schema.json",
+
+    "class": "mlos_bench.optimizers.ManualOptimizer",
+
+    "config": {
+        "max_cycles": 30,
+        "cycle_tunable_values": []
```
bpkroth, line 9:
```suggestion
        "tunable_values_cycle": [
            // Add one or more set of tunable values here as a dictionary:
            // {"tunable_param_1": "tunable_value_1", ... },
        ]
```

mlos_bench/mlos_bench/config/schemas/optimizers/optimizer-schema.json
```diff
@@ -201,6 +203,51 @@
...
+                            { "$ref": "#/$defs/config_base_optimizer" },
+                            {
+                                "type": "object",
+                                "properties": {
+                                    "max_cycles": {
+                                        "description": "The maximum number of cycles of tunable values to run the optimizer for.",
+                                        "type": "integer",
+                                        "minimum": 1
+                                    },
+                                    "cycle_tunable_values": {
```
bpkroth, line 235:
```suggestion
                                    "tunable_values_cycle": {
```

bpkroth: naming nit

mlos_bench/mlos_bench/config/schemas/optimizers/optimizer-schema.json
bpkroth: https://github.com/microsoft/MLOS/issues/867

mlos_bench/mlos_bench/optimizers/manual_optimizer.py
```diff
@@ -0,0 +1,51 @@
...
+        self._cycle_tunable_values: List[Dict[str, TunableValue]] = config.get(
+            "cycle_tunable_values", []
```
bpkroth, line 30:
```suggestion
            "tunable_values_cycle", []
```

mlos_bench/mlos_bench/optimizers/manual_optimizer.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def __init__(
+        self,
+        tunables: TunableGroups,
+        config: dict,
+        global_config: Optional[dict] = None,
+        service: Optional[Service] = None,
+    ):
+        super().__init__(tunables, config, global_config, service)
+        self._cycle_tunable_values: List[Dict[str, TunableValue]] = config.get(
```
bpkroth, line 29:
```suggestion
        self._tunable_values_cycle: List[Dict[str, TunableValue]] = config.get(
```

plrease rename elsewhere as well

mlos_bench/mlos_bench/tests/config/schemas/optimizers/test-cases/good/partial/manual_opt_schema_partial_config.jsonc
bpkroth: I see.  Yeah, that's ok, but let's document it with some comments in the example configs at least.  Thanks!

bpkroth: I made an initial attempt at that below, but actually on thinking about it more, I think a better approach is probably something like the following to make it more explicit:

```jsonc
"cycle_tunable_values": [],  // --> error
"cycle_tunable_values": null, // --> error

"cycle_tunable_values": [
  // explicit placeholder for the defaults, so they don't need to be copied:
  // this can be included in the example config and documented as such
  null,
  // alternatively, (if this is easy) - placeholder for defaults (but I think the schema might reject it currently):
  {},
  // explicit values:
  {"tunable_param_1": "tunable_value_1", ... },
],

bpkroth: The more I think about it, I actually think `{}` as a placeholder for `default` values is the easiest way to do this rather than `null`.

Let me make a new PR to allow supporting that quickly.

bpkroth: @motus does that make sense to you too?

The other alternative I could imagine is `null` as an allowed element in the `tunable_values_cycle` and internally interpret that as `restore_defaults()`.

motus:
In all, I am fine with the implementation, but I really don't like the name `ManualOptimizer`. Can we call it `FewShotOptimizer` or `ConstOptimizer` instead? @bpkroth , WDYT?

Also, can we remove a corresponding TODO comment from `OneShotOptimizer`?

bpkroth:
> In all, I am fine with the implementation, but I really don't like the name `ManualOptimizer`. Can we call it `FewShotOptimizer` or `ConstOptimizer` instead? @bpkroth , WDYT?

Works for me.

Or, can use your other terminology for now: `CyclicConfigSuggestor`
And consider to refactor/rename the rest in the package later on.

> Also, can we remove a corresponding TODO comment from `OneShotOptimizer`?

👍

mlos_bench/mlos_bench/tests/optimizers/manual_opt_test.py
eujing: Updated this to rely on `while manual_opt.not_converged():`, so no more coordination in the lengths!

mlos_bench/mlos_bench/tests/config/schemas/optimizers/test-cases/good/partial/manual_opt_schema_partial_config.jsonc
eujing: Updated to use the new features in #868!

motus (approved):
I think we can merge in the PR already. I am not quite happy with the `ManualOptimizer` name (it does not tell exactly what the class does), but I am not ready to rename all our optimizers to Suggestors yet. Let's merge it as ManualOptimizer and then have another PR to rename it to something more sensible, like `RepeatTrialsOptimizer` 

P.S. Also, can we run a formatter on all JSON files just like we do it for Python? It looks like formatting for some of the new JSON files is a bit off

bpkroth:
> I think we can merge in the PR already. I am not quite happy with the `ManualOptimizer` name (it does not tell exactly what the class does), but I am not ready to rename all our optimizers to Suggestors yet. Let's merge it as ManualOptimizer and then have another PR to rename it to something more sensible, like `RepeatTrialsOptimizer`

Thing is we can already repeat trials via the `Scheduler`.

I thought we'd settled on `FewShotOptimizer` for now and to remove the `TODO` comment in the `OneShotOptimizer`.

Anyways, I agree, we may want to take up renaming later on, but not right now - it'd be a large change.

> P.S. Also, can we run a formatter on all JSON files just like we do it for Python? It looks like formatting for some of the new JSON files is a bit off

Can you file an Issue for this?  We can take it up separately (e.g., via prettier or some such like we used to do).

bpkroth (approved):
See prior comments about naming, but good enough for me for now.

PR #861, opened 2024-09-27
Use scheduler in dummy runs

bpkroth:
Refactors tests in preparation for testing for #720.

Adds `status()` output for MockEnv as well, though with a pending FIXME on improving the Status type in the status output that we'll address in a future PR.


mlos_bench/mlos_bench/tests/storage/sql/fixtures.py
```diff
@@ -107,120 +105,99 @@ def mixed_numerics_exp_storage(
...
+        environment=env,
+        optimizer=opt,
+        storage=storage,
+        root_env_config=exp.root_env_config,
+    )
 
+    # Add some trial data to that experiment by "running" it.
+    with scheduler:
+        scheduler.start()
+        scheduler.teardown()
```
bpkroth, line 174: This basically replaces all of the prior dummy run code, but also limits the metadata we've injected into the Trial, so we need to adjust some of the expected test output checks as well.

motus (approved):
neat!

PR #864, opened 2024-10-02
Simplify devcontainer

bpkroth:
# Pull Request

## Title

Combines container build steps to reduce devcontainer size substantially.

---

## Description

Reduces container size from ~5.7GB to ~2.7GB by combining the following steps:

- install conda (previously in a base image layer)
- create the base mlos environment

This reduces the space substantially because conda's usual attempt to do hardlinks across environments is actually able to work.

The downside is that changes to base package level requirements require reinstalling all of conda again (i.e., the single large combined container layer is less cacheable).

Should help with issues like: https://github.com/Microsoft-CISL/sqlite-autotuning/issues/7

---

## Type of Change

- 🔄 Refactor
- Dev environment

---

## Testing

- `time make devcontainer`
- "Rebuild devcontainer" inside VSCode
- `docker image ls` to check the sizes

---


motus (approved):

.devcontainer/Dockerfile
```diff
@@ -1,27 +1,43 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-FROM mcr.microsoft.com/devcontainers/miniconda:3 AS base
+FROM mcr.microsoft.com/vscode/devcontainers/base AS base
```
bpkroth, line 4: This is the main change.
It reduces the initial layer to something generally pretty small.
However, it doesn't include miniconda, so the next step is to reinstall miniconda.
However, we do that along with preparing the mlos conda env all in one step since conda uses hardlinks to save space and combining those into a single step allows tar (the tool used by docker) retain that optimization rather than making copies across docker layers.

PR #868, opened 2024-10-04
Allow empty tunable values to represent the defaults

bpkroth:
# Pull Request

## Title

Allows an empty dictionary (object) to represent the default tunable values for tunable params.

---

## Description

This should make it easier to maintain configs that loop over a chosen set of tunable values, where some of the tunable values are the default values without needing to copy/paste those values from the tunable params definitions.

- See Also: [discussion in #855](https://github.com/microsoft/MLOS/pull/855#discussion_r1786896189)
---

## Type of Change

- ✨ New feature

---

## Testing

Extends existing testing infrastructure to check that this works.

---

### DIFF
--- a/mlos_bench/mlos_bench/config/schemas/tunables/tunable-values-schema.json
+++ b/mlos_bench/mlos_bench/config/schemas/tunables/tunable-values-schema.json
@@ -12,7 +12,6 @@
                     "type": ["string", "number", "boolean"]
                 }
             },
-            "minProperties": 1,
             "not": {
                 "required": ["tunable_values"]
             }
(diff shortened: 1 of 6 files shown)


bpkroth:
As noted in the linked discussion, an alternative could be to allow `null` to represent the defaults, but this seemed reasonably clean.

motus (approved):

motus:
Looks good! Maybe, all one more unit test with _some_ tunable values omitted to check that they are set to default and others are not?

bpkroth:
> Looks good! Maybe, all one more unit test with _some_ tunable values omitted to check that they are set to default and others are not?

I think that's handled already in the test I added.  I made it a little bit more explicit just now as well.

PR #887, opened 2024-11-22
More documentation updates

bpkroth:
# Pull Request

## Title

More documentation updates

---

## Description

- [x] Adds general documentation on philosophy and usage
- [x] Additional docs about Config structure
- [x] Adjusts README documentation links
- [x] Additional documentation about Storage classes

---

## Type of Change

- 📝 Documentation update
- 🧪 Tests

---

## Additional Notes (optional)

Follow on to #869

---

PR #890, opened 2024-12-04
Allow loading configs from JSON strings and test with new docstrings for tunables

### LINKED ISSUE #881: Add ability to load class instances from json config string, not just a file
This is mostly a nicety for writing interactive documentation in #869 

Most user interactions with `mlos_bench` go through json configs, so that's really what we want to document the most.

I'd like to make those more easily browsable in the Sphinx generated docs, which means putting the json example configs into docstrings, but we want to ensure that those continue to load correctly by using `doctest`

Right now, `ConfigPersistenceService` only does `load_environment` via a file, but it'd be pretty easy to extend it to support a json config string as well.

bpkroth:
# Pull Request

## Title

Allow loading configs from JSON strings not just files and add docs for Tunables to demonstrate and test it.

---

## Description

Adds basic support for loading a JSON config string directly.

The main purpose is to be able to show examples in docstrings that are testable (so that they don't become stale examples).

- [x] Add basic example of usage and testing in a tunables docstring example.
- [x] Add `pytest` unit tests for config string loading as well (we already have file loading tests).
- Closes #881 

---

## Type of Change

- ✨ New feature
- 📝 Documentation update
- 🧪 Tests

---

## Testing

- [x] Adds additional unit tests.

---

Split out from #887 

---

### DIFF
--- a/mlos_bench/mlos_bench/environments/__init__.py
+++ b/mlos_bench/mlos_bench/environments/__init__.py
@@ -116,9 +116,11 @@ Notes
 
 See Also
 --------
-:py:mod:`mlos_bench.config`
-:py:mod:`mlos_bench.services`
-:py:mod:`mlos_bench.tunables`
+:py:mod:`mlos_bench.config` : Overview of the configuration system.
+:py:mod:`mlos_bench.services` : Overview of the Services available to the
+    Environments and their configurations.
+:py:mod:`mlos_bench.tunables` : Overview of the Tunables available to the
+    Environments and their configurations.
 """
 
 from mlos_bench.environments.base_environment import Environment
--- a/mlos_bench/mlos_bench/environments/__init__.py
+++ b/mlos_bench/mlos_bench/environments/__init__.py
@@ -113,6 +113,12 @@ Notes
 - See `mlos_bench/config/environments/README.md
   <https://github.com/microsoft/MLOS/tree/main/mlos_bench/mlos_bench/config/environments/>`_
   for additional config examples in the source tree.
+
+See Also
+--------
+:py:mod:`mlos_bench.config`
+:py:mod:`mlos_bench.services`
+:py:mod:`mlos_bench.tunables`
 """
 
 from mlos_bench.environments.base_environment import Environment
--- a/mlos_bench/mlos_bench/services/config_persistence.py
+++ b/mlos_bench/mlos_bench/services/config_persistence.py
@@ -187,7 +187,11 @@ class ConfigPersistenceService(Service, SupportsConfigLoading):
             # If the path contains braces, it is likely already a json string,
             # so just parse it.
             _LOG.info("Load config from json string: %s", json)
-            config: Any = json5.loads(json)
+            try:
+                config: Any = json5.loads(json)
+            except ValueError as ex:
+                _LOG.error("Failed to parse config from JSON string: %s", json)
+                raise ValueError(f"Failed to parse config from JSON string: {json}") from ex
         else:
             json = self.resolve_path(json)
             _LOG.info("Load config file: %s", json)
--- a/mlos_bench/mlos_bench/services/config_persistence.py
+++ b/mlos_bench/mlos_bench/services/config_persistence.py
@@ -58,8 +58,9 @@ class ConfigPersistenceService(Service, SupportsConfigLoading):
     """
 
     BUILTIN_CONFIG_PATH = str(files("mlos_bench.config").joinpath("")).replace("\\", "/")
-    """A calculated path to the built-in configuration files shipped with the
-    mlos_bench package."""
+    """A calculated path to the built-in configuration files shipped with the mlos_bench
+    package.
+    """
 
     def __init__(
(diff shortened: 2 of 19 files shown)


mlos_bench/mlos_bench/tunables/tunable.py
bpkroth, line 1: docstrings for variables go under the variable

mlos_bench/mlos_bench/tunables/__init__.py
```diff
@@ -2,7 +2,209 @@
...
+...         // Quantize the range into 100 bins
+...         "quantization_bins": 100,
+...         // enable log sampling of the bins
+...         "log": true
+...       }
+...     }
+...   }
+... }
+... '''
+>>> tunables = service.load_tunables(jsons=[json_config])
```
bpkroth, line 163: Test of loading configs via a JSON string.
Helps keep the docs functional.

mlos_bench/mlos_bench/services/config_persistence.py
```diff
@@ -661,42 +674,44 @@ def load_services(
...
         service = Service({}, global_config, parent)
-        for fname in json_file_names:
-            config = self.load_config(fname, ConfigSchema.SERVICE)
+        for json in jsons:
+            config = self.load_config(json, ConfigSchema.SERVICE)
             service.register(self.build_service(config, global_config, service).export())
         return service
 
-    def _load_tunables(
+    def load_tunables(
```
bpkroth, line 687: Make public for testing.
Didn't add it to the SupportsConfigLoadingType though, since it's not used anywhere else atm.

mlos_bench/mlos_bench/services/config_persistence.py
```diff
@@ -176,22 +180,28 @@ def load_config(
...
+        if any(c in json for c in ("{", "[")):
+            # If the path contains braces, it is likely already a json string,
+            # so just parse it.
+            _LOG.info("Load config from json string: %s", json)
+            config: Any = json5.loads(json)
+        else:
+            json = self.resolve_path(json)
+            _LOG.info("Load config file: %s", json)
+            with open(json, mode="r", encoding="utf-8") as fh_json:
+                config = json5.load(fh_json)
```
bpkroth, line 195: This is the main change to allow testable docstrings with inline JSON configs.
Basically we just look at the provided string and see if it has a few non-path but JSON characters in it and then attempt to parse the string directly.

mlos_bench/mlos_bench/environments/__init__.py
```diff
@@ -113,6 +113,12 @@
 - See `mlos_bench/config/environments/README.md
   <https://github.com/microsoft/MLOS/tree/main/mlos_bench/mlos_bench/config/environments/>`_
   for additional config examples in the source tree.
+
+See Also
+--------
+:py:mod:`mlos_bench.config`
+:py:mod:`mlos_bench.services`
+:py:mod:`mlos_bench.tunables`
```
bpkroth, line 121: Backreferences between related documentation.

mlos_bench/mlos_bench/services/config_persistence.py
```diff
@@ -176,22 +180,32 @@ def load_config(
...
-            config = json5.load(fh_json)
+        if any(c in json for c in ("{", "[")):
+            # If the path contains braces, it is likely already a json string,
+            # so just parse it.
+            _LOG.info("Load config from json string: %s", json)
+            try:
+                config: Any = json5.loads(json)
+            except ValueError as ex:
+                _LOG.error("Failed to parse config from JSON string: %s", json)
+                raise ValueError(f"Failed to parse config from JSON string: {json}") from ex
```
bpkroth, line 191: Could alternatively try to fall through here and treat it as a file next.

bpkroth: @motus ?

motus (approved):

mlos_bench/mlos_bench/services/config_persistence.py
```diff
@@ -176,22 +183,28 @@ def load_config(
...
+        if any(c in json for c in ("{", "[")):
+            # If the path contains braces, it is likely already a json string,
+            # so just parse it.
+            _LOG.info("Load config from json string: %s", json)
+            config: Any = json5.loads(json)
+        else:
+            json = self.resolve_path(json)
+            _LOG.info("Load config file: %s", json)
+            with open(json, mode="r", encoding="utf-8") as fh_json:
+                config = json5.load(fh_json)
```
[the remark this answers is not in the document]
bpkroth: Think we discussed this offline.
It's not checking the first character.  It's checking *all* the characters in the string (well, until it encounters one of those two brace characters).

If they exist, then it's almost surely a JSON string and not a path, and should exist early-ish in the string (even if there are leading comments).

If they don't exist, then it's a path string, and it's probably less than 256 characters already, so the check is actually pretty cheap (cheaper than the `os.path.exists()` check at least).

Issue #920, opened 2025-01-10
Make sure we don't use raw strings for Status comparisons in the Storage backend


bpkroth:
that will produce a string `'Status.PENDING'`. Did you mean
```suggestion
                            self._schema.trial.c.status == Status.PENDING.name,
```
?

(Note to self: use `Status.*.name` everywhere in storage instead of string literals)

(Note 2: need to check for such things in unit tests)

_Originally posted by @motus in https://github.com/microsoft/MLOS/pull/720#discussion_r1909607472_

PR #921, opened 2025-01-10
use named constants instead of string literals when serializing Status values for storage

motus:
# Pull Request

## Title

Use constants instead of string literals when serializing `Status` values in MLOS storage

---

## Description

Replace all string literals for `Status` values in storage with corresponding `Status.*.name` constants.

- **Issue link**: Closes: #920 

---

## Type of Change

- 🔄 Refactor


bpkroth (approved):
LGTM

There's a related FIXME comment that was added in #720 regarding Trial's being created with UNKNOWN status but entered into the db with PENDING status, but we can fix that in a separate PR I think.

bpkroth:
Needs a CI fix from #720

PR #923, opened 2025-01-10
Create Trial with Status that matches Storage backend

bpkroth:
# Pull Request

## Title

Create Trial with Status that matches Storage backend

---

## Description

Previously Trial objects were created with an UNKNOWN status, but stored in the db with a PENDING status.

This PR corrects that mismatch.

---

## Type of Change

- 🛠️ Bug fix

---

## Testing

New unit test to check for matching status.

---


bpkroth:
See Also #921

motus (approved):

Issue #732, opened 2024-05-10
mlos_bench_service


bpkroth:
stub for now

bpkroth:
@eujing

bpkroth:
May want to split some of these tasks out to separate issues later on

yshady:
@bpkroth  should be easy I have examples of interactive notebooks in my internal repo as well as the streamlit being quite a transferable process to a notebook experience!

yshady:
can follow same workflow as the sidemenu

yshady:
Requirement would be a user can run at least one experiment manually first

eujing:
From the mysql side, we currently have something similar that I can work on generalizing. It is basically a FastAPI app (we have been calling it a "runner") with the following endpoints.

Experiment-related:
- GET /experiments -> Listing experiments
- POST /experiments/start -> JSON body (and associated pydantic model for validation) to create experiment config files, and essentially does `docker run {mlos_bench image}` with relevant arguments
- POST /experiments/stop/{experiment_id} -> No body, but essentially does `docker stop {image name}`

Front-end related, mainly for populating the JSON body to POST /experiments/start:
- GET /options/mlos_configs -> List CLI config files for use with `mlos_bench --config`'s value.
- GET /options/benchbase_configs -> List benchbase XML config files
- GET /options/tunable_groups -> List tunable group names, for selection to include in an experiment
- GET /options/client_skus -> List available client VM SKUs for the subscription (makes an az API call)
- GET /options/server_skus -> List available server SKUs (read off a curated CSV file)

We have 3 docker images, for this runner, the dashboard that uses it, and `mlos_bench` itself.
The first two are started with docker compose, along with nginx to help with some auth and HTTPS access to the dashboard.
Mounts for the runner container:
- The base directory of the repo for our MLOS configs. This is also mounted to each "child" `mlos_bench` container that is started, including the relevant generated experiment config file.
- The host's docker socket, to allow management of multiple mlos_bench containers.

yshady:
Im really glad you guys are still using FastAPI from summer, and even building it out. Happy to see it!

bpkroth:
Portions of this make sense, but I'd rather have it do more iteraction with the storage backend, particularly on the runner side of things.

Right now there's basically an assumption that `POST /experiments/start/experiment_id` directly invokes a `docker run {mlos_bench image}` but that will cause scaling issues, especially if training the model remains local to the `mlos_bench` process as it is now.

[...]

Issue #838, opened 2024-08-12
mlos_webui design requirements


bpkroth:
For starters, `mlos_webui` seems an appropriate moniker to me.  It should support

1. [ ] Visualization of `ExperimentData` from a shared `Storage` location.
3. [ ] Execution of existing `mlos_bench` configs it has access to.
   (e.g., via SP access to a git repo).
   - [ ] We should allow it to be parallelized across Experiments via GridSearch (e.g., to schedule different instances of a tuning Experiment across different scenarios, like storage types or VM sizes, or benchmark type).
   - [ ] Error logs should be visible.
   - [ ] Experiments should be resumable.
4. [ ] Editing of such configs.
   (with versioning)

- [ ] Authentication should be required.
- [ ] Fine grained authorization (e.g., restricting access to experiments by group or some such) is out of scope *for now* (i.e., if you can authenticate, you can perform any action).
- [ ] Code should be modular enough to allow easily adding features (e.g., new visualizations).

_Originally posted by @bpkroth in https://github.com/microsoft/MLOS/pull/824#discussion_r1714331354_

bpkroth:
@eujing 
@yshady

bpkroth:
See Also #732

yshady:
Thanks for compiling a list of tasks/features @bpkroth 

I’d also love support for users to be able to 
- be able to configure benchmarks from a gui 
- Error logs should show the end x lines and parse out potential error lines (user preference)
- be able to monitor active experiments and see which launched experiments have quit before hitting the needed trial repeat number, this can be used as a tracking method 
- automated correlation detection / validation between Params and a target metric would be a heaven sent 
- for a experiment grid showing pie chart for each part where user can see pending, failed, and success metrics 

Essentially if we can map out everything we do into a web up with no code I think this will be a far better costumer experience for those who may want to create their own benchmarks and their own tuning teams

yshady:
I think creating a validation system where controls and optimized configs are done automatically on some level will be super valuable.

instead on launching many experiments manually we might want to create a system to find and validate a number of configs given a estimated monetary budget

yshady-acheev:
> For starters, `mlos_webui` seems an appropriate moniker to me. It should support
> 
> 1. [ ]  Visualization of `ExperimentData` from a shared `Storage` location.2. [ ]  Execution of existing `mlos_bench` configs it has access to.
>       (e.g., via SP access to a git repo).
>       
[...]

PR #931, opened 2025-01-14
Schema changes example for mlos_benchd service

bpkroth:
# Pull Request

## Title

Schema changes for mlos_benchd service.

## Description

Schema changes for mlos_benchd service.

Storage APIs to adjust these to come in a future PR.

- See #732

## Type of Change

- ✨ New feature

## Testing

Local, CI

@eujing have a look at the [commit history in the PR](https://github.com/microsoft/MLOS/pull/931/commits) for a sense of what's going on.
This can probably be merged as is. Happy to discuss further though.


bpkroth:
Rough idea here is in the comments in the `schema.py` file.

At a high level, imagine that we have 1+ VMs running `mlos_benchd` as a service, continually polling the backend storage to see if there are new Experiments submitted that are `PENDING`.

> Note: Here a "Worker" runs `mlos_benchd` to spawn multiple `mlos_bench` processes, one per `Experiment`.  Where as within each `mlos_bench` 1+ `TrialRunners` may run multiple `Trials` for that `Experiment` in parallel (wip - #380).

At `Worker` starts a transaction:

```python
# quick pseudo code
while true:
  sleep(1)
  with conn.begin() as conn:
     try: 
       experiment_row = conn.select(schema.experiments)
           .where(
               schema.experiments.status == Status.PENDING.name,
               schema.experiments.worker_name is None,
               schema.experiments.start_ts >= datetime.utcnow(),
           )
           .limit(1)
      if experiment_row:
         # try to grab
         result = conn.update(schema.experiments)
           .values({
              schema.experiments.worker_name: sys.hostname,
              schema.experiments.status: status.READY.name,
           })
           .where(
              schema.experiments.worker_name is None,
              schema.experiments.experiment_id == experiment_row.experiment_id,
            )
         if result:
           # succeeded, commit the transaction and return
           conn.commit()
           # return this to calling code to spawn a new `mlos_bench`
           # process to fork and execute this Experiment on this host
           # in the background
           return experiment_row.experiment_id
         else:
           # someone else probably grabbed it
           conn.rollback()
  except SqlException as e:
     # probably a conflict 
     conn.rollback()
# try again in a moment
```

mlos_bench/mlos_bench/environments/status.py
bpkroth, line 1: Could probably move this change to a different PR.

mlos_bench/mlos_bench/storage/sql/schema.py
eujing, line 1: I think these column additions to `Experiment` make sense to me!

However, I was thinking we might need additional key-value information for optional global override values, which we rely a lot on for launching experiments.
We could rely on the git functionality to pull in all required global JSON files, but there are several sensitive overrides like credentials that are usually not committed into git.

What do you think of having an additional JSON column (or separate key-value table) for flexibility with these overrides? Just thinking further ahead for execution of these submitted experiments, but can be done in a follow-up PR.

bpkroth: Yeah all of that is relevant but I haven't settled on quite the right format yet.
Instead of always using git we could have the webui store it in some share/blob store and then pull that down to the worker when necessary.  That way we wouldn't need the credentials for the git side long term.
As for the secrets, management of that as a service is also challenging.  KeyVaults are the usual strategy, but I'm not quite sure how to make that cloud agnostic atm.

Anyways, I think it better to separate these two things.

1. Create a basic mlos_benchd service that assuming it already has local access to the appropriate configs can execute the mlos_bench command with the right CLI options as necessary.
   There's plenty of other things involved in getting that to work including changes to the storage APIs, potentially a web frontend to make use of them (though I think notebooks are still probably fine for the initial hacking around), the `mlos_benchd` service script itself, etc.
2. Allow the backend storage to keep and manage all the configs directly.

eujing: Ok agreed! I can start working with that step 1 first.

eujing (approved):

mlos_bench/mlos_bench/storage/sql/schema.py
bpkroth: @motus do you want to comment on this at all before we proceed?

motus: I was about to approve that PR yesterday, but now I have second thoughts. We're breaking the 3NF and I am afraid the redundant columns will just confuse the users.

  * First, the timestamps. This information can be easily retrieved from the `trial` table as `MIN(ts_start), MAX(ts_end)` - unless we want to use separate timestamps for the experiment itself - e.g., when we create a new experiment that is the result of a merge of several other experiments, etc.
  * Also, the semantics of `ts_end` is not clear to me. Is that the timestamp of the end of the last trial, or we populate it only when the experiment is over (i.e., we are done with _all_ trials)?
  * Same about the `status` - is it the status of the current running trial, or the last completed trial, or we update it only when we are done with all trials of the experiment? What if we want to run a few extra trials after its completion?
  * Also, can the experiment have status like `FAILED` or `TIMED_OUT` and what is the meaning of such status when applied to the entire experiment?
  * Finally, the `worker_name` (maybe, `worker_hostname`?) and `worker_pid` fields - I assume they identify the process that runs MLOS (i.e., the scheduler and optimizer). This is a bit confusing because I thought that in the distributed/parallel setup we'll have multiple workers running trials for the same experiment - and the `trial` table will have the corresponding fields or key/value parameters. It would make sense call the instance that runs the scheduler the *Driver*. Please correct me if I am wrong

Anyway, I'll approve the PR as it looks perfectly fine from the technical point of view, but let's discuss this change briefly before merging.

motus (approved):
Approved, but let's have a short discussion before merging it in

mlos_bench/mlos_bench/storage/sql/schema.py
bpkroth: The columns don't really need to be visible to users directly.  They're meant to interact with this thru (to be developed) Storage APIs for creating and scheduling Experiments.

- I agree that Experiment doesn't need a `ts_end`, but a `ts_start` could be used very similarly to a `ts_start` for Trials.  Basically it would delay the `mlos_benchd` service from starting an `mlos_bench` instance for that Experiment until the appointed `ts_start`.  @eujing for instance mentioned wanting to schedule regular benchmarks to evaluate service performance over time.  Simply using `MIN(trial_start)` would likely be insufficient since it would require creating both an Experiment entry and a Trial.  The latter requires loading additional configs for the Experiment to create the extra entries and is usually the realm of `mlos_bench`, but we could adjust this to make that the protocol instead if we really wanted to.
- `status` I think the `mlos_benchd` service can use to update the last known state of the child `mlos_bench` instance.  Leaving it in a `PENDING` state means it hasn't been started yet.  `FAILED` could indicate that the child process exited non-zero.  `TIMED_OUT` may not have a corollary, but I think that's OK.  Unless we extend the Stopping Condtions (#427) to include things like `max_budget` or `max_time` for an Experiment as well.
- Can absolutely rename to `worker_hostname` instead of `worker_name`.
  And yes, as I mentioned above, there are two separate notions of parallelism here:
  - Experiments (e.g., two devs/users running their own validation of MySQL configs or on different benchmarks, etc.)
  - Within an Experiment, multiple Trials (via TrialRunners)
  
  And sure, we could adjust the terminology to be `Driver` instead of `Worker` if we want.
  
  My thought for the moment was that since currently `mlos_bench` does the model training (we may change that to outsource it in the future, but for now), it could become a bottleneck if many Experiments want to run in parallel, so we might want to have multiple VMs running `mlos_benchd` processes, and in turn `mlos_bench` children.
  
  The would all poll the Storage backend to pick up work to do.
  
  A Frontend client (#838) would be responsible for injecting data to the backend Storage via the Storage APIs (again, possible to have more than one of these, e.g., a Notebook interface as well).

Make sense?

motus: @bpkroth makes perfect sense. I think I've overreacted a bit 😺 Let's rename the `worker_*` columns and merge it in.

P.S. rename them to `driver_hostname` and `dirver_pid`, maybe? anything but worker* will do

bpkroth: `s/worker/driver/g`

What about `ts_end`?  Remove it?

bpkroth: Rename is ok.  I think we'll leave the ts_end for now.

Issue #926, opened 2025-01-13
Implement TUNA in mlos_bench


bpkroth:
Placeholder issue to track to implementing the TUNA noisy optimization algorithm in an mlos_bench scheduler now that the TrialRunner abstraction is available (arxiv citation to be added).

May also want to finish implementing a basic ParallelTrialScheduler first (#380).

Probably also need to finish making some singleton VM ARM templates to make use of that (e.g., parameterized by the `$trial_runner_id`).

@jsfreischuetz for awareness and tracking

bpkroth:
Adding a few of my notes from our discussion.  @jsfreischuetz please adjust/correct as you see them (just wanted to get them down while they were fresh in my mind).

# PRs needed for different features

1. Basic `ParallelTrialScheduler`
  #380
  #939
  Contrary to our discussion, since MLOS already provides `--trial-config-repeat-count` to repeat a config some number of times, this would effectively already be the same as the `null` aggregation policy we discussed (i.e., simply re-register the score each time we finish a config trial).

2. Extend `ParallelTrialScheduler` to support new aggregation policies.

  This is where we can include config syntax to select the config score aggregation policy (e.g., `min`, `avg`, `avg - stddev`, etc.)

  May need some facilities to delay registering until after some number of results have returned.

  Later as we increase the repeat budget we may need to consider rebuilding the optimizer in order "re-register" the new aggregated score at the higher budget.

  > Come to think of it, we may need to have a separate PR on incorporating budget from the `Optimizer` (and in turn `mlos_core`) or managing that only in the `Scheduler`.

  This may also impact how `--trial-config-repeat-count` gets interpretted (e.g., "max budget repeats with a fixed increase step function?", or something more flexibly defined for that Scheduler type)

  This may also interact with the `register_pending` APIs in mlos_core in order to mark a config as "in progress".

3. Add outlier detection capabilities to the aggregation policy.

  Make sure this has a configurable variance limit with a default of 30% (per the paper).

4. Configs for system stats collection.

  These should be relatively generic and able to be done for any Linux OS.

  We could implement these only as a `RemoteEnv` config, but I think it will be better to implement as a new `OsStatsEnv` (possibly as a subclass of `RemoteEnv` with stock configs that tell it how to do it) and `SupportsOsStatsService` type that can manage this in a handful of ways.

[...]

PR #939, opened 2025-01-22
WIP: Parallel trial scheduler stub using multiprocessing.Pool

bpkroth:
@jsfreischuetz here's the stub idea I had for making a parallel trial scheduler using `multiprocessing.Pool`.

This would be implemented in something like a class `ParallelTrialScheduler` in the `mlos_bench.schedulers` module.

Perhaps some aspects of the APIs would have to change, and certainly lots and lots of tests and some error handling and interaction with the Storage backend instead of just local `dicts` for state management.

I think we should probably start with something simple like this in order to complete #380. Maybe for starters it can just handle fanning out the execution of trial repeats across backend workers.

Then later your work for #926 can expand it do budgeted config scheduling and then later model adjusted scoring and unstable config detection.

For testing, I'm thinking we setup something like `LocalEnv` and just have them run a simple script that returns known data (e.g., maybe even just an echo back the value it's given in the config or a computation of it or some such).

Let me know if you have questions.

### DIFF
--- a/mlos_bench/mlos_bench/schedulers/forking_worker_pool_example.py
+++ b/mlos_bench/mlos_bench/schedulers/forking_worker_pool_example.py
@@ -1,0 +1,205 @@
+#
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+#
+"""Simple parallel trial scheduler and optimization loop implementation stub code."""
+import json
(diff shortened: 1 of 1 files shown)


bpkroth:
Closing this as it was only meant as an example anyways.

PR #824, opened 2024-07-26
GUI panel for mlos_bench

yshady:

PR #944, opened 2025-01-23
Add Enhanced MLOS Analyzer Dashboard with Advanced Visualizations as a Library

yshady-acheev:
Add Enhanced MLOS Analyzer Dashboard with Advanced Visualizations

Added comprehensive dashboard for MLOS experiment analysis with new visualization modules:
- Time series analysis with moving averages
- Distribution plots and violin comparisons
- Performance radar charts
- Parallel coordinates visualization
- Enhanced statistical analysis

Type of Change:
- ✨ New feature
- 🔄 Refactor Skeleton Library for GUI #824 

Testing:
- Tested visualization modules with sample experiment data using SQLite
- Verified all plots render correctly
- Validated statistical analysis outputs
- Confirmed interactive features work as expected
- Checked performance with large datasets

Additional Notes:
Code organization follows src layout pattern. Added type hints and docstrings for maintainability.

### DIFF
--- a/mlos_analyzer/README.md
+++ b/mlos_analyzer/README.md
@@ -1,0 +1,16 @@
+# MLOS Analyzer
+
+A comprehensive library for analyzing and visualizing MLOS experiment results.
+
+## Installation
+```bash
+pip install -r requirements.txt
+python setup.py install
+```
+
+## Features
+- FastAPI backend for experiment data retrieval
+- Correlation analysis
+- Failure metrics visualization
+- Statistical analysis
+- Advanced plotting capabilities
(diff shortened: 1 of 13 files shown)


yshady-acheev:
> I think we discussed putting these into the `mlos_viz` package.
> 
> I don't think we need to introduce a new package for these.

My understanding is it was going to be part of "mlos_webui"

bpkroth:
> > I think we discussed putting these into the `mlos_viz` package.
> > I don't think we need to introduce a new package for these.
> 
> My understanding is it was going to be part of "mlos_webui" or something like this and it should be seperate based on #838
> 
> I also assumed it would someway be put into mlos_viz later on
> 
> I can reformat later this month

To my mind most of what's here are the data analysis/visualization routines.  Those should go in mlos_viz.

After that, folks can choose to call them in either a notebook or we can make use of them in the mlos_webui, but that will be a separate PR.  Make sense?

Also, remember that we'll want to enable all the usual code quality checks throughout these (e.g., type checks, linting, formatting, etc.).

Thanks!

PR #945, opened 2025-01-25
Fix dtype casting issue with numpy in LlamaTune adapter

### LINKED ISSUE #935: LlamaTune incompatible with numpy>=2.0
LlamaTune has issues with numpy>=2.0 reporting errors like the following:

```
FAILED mlos_core/mlos_core/tests/spaces/adapters/llamatune_test.py::test_num_low_dims[2-param_space_kwargs1] - ValueError: Configuration(values={
  'int_0': 245,
  'int_1': 10,
  'int_2': 54,
})
The above configuration was not suggested by the optimizer. Approximate reverse mapping is currently disabled; thus *only* configurations suggested previously by the optimizer can be registered.
```

@kkanellis do you think you could have a look at this?

kkanellis:
# Pull Request

## Description

This PR addresses a breaking change in NumPy's [universal function](https://numpy.org/doc/stable/reference/ufuncs.html) (ufunc) behavior in version 2.0, specifically affecting `np.clip` in the LlamaTune adapter code.

**Key Changes**:
- Resolves an issue where configuration retrieval fails due to type casting differences
- Explicitly casts `np.clip` output from NumPy data type (e.g., `int64`) to native Python type (e.g., `int`)

**Problem Context:**
In NumPy 2.0+, when ufuncs receive Python scalars as input, the output is [no longer](https://numpy.org/doc/stable/reference/generated/numpy.can_cast.html#numpy.can_cast) automatically cast to Python scalars. This prevents retrieving previously suggested configurations stored in `self._suggested_configs` dict during `self.inverse_transform` method calls.

**Technical Solution:**
Implement explicit type casting to ensure consistent configuration representation in `self._suggested_configs` across `self.transform` and `self.inverse_transform` method calls.

- **Closes #935** 

## Type of Change

- 🛠️ Bug fix

## Testing

- **Environment**: Ubuntu 22.04 with Python 3.13
- **NumPy Versions Tested**: 1.26.4 and 2.2.2
- **Verification**: Existing test suite passed across both versions

### DIFF
--- a/mlos_core/mlos_core/spaces/adapters/llamatune.py
+++ b/mlos_core/mlos_core/spaces/adapters/llamatune.py
@@ -16,7 +16,6 @@ See Also: `LlamaTune: Sample-Efficient DBMS Configuration Tuning
 import os
 from importlib.metadata import version
 from typing import Any
-from packaging.version import Version
 from warnings import warn
 
 import ConfigSpace
@@ -25,12 +24,14 @@ import numpy as np
 import numpy.typing as npt
 import pandas as pd
 from ConfigSpace.hyperparameters import NumericalHyperparameter
+from packaging.version import Version
 from sklearn.preprocessing import MinMaxScaler
 
 from mlos_core.spaces.adapters.adapter import BaseSpaceAdapter
 from mlos_core.util import normalize_config
 
 _NUMPY_VERS = Version(version("numpy"))
+
 
 class LlamaTuneAdapter(BaseSpaceAdapter):  # pylint: disable=too-many-instance-attributes
     """Implementation of LlamaTune, a set of parameter space transformation techniques,
--- a/mlos_core/mlos_core/spaces/adapters/llamatune.py
+++ b/mlos_core/mlos_core/spaces/adapters/llamatune.py
@@ -385,6 +385,15 @@ class LlamaTuneAdapter(BaseSpaceAdapter):  # pylint: disable=too-many-instance-a
 
                 orig_value = param.to_value(value)
                 orig_value = np.clip(orig_value, param.lower, param.upper)
+
+                try:
+                    # Convert numpy types to native Python types (e.g., np.int64 to int)
+                    # This was performed automatically in NumPy<2.0, but not anymore.
+                    # see, https://numpy.org/doc/stable/reference/generated/numpy.can_cast.html
+                    orig_value = orig_value.item()
+                except AttributeError:
+                    pass
+
             else:
                 raise NotImplementedError(
                     "Only Categorical, Integer, and Float hyperparameters are currently supported."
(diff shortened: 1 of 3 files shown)


bpkroth:
Thanks for digging into this. Adjusting slightly to also remove the numpy version pinning.

mlos_viz/mlos_viz/base.py
```diff
@@ -40,7 +41,7 @@ def ignore_plotter_warnings() -> None:
     adding them to the warnings filter.
     """
     warnings.filterwarnings("ignore", category=FutureWarning)
-    if _SEABORN_VERS <= "0.13.1":
+    if _SEABORN_VERS <= Version("0.13.1"):
```
bpkroth, line 44: Unrelated bug, but fixing it here too.

mlos_core/mlos_core/spaces/adapters/llamatune.py
```diff
@@ -385,6 +389,13 @@ def _transform(self, configuration: dict) -> dict:
 
                 orig_value = param.to_value(value)
                 orig_value = np.clip(orig_value, param.lower, param.upper)
+
+                if _NUMPY_VERS >= Version("2.0"):
```
bpkroth, line 393: Made the try/except a little more explicit with a version check.

bpkroth (dismissed):
Made a few small tweaks, but LGTM. Thanks for tracking this down!

bpkroth:
Actually, it looks like this is still failing a test:

```
pytest -n0 mlos_bench/mlos_bench/tests/optimizers/toy_optimization_loop_test.py -k smac

...

    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/opt/conda/envs/mlos/lib/python3.13/site-packages/smac/main/smbo.py", line 431, in save
    self._runhistory.save(path / "runhistory.json")
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/envs/mlos/lib/python3.13/site-packages/smac/runhistory/runhistory.py", line 805, in save
    assert self._running == len(self._running_trials)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

That doesn't happen with numpy<2.0, but with numpy 2.2.2 it does happen whether I use the version check or the try/except.

Can you please take another look?

Thanks!

bpkroth:
@kkanellis were you able to track the source of this down at all?
From my quick glance, you're right that smac3 updated to v2.3.0 and started allowing numpy 2.0 (prior apparently both we and smac3 were disallowing it).
However, now there's a different bug with smac3.
Would be good to at least track that down and submit an Issue upstream to them.

kkanellis:
I will try to track this down tomorrow.

In any case, since this look like a different bug/issue with SMAC, I propose merging this (w/o the change on the numpy version), and open a new issue just for this.

bpkroth (approved):
As discussed, merging as is since we are fairly confident this fixes one issue.
Will lift the numpy 2.0 restriction later once the SMAC issue is tracked down.

PR #949, opened 2025-02-05
Basic implementation for mlos_benchd service

eujing:
# Pull Request

## Basic implementation for mlos_benchd service

This PR introduces:
- Creation of `Experiment` via the storage API, separate from running via CLI.
- A new `mlos_benchd` script that polls storage for runnable experiments, then executes them.

This allows for the separation of experiment creation (e.g. scheduling), and their execution (on potentially multiple hosts).
The new `mlos_benchd` script can run on any host to poll for new experiments, as long as they monitor the right storage backend.

Builds upon schema changes from #931 

### Example usage

Create new experiment, via notebook or any python environment:
```python
from mlos_bench.storage import from_config
storage = from_config(config="mlos_bench/mlos_bench/config/storage/sqlite.jsonc")
exp = storage.experiment(experiment_id="test-exp-01", trial_id=0, root_env_config="mlos_bench/mlos_bench/config/environments/apps/fake/test_local_env.jsonc", description="description", tunables={}, opt_targets={"score": "min"})
exp.save()
```

mlos_benchd:
```sh
/workspaces/MLOS (main) $ python mlos_bench/mlos_bench/mlos_benchd.py --storage "mlos_bench/mlos_bench/config/storage/sqlite.jsonc"
...
No runnable experiment found. Sleeping for 1 second.
No runnable experiment found. Sleeping for 1 second.
No runnable experiment found. Sleeping for 1 second.
...
2025-02-05 20:35:06,208 launcher.py:51 __init__ INFO Launch: mlos_bench
...
2025-02-05 20:35:08,270 base_scheduler.py:275 get_best_observation INFO Env: Local Shell Test Environment best score: {'score': 123.4}
2025-02-05 20:35:08,271 run.py:74 _main INFO Final score: {'score': 123.4}
No runnable experiment found. Sleeping for 1 second.
```

Limitations:
- Assumes all experiment-related config files are already available on the host
- Does not support per-experiment global overrides yet

## Description

- **Issue link**: #732 

## Type of Change

_Indicate the type of change by choosing one (or more) of the following:_

- ✨ New feature

## Testing

Manual testing, unit tests to come.

_Add any additional context or information for reviewers._

### DIFF
--- a/mlos_bench/mlos_bench/mlos_benchd.py
+++ b/mlos_bench/mlos_bench/mlos_benchd.py
@@ -1,0 +1,73 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+#
+"""
+mlos_bench background execution daemon.
+
+This script is responsible for polling the storage for runnable experiments and
+executing them in parallel.
+
+See the current ``--help`` `output for details.
+"""
+import argparse
+import time
+from concurrent.futures import ProcessPoolExecutor
+
+from mlos_bench.run import _main as mlos_bench_main
+from mlos_bench.storage import from_config
+
+
+def _main(args: argparse.Namespace) -> None:
+    storage = from_config(config=args.storage)
+    poll_interval = float(args.poll_interval)
+    num_workers = int(args.num_workers)
+
+    with ProcessPoolExecutor(max_workers=num_workers) as executor:
+        while True:
+            exp_id = storage.get_runnable_experiment()
+            if exp_id is None:
+                print(f"No runnable experiment found. Sleeping for {poll_interval} second(s).")
+                time.sleep(poll_interval)
+                continue
+
+            exp = storage.experiments[exp_id]
+            root_env_config, _, _ = exp.root_env_config
+
+            executor.submit(
+                mlos_bench_main,
+                [
+                    "--storage",
+                    args.storage,
+                    "--environment",
+                    root_env_config,
+                    "--experiment_id",
+                    exp_id,
+                ],
+            )
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="mlos_benchd")
+    parser.add_argument(
+        "--storage",
+        required=False,
+        help=(
+            "Path to the storage configuration file. "
+            "If omitted, use the ephemeral in-memory SQL storage."
+        ),
+    )
+    parser.add_argument(
+        "--num_workers",
+        required=False,
+        default=1,
+        help="Number of workers to use. Default is 1.",
+    )
+    parser.add_argument(
+        "--poll_interval",
+        required=False,
+        default=1,
+        help="Polling interval in seconds. Default is 1.",
+    )
+    _main(parser.parse_args())
--- a/mlos_bench/mlos_bench/storage/base_storage.py
+++ b/mlos_bench/mlos_bench/storage/base_storage.py
@@ -23,12 +23,16 @@ mlos_bench.storage.base_trial_data.TrialData :
 """
 
 import logging
+import os
+import platform
 from abc import ABCMeta, abstractmethod
 from collections.abc import Iterator, Mapping
 from contextlib import AbstractContextManager as ContextManager
 from datetime import datetime
 from types import TracebackType
 from typing import Any, Literal
+
+from pytz import UTC
 
 from mlos_bench.config.schemas import ConfigSchema
 from mlos_bench.dict_templater import DictTemplater
@@ -133,6 +137,17 @@ class Storage(metaclass=ABCMeta):
             the results of the experiment and related data.
         """
 
+    @abstractmethod
+    def get_runnable_experiment(self) -> str | None:
+        """
+        Get the ID of the experiment that can be run.
+
+        Returns
+        -------
+        experiment_id : str | None
+            ID of the experiment that can be run.
+        """
+
     class Experiment(ContextManager, metaclass=ABCMeta):
         # pylint: disable=too-many-instance-attributes
         """
@@ -150,6 +165,7 @@ class Storage(metaclass=ABCMeta):
             root_env_config: str,
             description: str,
             opt_targets: dict[str, Literal["min", "max"]],
+            ts_start: datetime | None = None,
         ):
             self._tunables = tunables.copy()
             self._trial_id = trial_id
@@ -159,6 +175,11 @@ class Storage(metaclass=ABCMeta):
             )
             self._description = description
             self._opt_targets = opt_targets
+            self._ts_start = ts_start or datetime.now(UTC)
+            self._ts_end: datetime | None = None
+            self._status = Status.PENDING
+            self._driver_name: str | None = None
+            self._driver_pid: int | None = None
             self._in_context = False
 
         def __enter__(self) -> "Storage.Experiment":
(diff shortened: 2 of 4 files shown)


mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -159,6 +175,11 @@ def __init__(  # pylint: disable=too-many-arguments
             )
             self._description = description
             self._opt_targets = opt_targets
+            self._ts_start = ts_start or datetime.now(UTC)
+            self._ts_end: datetime | None = None
+            self._status = Status.PENDING
```
bpkroth, line 180: This should match what was stored in the backend for resumable Experiments, right?

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -209,6 +230,9 @@ def _setup(self) -> None:
 
             This method is called by `Storage.Experiment.__enter__()`.
             """
+            self._status = Status.RUNNING
```
bpkroth, line 233: Maybe add some asserts on expected status to check for invalid state transitions.
Wouldn't be terrible to document the expected state transitions in a README.md or docstring either.

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -209,6 +230,9 @@ def _setup(self) -> None:
...
+            self._driver_name = platform.node()
+            self._driver_pid = os.getpid()
```
bpkroth, line 235: These seem ok to initialize the values from `None`, but might be problematic for resuming an Experiment.

There are a few cases I can think of:
1. An individual `mlos_bench` driver process fails and needs to be restarted.
2. The `mlos_benchd` service dies, but the `mlos_bench` process is still running.
3. The whole driver host fails and either needs to be restarted on that same backend or else picked up by a new one (for simplicity, let's assume we only support the former for now.  We can add a heartbeat mechanism later to support the latter).

mlos_bench/mlos_bench/storage/sql/experiment.py
```diff
@@ -49,12 +50,12 @@ def __init__(  # pylint: disable=too-many-arguments
...
             description=description,
             opt_targets=opt_targets,
+            ts_start=ts_start,
         )
         self._engine = engine
         self._schema = schema
 
-    def _setup(self) -> None:
-        super()._setup()
+    def _ensure_persisted(self) -> None:
```
bpkroth, line 58:
```suggestion
    def _save(self) -> None:
```

or 

```suggestion
    def _try_save(self) -> None:
```

or

```suggestion
    def _persist(self) -> None:
```

mlos_bench/mlos_bench/storage/sql/experiment.py
```diff
@@ -125,6 +128,39 @@ def _setup(self) -> None:
                         exp_info.git_commit,
                     )
 
+    def save(self) -> None:
+        self._ensure_persisted()
+
+    def _setup(self) -> None:
+        super()._setup()
+        self._ensure_persisted()
+        with self._engine.begin() as conn:
```
bpkroth, line 137: Might need to separate that out to an `_update` method, or just incorporate it into `_save`

bpkroth: Could also rename `_save` to `_create` to separate the `INSERT` from `UPDATE` calls

mlos_bench/mlos_bench/storage/sql/storage.py
```diff
@@ -109,3 +113,48 @@ def experiments(self) -> dict[str, ExperimentData]:
...
+                            # process to fork and execute this Experiment on this host
+                            # in the background
+                            return str(experiment_row.exp_id)
+                        else:
+                            # someone else probably grabbed it
+                            trans.rollback()
+                except exc.SQLAlchemyError:
+                    # probably a conflict
+                    trans.rollback()
+
```
bpkroth, line 159:
```suggestion
```

mlos_bench/mlos_bench/mlos_benchd.py
```diff
@@ -0,0 +1,70 @@
...
+                    "--environment",
+                    root_env_config,
+                    "--experiment_id",
+                    exp_id,
+                ],
+            )
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="mlos_benchd")
```
bpkroth, line 52: Add a `long` text description, similar to `mlos_bench.run`

mlos_bench/mlos_bench/mlos_benchd.py
```diff
@@ -0,0 +1,70 @@
...
+        help="Path to the storage configuration file.",
+    )
+    parser.add_argument(
+        "--num_workers",
+        required=False,
+        default=1,
+        help="Number of workers to use. Default is 1.",
+    )
+    parser.add_argument(
+        "--poll_interval",
```
bpkroth, line 65: Also provide `poll-interval` variants, like `mlos_bench.launcher`

mlos_bench/mlos_bench/mlos_benchd.py
```diff
@@ -0,0 +1,70 @@
...
+
+            executor.submit(
+                mlos_bench_main,
+                [
+                    "--storage",
+                    args.storage,
+                    "--environment",
+                    root_env_config,
+                    "--experiment_id",
+                    exp_id,
```
bpkroth, line 46: We'll eventually need to include other things here too:
- cli config
- globals (could be subsumed by cli config)
- working dir (maybe we adjust the storage backend to store the target directory and cli-config and optionally globals instead of root_env_config)

bpkroth (commented):
This is a great start!  Thanks so much ❤️ 

I left a few comments for initial changes.

We'll also need tests.

mlos_bench/mlos_bench/mlos_benchd.py
```diff
@@ -0,0 +1,70 @@
...
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+#
+"""
+mlos_bench background execution daemon.
+
+This script is responsible for polling the storage for runnable experiments and
+executing them in parallel.
+
+See the current ``--help`` `output for details.
```
bpkroth, line 12:
```suggestion
See the current ``--help`` output for details.
```

mlos_bench/mlos_bench/mlos_benchd.py
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+#
+"""
+mlos_bench background execution daemon.
+
+This script is responsible for polling the storage for runnable experiments and
```
bpkroth, line 9:
```suggestion
This script is responsible for polling the storage for runnable :py:clas:`.Experiment`s and
```

bpkroth: Some minor tweaks to help make all of the docstring generation cross referencing.  Might need some tweaks.

mlos_bench/mlos_bench/mlos_benchd.py
```diff
@@ -0,0 +1,70 @@
...
+        default=1,
+        help="Number of workers to use. Default is 1.",
+    )
+    parser.add_argument(
+        "--poll_interval",
+        required=False,
+        default=1,
+        help="Polling interval in seconds. Default is 1.",
+    )
+    _main(parser.parse_args())
```
bpkroth, line 70: For testing, may also want some sort of hidden argument or environment variable used to set the `MAX_ITERATIONS`.

PR #959, opened 2025-03-11
bugfix: local env not passed to the child process when shell_env_params present

motus:
# Pull Request

## Title

Copy local environment variables and explicitly pass them to the child process.

## Description

When we explicitly provide the environment variables to pass to the `subprocess.run()` call, only those variables get passed to the child process. That prevents the child process from seeing the environment variables of the parent MLOS process. We've fixed that issue for MLOS on Windows but not for other systems. This PR fixes it for Linux and MacOS.

## Type of Change

- 🛠️ Bug fix
- 🧪 Tests

## Testing

Update the existing unit tests to check for MLOS parameters *and* parent environment variables being passed to the child process.

### DIFF
--- a/mlos_bench/mlos_bench/services/local/local_exec.py
+++ b/mlos_bench/mlos_bench/services/local/local_exec.py
@@ -206,12 +206,12 @@ class LocalExecService(TempDirContextService, SupportsLocalExec):
         if env_params:
             env = {key: str(val) for (key, val) in env_params.items()}
 
+        env_copy = environ.copy()
         if sys.platform == "win32":
             # A hack to run Python on Windows with env variables set:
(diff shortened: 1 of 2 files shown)


bpkroth:
Admittedly the previous fix was a bit of a hack, but I think we should clarify (and maybe even document ;) ) something along with this fix:

**How do variables propogate?**

One could argue that we need to list *expected* variables in `required_args` or `const_args`, right?  The latter could also reference the environment variable in its default even since we do variable expansion from local environment variables when they aren't in in the `$global_config`.

bpkroth (approved):
Looks fine, but let's document/test how this is expected/supposed to interact with `required_args` and `const_args`

Issue #542, opened 2023-10-17
mlos_bench: implement optional early abort logic for time based trials


bpkroth:
Further generalizing this via async `telemetry` collection during the process might be nice too.

bpkroth:
Some additional notes:

In cases of throughput or latency based benchmarks, it's not totally clear how to detect whether a specific trial is worse than a previous one, since some trial could theoretically speed up later or some such.

But, for raw time based ones, what we could do, would be to track the worst value seen so far, and then abort if we exceed that.
To do that, we'd need some additional metadata that this benchmark was in fact seeking to minimize wallclock time.

What's tricky is how we incorporate metrics from that.
Imagine for instance that you wanted to explain why some params/trials were bad.  But in aborting some trials, you give up on gathering that data.

Moreover, we can't actually store a real time value for that trial, since we abort it early.  Instead we need to store it in the DB as "ABORTED" or somesuch and then each time we train the optimizer fabricate a value for it.
Likely $W+\epsilon$ where $W$ is the worst value seen *up until that point* (i.e., serially examining historical trial data).

bpkroth:
Per discussions, we need:
- a new Environment phase - `abort`
  - the plan will be for users to add that as commands to their Environment configs that instruct the system how to cancel and cleanup a currently running `run` phase
  - that will get executed asynchronously
- an additional config option to inform the scheduler when to invoke early abort logic for time based benchmarks
  - specifically it needs to know which metrics to look at
  - this should probably be _per environment_
    - for instance, right now we don't often teardown the VM for each trial, so the first time the VM gets setup, that necessarily takes longer, so we shouldn't include that in our elapsed time metrics
    - could be that we start tracking the elapsed time for every single phase in each environment for each trial and try to infer things, or ...
    - we also add a `status` or `telemetry` phase that includes commands used to asynchronously poll status of a `run` phase (or should it also support other phases?) in order to feed in-progress metrics back into the system and allow specifying that one of those (maybe just an implicit elapsed time, but probably not if sometimes the db needs to be reloaded for instance and other times it doesn't so the run phase overall may take longer on occassion even if the actual benchmark portion sometimes doesn't)

bpkroth:
Async status polling should now be possible using a separate EventLoopContext instance in each TrialRunner now that #720 has been merged

PR #967, opened 2025-04-23
Add a Parallel Task Schedular

### LINKED ISSUE #380: parallel trial execution

jsfreischuetz:
# Pull Request

## Add a parallel task scheduler and associated test cases.

## Description

This PR adds a parallel task scheduler, associated schema additions, and test cases.
This requires minimal changes to the trial runner to ensure the results are properly registered with the storage backend. Importantly, tasks need to be registered in the main thread (https://stackoverflow.com/questions/6297404/multi-threaded-use-of-sqlalchemy), requiring a callback for asynchronous tasks. This does not change how anything needs to be done for synchronous schedulers.

Closes #380

## Type of Change

- ✨ New feature

## Testing

I have added a test case to ensure that the scheduler properly schedules all of the tasks.
This test case is similar to how the synchronous scheduler is made.

mlos_bench/mlos_bench/config/schemas/schedulers/scheduler-schema.json was formatted when I opened it in the dev container. I'm guessing there is no check for the format of the JSONC file. However, black will format it if given the opportunity. It is possible, but potentially somewhat difficult, to revert some of these changes without having black ignore the file.


bpkroth:
@jsfreischuetz leaving a few of my notes here from our chat.

1. Preference for multiprocessingpool over threads, similar to #939 
   > In the future we will likley also add async status polling as a separate thread *within* each TrialRunner (see: https://github.com/microsoft/MLOS/issues/542#issuecomment-2584680562)
2. We _think_ that we _should_ be able to have the individual (forked) `TrialRunner` instances connect to the same `Storage` (e.g., local sqlite.db via sqlalchemy), but it may have some backend nuances (e.g., does sqlite require file locking, or does it manage that for us?  do we need to make sure that the engine is disconnected prior to forking so it can reconnect individually after?, etc.)
3. There seems to be a bug where the `trial_runner_id` is not restored when `pending_trials` are fetched from the Storage backend.  We should fix that independently and add a test for it (new PR).
4. We discussed why we want to separate `assign_trial_runners` and `run_schedule` (which *just* runs the Trials which are ready to be run and assigned a runner).
5. However, you pointed out that there are advantages to allowing some Trials to be late schedueld (e.g., in the context of TUNA, when the budget is 1 we could run them on the very first TrialRunner that becomes available).  For that, we may want to extend support for a special case of e.g., `-1` to indicate *any* TrialRunner and `-2` for any TrialRunner that hasn't already run this config (and assign the true value late) or else adjust what `runnable_trials` should mean (see 5).  The point of this is to reduce the impact of stragglers in the overall runtime (though we could alleviate that later with support for early abort - #542).
6. We discussed fixing `_run_schedule` in the base `Scheduler` class to use an `storage.runnable_trials()` call after running `assign_trial_runners` to further filter for Trials which have had a TrialRunner assigned (or can accept *any* TrialRunner late as decribed above).

Did I miss anything?

jsfreischuetz:
I have made changes related to the refactor of separating the logic of assigning tasks to workers, and actually scheduling the workers.

Using processes as the main unit of parallelism has significant problems because of the management of certain objects deep within the codebase, which are unpickleable. I have attempted a few solutions to recreate these objects in every function, referencing these objects to remove the replication of these objects to new processes. However, this causes significant issues with ensuring the schema is created, since we can no longer guarantee this in the constructors. It may be possible, but it would require a significant refactor of the storage engine.

Because this implementation uses a thread pool rather than a process pool, I have used future.concurrent, as the documentation for multiprocessing suggests that their interface for thread pools doesn't make sense considering it was designed for process pools. (https://docs.python.org/3/library/multiprocessing.html)

mlos_bench/mlos_bench/environments/status.py
```diff
@@ -26,6 +27,7 @@ def is_good(self) -> bool:
             Status.READY,
             Status.RUNNING,
             Status.SUCCEEDED,
+            Status.SCHEDULED,
```
bpkroth, line 30: Why isn't PENDING good enough for that?

mlos_bench/mlos_bench/schedulers/parallel_scheduler.py
```diff
@@ -0,0 +1,144 @@
+#
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+#
+"""A simple single-threaded synchronous optimization loop implementation."""
```
bpkroth, line 5: update docstring

mlos_bench/mlos_bench/schedulers/parallel_scheduler.py
```diff
@@ -0,0 +1,144 @@
...
+from mlos_bench.environments.status import Status
+from mlos_bench.schedulers.base_scheduler import Scheduler
+from mlos_bench.storage.base_storage import Storage
+from mlos_bench.tunables.tunable_groups import TunableGroups
+
+_LOG = logging.getLogger(__name__)
+
+
+class ParallelScheduler(Scheduler):
+    """A simple multi-threaded asynchronous optimization loop implementation."""
```
bpkroth, line 25:
```suggestion
    """A simple multi-process asynchronous optimization loop implementation."""
```

mlos_bench/mlos_bench/schedulers/parallel_scheduler.py
bpkroth: Also, please add some more complex docstrings that explain what this is module can help with and how it does that that include some executable usage details of the config insantiation at least.  There should be some other examples elsewhere in the source tree that you can crib off of or I can help cook up an example in abit.

mlos_bench/mlos_bench/schedulers/parallel_scheduler.py
bpkroth: Also a comment that references the module can be used to help jump to the docstring with example config usage easily.  This is mostly a nicety for navigating online docs.

mlos_bench/mlos_bench/schedulers/parallel_scheduler.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+
+        super().__init__(*args, **kwargs)
+        self.pool = ThreadPoolExecutor(max_workers=len(self._trial_runners))
```
bpkroth, line 30: We discussed doing this with a process pool

mlos_bench/mlos_bench/storage/sql/experiment.py
```diff
@@ -235,13 +235,11 @@ def _get_key_val(conn: Connection, table: Table, field: str, **kwargs: Any) -> d
...
         )
 
-    def pending_trials(self, timestamp: datetime, *, running: bool) -> Iterator[Storage.Trial]:
+    def filter_trials_by_status(
+        self, timestamp: datetime, status: list[Status]
+    ) -> Iterator[Storage.Trial]:
         timestamp = utcify_timestamp(timestamp, origin="local")
         _LOG.info("Retrieve pending trials for: %s @ %s", self._experiment_id, timestamp)
-        if running:
-            pending_status = [Status.PENDING.name, Status.READY.name, Status.RUNNING.name]
```
bpkroth, line 242: What I don't love about the new API is it requires us to list all of these now.  Maybe we can add an alias for the "runnable" statuses in the `Status` module?

mlos_bench/mlos_bench/storage/sql/experiment.py
```diff
@@ -235,13 +235,11 @@ def _get_key_val(conn: Connection, table: Table, field: str, **kwargs: Any) -> d
             row._tuple() for row in cur_result.fetchall()  # pylint: disable=protected-access
         )
 
-    def pending_trials(self, timestamp: datetime, *, running: bool) -> Iterator[Storage.Trial]:
+    def filter_trials_by_status(
+        self, timestamp: datetime, status: list[Status]
```
bpkroth, line 239:
```suggestion
        self, timestamp: datetime, statuses: list[Status]
```

mlos_bench/mlos_bench/storage/sql/experiment.py
```diff
@@ -281,6 +279,13 @@ def pending_trials(self, timestamp: datetime, *, running: bool) -> Iterator[Stor
                     config=config,
                 )
 
+    def pending_trials(self, timestamp: datetime, *, running: bool) -> Iterator[Storage.Trial]:
```
bpkroth, line 282:
```suggestion
    def pending_trials(self, timestamp: datetime, *, running: bool) -> Iterator[Storage.Trial]:
```

I'm OK to get rid of this one and replace the others with `filter_trials_by_status` and feed it `Status.RUNNABLE_STATUSES` which returns that list as a literal or something.

Maybe separate that out to a different PR for cleanliness.

mlos_bench/mlos_bench/storage/base_storage.py
```diff
@@ -307,6 +307,29 @@ def load(
                 Trial ids, Tunable values, benchmark scores, and status of the trials.
             """
 
+        @abstractmethod
+        def filter_trials_by_status(
+            self,
+            timestamp: datetime,
+            status: list[Status],
+        ) -> Iterator["Storage.Trial"]:
```
bpkroth, line 315: You could potentially also extend this to `trial_runner_id >= 0` type of optional filter as well.

mlos_bench/mlos_bench/schedulers/parallel_scheduler.py
jsfreischuetz: Do you have an example of this? None of the other schedulers are doing this

mlos_bench/mlos_bench/config/schemas/schedulers/scheduler-schema.json
```diff
@@ -29,18 +27,23 @@
...
                 },
                 "trial_config_repeat_count": {
                     "description": "Number of times to repeat a config.",
                     "type": "integer",
                     "minimum": 1,
-                    "examples": [3, 5]
+                    "examples": [
+                        3,
+                        5
+                    ]
```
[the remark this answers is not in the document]
jsfreischuetz: yes

mlos_bench/mlos_bench/environments/status.py
jsfreischuetz: No, you need to differentiate between pending and scheduled. Scheduled implies that you have assigned a worker, scheduled implies that it has been given a process, but hasn't necessarily started yet (where it would get switched to running)

mlos_bench/mlos_bench/storage/base_storage.py
jsfreischuetz: Possible, but probably just a different function?

mlos_bench/mlos_bench/storage/sql/experiment.py
jsfreischuetz: This was already done in the previous version. I copied these lines

mlos_bench/mlos_bench/environments/status.py
bpkroth: Seems to me that `status=PENDING AND trial_runner_id != NULL` is the same as `SCHEDULED`, no?

bpkroth:
Moving this effort to #971

PR #970, opened 2025-04-28
Delay entering TrialRunner context until run_trial

bpkroth:
# Pull Request

## Title

Delay entering `TrialRunner` context until `run_trial`.

## Description

This is part of an attempt to try and see if can work around issues with `multiprocessing.Pool` needing to pickle certain objects when forking.

For instance, if the Environment is using an SshServer, we need to start an EventLoopContext in the background to handle the SSH connections and threads are not picklable.

Nor are file handles, DB connections, etc., so there may be other things we also need to adjust to make this work.

See Also #967

## Type of Change

- 🛠️ Bug fix
- 🔄 Refactor

## Testing

- Light so far (still in draft mode)
- Just basic existing CI tests (seems to not break anything)

I think this is incomplete.  To support forking inside the Scheduler and *then* entering the context of the given TrialRunner, we may also need to do something about the Scheduler's Storage object.

For now this is a draft PR to allow @jsfreischuetz and I to play with alternative organizations of #967.

### DIFF
--- a/mlos_bench/mlos_bench/schedulers/base_scheduler.py
+++ b/mlos_bench/mlos_bench/schedulers/base_scheduler.py
@@ -200,8 +200,9 @@ class Scheduler(ContextManager, metaclass=ABCMeta):
         _LOG.debug("Scheduler START :: %s", self)
         assert self.experiment is None
         assert not self._in_context
-        for trial_runner in self._trial_runners.values():
-            trial_runner.__enter__()
+        # NOTE: We delay entering the context of trial_runners until it's time
+        # to run the trial in order to avoid incompatibilities with
+        # multiprocessing.Pool.
         self._optimizer.__enter__()
         # Start new or resume the existing experiment. Verify that the
         # experiment configuration is compatible with the previous runs.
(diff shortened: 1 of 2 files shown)


motus (approved):
Looks nice and straightforward, although I have questions about the `TrialRunner` context. Do we need one for each instance of the `TrialRunner`? What's the life cycle? Let's merge this PR and discuss it

bpkroth:
> Looks nice and straightforward, although I have questions about the `TrialRunner` context. Do we need one for each instance of the `TrialRunner`? What's the life cycle? Let's merge this PR and discuss it

The idea is roughly to create a `TrialRunner` for each number of parallel trials we want to support, or else for the number of workers we want to sample noise in the target system from (e.g., in the case of Tuna).

The only real reason for delaying entering the context is to allow pickling the TrialRunner when passing it to a mp.Pool worker process (nuances of how Python does multiprocessing pools).

Since an Environment _could_ use something that requires state like a background tread (e.g., if it uses SshService so needs an EventLoopContext, which we'll eventually want for background async status polling too), then we don't want to enter the Environment context until we're ready to run it since those threads are not picklable.

From my looking around, the only thing I can think of that might be a slight overhead is that as we run multiple trials we enter and exit the context somewhat more frequently and so create and destroy those background threads.

But that happens on the order of minutes, not even seconds generally, so it's not a huge issue I think.

Aside from that `teardown` is executed as a separate step than leaving context anyways, so I don't think there's a huge risk there.

But happy to discuss more if you noticed something I didn't.

PR #974, opened 2025-05-09
Make Storage objects picklable and resumable by mp.Pool workers

bpkroth:
# Pull Request

## Title

Make Storage objects picklable and resumable by mp.Pool workers.

## Description

To support a `ParallelTrialScheduler` with `Trials` running thru `TrialRunners` in separate processes using `multiprocessing.Pool` we need to be able to "send" all necessary state to "blank" python processes using pickle since python doesn't actually use a `fork` strategy for multiprocessing.

This change enables that by doing the following:

1. Allow pickling the core `Storage` class by hooking `__getstate__` and `__setstate__` to disconnect the `Engine` prior to pickling and reconnect on unpickling (restore).
2. Allow recreating an `Experiment` and `Trial` object by fetching it from the `Storage` by `experiment_id` and `trial_id` respectively.

These two things allow gathering the details necessary for a `TrialRunner` running in a separate child process to reconnect to the DB independently, `run_trial`, and save the telemetry and results to the DB.
After which, the "MainProcess" can `load` the results from the Experiment from the Storage again to `bulk_register` the scores from those completed `Trials` and start a new set.

## Type of Change

- 🛠️ Bug fix
- ✨ New feature
- 🔄 Refactor
- 🧪 Tests

## Testing

- [x] New CI tests

This change should have no impact on the existing `SyncScheduler`.

### DIFF
--- a/mlos_bench/mlos_bench/environments/status.py
+++ b/mlos_bench/mlos_bench/environments/status.py
@@ -7,8 +7,8 @@
 import enum
 import logging
 
-
 _LOG = logging.getLogger(__name__)
+
 
 class Status(enum.Enum):
     """Enum for the status of the benchmark/environment Trial or Experiment."""
--- a/mlos_bench/mlos_bench/environments/status.py
+++ b/mlos_bench/mlos_bench/environments/status.py
@@ -91,4 +91,4 @@ class Status(enum.Enum):
         """Check if the status of the benchmark/environment Trial or Experiment is
         TIMED_OUT.
         """
-        return self == Status.FAILED
+        return self == Status.TIMED_OUT
--- a/mlos_bench/mlos_bench/environments/status.py
+++ b/mlos_bench/mlos_bench/environments/status.py
@@ -5,7 +5,10 @@
 """Enum for the status of the benchmark/environment Trial or Experiment."""
 
 import enum
+import logging
 
+
+_LOG = logging.getLogger(__name__)
 
 class Status(enum.Enum):
     """Enum for the status of the benchmark/environment Trial or Experiment."""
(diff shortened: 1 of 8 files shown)


bpkroth:
Windows doesn't allow for multiple processes to open the same file. I think I will just exclude this on Windows with an error that it's not supported there.

bpkroth:
> <details>
> <summary>Comments suppressed due to low confidence (1)</summary>
> 
> **mlos_bench/mlos_bench/environments/status.py:37**
> * The is_timed_out method is checking for Status.FAILED instead of Status.TIMED_OUT, which appears to be a logic bug that could lead to incorrect status reporting.
> ```
> return self == Status.FAILED
> ```
> </details>
> 
> 

This last comment is correct. We should correct that too.

mlos_bench/mlos_bench/environments/status.py
```diff
@@ -73,4 +91,4 @@ def is_timed_out(self) -> bool:
         """Check if the status of the benchmark/environment Trial or Experiment is
         TIMED_OUT.
         """
-        return self == Status.FAILED
+        return self == Status.TIMED_OUT
```
bpkroth, line 94: Unrelated bug that copilot found in its review.

mlos_bench/mlos_bench/environments/status.py
```diff
@@ -19,6 +22,21 @@ class Status(enum.Enum):
     FAILED = 6
     TIMED_OUT = 7
 
+    @staticmethod
+    def from_str(status_str: str) -> "Status":
```
bpkroth, line 26: Previously we did not restore `Status` on `Trials` restored from the DB.  This helps do that.

bpkroth: Should maybe check to see how we were doing that in the `TrialData` class.

bpkroth: Did that.

bpkroth, line 26: Hmm, probably need tests for this too.

mlos_bench/mlos_bench/tests/storage/test_storage_pickling.py
```diff
@@ -0,0 +1,97 @@
...
+                "class": "mlos_bench.storage.sql.storage.SqlStorage",
+                "config": {
+                    "drivername": "sqlite",
+                    "database": db_path,
+                    "lazy_schema_create": False,
+                },
+            }
+        )
+
+        storage = from_config(config_str)
```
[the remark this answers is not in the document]
bpkroth: Sure, I can add that in a new PR.

PR #975, opened 2025-05-09
Rename and reorganize some Scheduler methods

bpkroth:
# Pull Request

## Title

Rename and reorganize some Scheduler methods

## Description

In preparation for ParallelScheduler PR (#971) this PR

1. Renames a few methods for clarity (e.g., `schedule_trial` --> `add_trial_to_queue`) so that the role of other methods (e.g., `assign_trial_runners`) is more clear and readily overridable.
2. Moves some methods to the base class.
   This is done based on an observation that some of the methods (e.g., the main `start` loop) are largely reusable across both SyncScheduler and ParallelScheduler if the method separation is done a little bit more fine grained such that each can override the only the parts they need to.

Additionall, it also

1. Also adds a subtle tweak on the `pending_trials` method in order to filter based on whether or not the trial already has a runner assigned or not.
2. Adjusts `TrialRunner.run_trial` to return the results of the Trial.
   This isn't really used anywhere, but it is helpful to check that TrialRunners run in a child process finished successfully by using that result as a check in the return value, even though for Optimizer bulk_registering it actually pulls the results back off of Storage in the MainProcess.

## Type of Change

- 🔄 Refactor
- 🧪 Tests

## Testing

- [x] New CI tests for tweaks in `pending_trials`
- Existing CI tests

### DIFF
--- a/mlos_bench/mlos_bench/schedulers/base_scheduler.py
+++ b/mlos_bench/mlos_bench/schedulers/base_scheduler.py
@@ -520,9 +520,9 @@ class Scheduler(ContextManager, metaclass=ABCMeta):
             )
         )
         for trial in pending_trials:
-            if trial.trial_runner_id is None:
-                logging.warning("Trial %s has no TrialRunner assigned yet.")
-                continue
+            assert (
+                trial.trial_runner_id is not None
+            ), f"Trial {trial} has no TrialRunner assigned yet."
             self.run_trial(trial)
 
     def not_done(self) -> bool:
--- a/mlos_bench/mlos_bench/schedulers/base_scheduler.py
+++ b/mlos_bench/mlos_bench/schedulers/base_scheduler.py
@@ -240,7 +240,6 @@ class Scheduler(ContextManager, metaclass=ABCMeta):
         self._in_context = False
         return False  # Do not suppress exceptions
 
-    @abstractmethod
     def start(self) -> None:
         """Start the scheduling loop."""
         assert self.experiment is not None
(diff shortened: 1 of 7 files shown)


mlos_bench/mlos_bench/schedulers/base_scheduler.py
```diff
@@ -287,54 +306,88 @@ def load_tunable_config(self, config_id: int) -> TunableGroups:
...
+        # Check if the optimizer has converged or not.
         not_done = self.not_done()
         if not_done:
             tunables = self.optimizer.suggest()
-            self.schedule_trial(tunables)
-
+            self.add_trial_to_queue(tunables)
         return not_done
 
-    def schedule_trial(self, tunables: TunableGroups) -> None:
```
bpkroth, line 311: The term "schedule" was confusing to some as it sometimes was thought to imply "assign a runner", sometimes thought to imply "assign a start time", and sometimes thought to simply be "add to queue".  This changesit to explicitly only be the latter two.

mlos_bench/mlos_bench/schedulers/base_scheduler.py
```diff
@@ -434,25 +494,45 @@ def get_trial_runner(self, trial: Storage.Trial) -> TrialRunner:
         assert trial_runner.trial_runner_id == trial.trial_runner_id
         return trial_runner
 
-    def _run_schedule(self, running: bool = False) -> None:
+    def run_schedule(self, running: bool = False) -> None:
```
bpkroth, line 497: Make this public so it's more obvious that it can be overriden by subclasses.

mlos_bench/mlos_bench/schedulers/sync_scheduler.py
```diff
@@ -15,24 +15,10 @@
 class SyncScheduler(Scheduler):
     """A simple single-threaded synchronous optimization loop implementation."""
 
-    def start(self) -> None:
```
bpkroth, line 18: Moved to base class as it was nearly identical to what the ParallalScheduler would use.

mlos_bench/mlos_bench/schedulers/trial_runner.py
```diff
@@ -168,7 +169,7 @@ def run_trial(
         self,
         trial: Storage.Trial,
         global_config: dict[str, Any] | None = None,
-    ) -> None:
+    ) -> tuple[Status, datetime, dict[str, TunableValue] | None]:
```
bpkroth, line 172: TODO: Change the docstring to match.

bpkroth: Hmm, I guess it already was that way.

mlos_bench/mlos_bench/schedulers/base_scheduler.py
```diff
@@ -242,7 +242,6 @@ def __exit__(
         self._in_context = False
         return False  # Do not suppress exceptions
 
-    @abstractmethod
     def start(self) -> None:
         """Start the scheduling loop."""
```
bpkroth, line 246: Maybe expand on this documentation so that it notes what the other functions it calls and in what sequence so someone reading the docs could know how overrides might interact.

motus (approved):
Looks very nice!

mlos_bench/mlos_bench/schedulers/base_scheduler.py
motus: makes perfect sense

mlos_bench/mlos_bench/schedulers/base_scheduler.py
motus: yep. we can do it in the next PR

motus (approved):
Looks very nice and clean! Thanks!

PR #985, opened 2025-06-02
Fixup and refactor root_env_config resolution

bpkroth:
# Pull Request

## Title

Fixup refactor/allow git repo root env resolution from different dirs and restores

## Description

- Fixes a bug when running tests from `mlos_bench/mlos_bench/tests` directory in which the root env config is initially resolved from the root of the repo and storage interally as `mlos_bench/mlos_bench/tests/environment.jsonc` and next re-used by another component (e.g., `Scheduler`) as that new path which doesn't exist when run from the `mlos_bench/mlos_bench/tests` dir.
- Now, we keep track of both relative and absolute paths *in memory* and use the latter when necessary.
- Additionally, reuses the git repo info stored in the DB directly for restoring an Experiment from Storage (e.g., for forked TrialRunners in the ParallelScheduler) rather than trying to resolve it independently.
  This part is mostly cosmetic, since currently those values aren't used by TrialRunners or Environments, but is good to fix for potential future works.

## Type of Change

- 🛠️ Bug fix
- 🔄 Refactor
- 🧪 Tests

## Testing

- Local
- CI unit tests

This is part of a series of fixups and should be merged after #991 

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,7 +1,11 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-FROM mcr.microsoft.com/vscode/devcontainers/base AS base
+# TODO: FIXME: Temporarily pin to bookworm until latest (trixie) base image is fully supported.
+# See Also:
+# - <https://github.com/microsoft/MLOS/issues/1002>
+# - .devcontainer/devcontainer.json
+FROM mcr.microsoft.com/vscode/devcontainers/base:bookworm AS base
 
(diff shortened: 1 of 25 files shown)


mlos_bench/mlos_bench/util.py
```diff
@@ -274,33 +274,147 @@ def check_required_params(config: Mapping[str, Any], required_params: Iterable[s
         )
 
 
-def get_git_info(path: str = __file__) -> tuple[str, str, str]:
+def get_git_root(path: str = __file__) -> str:
```
bpkroth, line 277: The git changes should be merged in #991 first to make it easier to review this PR.

